PR #19, opened 2018-11-13
[CEDS-729] - added cancellation journey for integration.

raghu1978:


app/config/AppConfig.scala
```diff
@@ -51,7 +51,9 @@ class AppConfig @Inject() (override val runModeConfiguration: Configuration, val
   lazy val customsDeclarationsUrl = loadConfig("urls.customsDeclarations")
 
   lazy val customsDeclarationsEndpoint = baseUrl("customs-declarations")
-  lazy val submitImportDeclarationUri = getConfString("customs-declarations.submit-uri",
+  lazy val submitExportDeclarationUri = getConfString("customs-declarations.submit-uri",
+    throw new IllegalStateException("Missing configuration for Customs Declarations submission URI"))
+  lazy val submitCancellationUri = getConfString("customs-declarations.cancel-uri",
     throw new IllegalStateException("Missing configuration for Customs Declarations submission URI"))
```
PatrykRudnicki: Change an exception to:
`"Missing configuration for Customs Declarations cancellation URI"`

app/controllers/CancelDeclarationController.scala
```diff
@@ -0,0 +1,126 @@
...
+                )
+              )
+            )
+        }
+      }
+    }
+  )
+}
+
+  private def createCancellationMetadata(form:CancelDeclarationForm):MetaData =
```
PatrykRudnicki: We can stay with this here, but I think that we should create services to put methods inside services, not in Controller. We have a lot of code in controllers that we shouldn't have.

conf/application.conf
```diff
@@ -78,7 +78,8 @@ microservice {
 
     customs-declarations {
       host = localhost
-      port = 6790
+     # port = 9820
```
PatrykRudnicki: Revert this file. Stay with correct port. No comments in `application.conf`

app/controllers/CancelDeclarationController.scala
raghu1978: this method will not exist after we have UI navigation built

conf/application.conf
raghu1978: added comment for local testing to point to customs-declarations if needed

PR #22, opened 2018-11-22
CEDS-743: view start page

NC-1234:


PatrykRudnicki:
@raghu1978 I think that in this place we shouldn't have authentication. After this screen, you'll be redirected to the login page, so here it should be without authentication.

test/controllers/StartControllerSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+    "return 200 for a GET" in {
+      val result = route(app, getRequest(uri)).get
+
+      status(result) must be (OK)
+    }
+    "display radio button to start now" in {
+      val result = route(app, getRequest(uri)).get
+      val stringResult = contentAsString(result)
+
+      stringResult must include(messages("details of where you’re sending the export"))
```
PatrykRudnicki: Please use strings from `messages.en`

PatrykRudnicki:
We can think that we need to use redirect method inside a controller or normal `href` in view will be enough

PR #23, opened 2018-11-22
[CEDS-764] - frontend metrics for submission, cancel, movements consolidation

raghu1978:


test/base/CustomExportsBaseSpec.scala
```diff
@@ -58,13 +59,14 @@ trait CustomExportsBaseSpec extends PlaySpec
   protected val contextPath: String = "/customs-declare-exports"
 
   lazy val mockCustomsCacheService: CustomsCacheService = mock[CustomsCacheService]
-
+    val mockMetrics : ExportsMetrics = mock[ExportsMetrics]
```
PatrykRudnicki: propose to use `lazy val`

PR #29, opened 2018-11-28
Movement journey

PatrykRudnicki:


app/services/CustomsCacheService.scala
```diff
@@ -44,4 +51,53 @@ class CustomsCacheService @Inject()(
   override implicit val crypto: CompositeSymmetricCrypto = applicationCrypto.JsonCrypto
 
   override def shortLiveCache: ShortLivedHttpCaching = caching
+
+  def fetchMovementRequest(cacheId: String)(implicit hc: HeaderCarrier,
+    executionContext: ExecutionContext): Future[Option[InventoryLinkingMovementRequest]] = {
+      fetch(cacheId).map {
+        case Some(cacheMap) =>
```
PatrykRudnicki: This whole cases can be changed to `fetch(cacheId).map(fun(_))` it's the same but short

PatrykRudnicki:
What about tests for method `fetchMovementRequest`?

PR #36, opened 2018-12-07
CEDS-766 Declaration Type Page

MaciejRewera:


test/controllers/supplementary/DeclarationTypePageControllerSpec.scala
```diff
@@ -0,0 +1,178 @@
...
+    "return 303 code" in {
+      withCaching[DeclarationType](None)
+
+      val validForm = ExportsTestData.correctDeclarationType
+      val result = route(app, postRequest(uri, validForm)).get
+
+      status(result) must be(SEE_OTHER)
+    }
+
+    pending
```
PatrykRudnicki: Please remove pending from tests

MaciejRewera: Tests are based on acceptance criteria, but one of them is to redirect to consignment-reference page, which will be implemented in the next sprint. This is why I decided to write tests now and mark them as pending.

PatrykRudnicki:
Nice cleaning in few places and good idea to generate form in companion object

PR #41, opened 2018-12-18
[CEDS-808] Page for declaration additional actors

PatrykRudnicki:


app/forms/supplementary/DeclarationAdditionalActors.scala
```diff
@@ -28,18 +29,41 @@ case class DeclarationAdditionalActors(
...
+    "partyType" -> mandatoryIfNot(
+      "eori",
+      "",
+      text().verifying("supplementary.partyType.error", input => allowedPartyTypes(input))
     )
   )(DeclarationAdditionalActors.apply)(DeclarationAdditionalActors.unapply)
 
   def form(): Form[DeclarationAdditionalActors] = Form(mapping)
+
+  def toMetadataProperties(actors: DeclarationAdditionalActors): Map[String, String] =
```
MaciejRewera: I would suggest moving it to the case class. Makes more sense to me.

PR #46, opened 2018-12-19
CEDS:819: Not Eligible Page

NC-1234:


app/controllers/NotEligibleController.scala
```diff
@@ -0,0 +1,42 @@
...
+import services.CustomsCacheService
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+import views.html.not_eligible
+
+import scala.concurrent.Future
+
+class NotEligibleController @Inject()(
+  appConfig: AppConfig,
+  override val messagesApi: MessagesApi,
+  authenticator: AuthAction,
```
PatrykRudnicki: Please should use `authenticate` instead of `authenticator`.

app/controllers/NotEligibleController.scala
```diff
@@ -0,0 +1,42 @@
...
+  errorHandler: ErrorHandler,
+  customsCacheService: CustomsCacheService
```
PatrykRudnicki: You don't need to have here `errorHandler` and `customsCacheService`. Please inject only those classes that you use.

test/controllers/supplementary/NotEligibleControllerSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  "NotEligible" should {
+    "return 200 with a success" in {
+      authorizedUser()
+
+      val result = route(app, getRequest(uri)).get
+
+      status(result) must be(OK)
+    }
+    "display page content" in {
```
MaciejRewera: Add authorizedUser() at the beginning of the test.
I am surprised it passed, but it shouldn't.

PR #53, opened 2019-01-02
created couple of pages

harishkancharla:


app/forms/MovementForms.scala
```diff
@@ -32,13 +32,26 @@ object ChoiceForm {
...
     "choice" -> text().verifying("Incorrect value", correctChoice.contains(_))
   )(ChoiceForm.apply)(ChoiceForm.unapply)
 }
 
+case class RoleForm(roleForm: String)
+
+object RoleForm {
+  implicit val format = Json.format[RoleForm]
+
+  //TODO change to enum
```
PatrykRudnicki: Not enum, it can be sealed trait and case classes, enum is not good :) I think we can stay with `Seq`. It can be `Set` too

app/controllers/MovementController.scala
```diff
@@ -27,6 +27,7 @@ import play.api.data.Form
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.mvc.{Action, AnyContent}
 import services.CustomsCacheService
+import uk.gov.hmrc.http.cache.client.CacheMap
```
PatrykRudnicki: You don't need this import

app/controllers/MovementController.scala
```diff
@@ -54,7 +55,12 @@ class MovementController @Inject()(
         Future.successful(BadRequest(choice_page(appConfig, formWithErrors))),
       form =>
         customsCacheService.cache[ChoiceForm](appConfig.appName, choiceId, form).map { _ =>
-          Redirect(controllers.routes.MovementController.displayDucrPage())
+            if(form.choice.toString == "SMP") {
```
PatrykRudnicki: Instead of `if` my propose is to use pattern matching. In the future, we will have three different journeys (arrival and departure have the same journey, always redirection to ducr page)

PR #64, opened 2019-01-11
CEDS-877 Supervising customs office page

PatrykRudnicki:


app/forms/supplementary/GoodsLocation.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  val formId = "GoodsLocation"
+
+  val mapping = Forms.mapping(
+    "country" -> optional(
+      text()
+        .verifying("supplementary.country.empty", _.trim.nonEmpty)
+        .verifying(
+          "supplementary.country.error",
+          input => input.isEmpty || !allCountries.filter(country => country.countryName == input).isEmpty
```
MaciejRewera: It gets entangled, so I would suggest to use nonEmpty() method at least.
And you can use utils.validators.FormFieldValidator to check length.

PatrykRudnicki: Yeah, we can use non-empty. I'll change it.

About using `utils.validators.FormFieldValidator` I don't know why should I use it. I don't check the length. I'm online checking that country exists on the list.

PatrykRudnicki: I see it, in different places. Yeah, agree

app/forms/supplementary/GoodsLocation.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  val formId = "GoodsLocation"
+
+  val mapping = Forms.mapping(
+    "country" -> optional(
+      text()
+        .verifying("supplementary.country.empty", _.trim.nonEmpty)
+        .verifying(
+          "supplementary.country.error",
+          input => input.isEmpty || allCountries.filter(country => country.countryName == input).nonEmpty
```
MaciejRewera: I am wondering if it would be better to use `input.trim.isEmpty` ?

app/forms/supplementary/SupervisingCustomsOffice.scala
```diff
@@ -0,0 +1,36 @@
...
+import play.api.libs.json.Json
+
+case class SupervisingCustomsOffice(office: Option[String])
+
+object SupervisingCustomsOffice {
+  implicit val format = Json.format[SupervisingCustomsOffice]
+
+  val mapping = Forms.mapping(
+    "supervisingCustomsOffice" -> optional(
+      text().verifying("supplementary.supervisingCustomsOffice.error", _.length <= 8)
```
MaciejRewera: Here you can use `noLongerThan(_, 8)`

app/forms/supplementary/GoodsLocation.scala
PatrykRudnicki: Yup

app/forms/supplementary/SupervisingCustomsOffice.scala
PatrykRudnicki: Yeah, thanks!

PatrykRudnicki:
Review fixes done! Merging in 3, 2, 1...

PR #71, opened 2019-01-16
CEDS-883 Procedure Codes

MaciejRewera:


app/controllers/supplementary/LocationController.scala
```diff
@@ -41,21 +42,20 @@ class LocationController @Inject()(
 
   def displayForm(): Action[AnyContent] = authenticate.async { implicit request =>
     customsCacheService.fetchAndGetEntry[GoodsLocation](appConfig.appName, formId).map {
-      case Some(data) => Ok(goods_location(appConfig, form.fill(data)))
-      case _          => Ok(goods_location(appConfig, form))
+      case Some(data) => Ok(goods_location(appConfig, form().fill(data)))
```
PatrykRudnicki: You don't need to have `()` here. If you define a method as `def form()` you can invoke it with or without parenthesis. It's good to show that it's a method, but do we need a method for it?

It's correct but unnecessary in my opinion

app/forms/supplementary/ConsignmentReferences.scala
```diff
@@ -42,7 +42,7 @@ object ConsignmentReferences {
     "ducr" -> optional(Ducr.ducrMapping),
     "lrn" -> text()
       .verifying("supplementary.consignmentReferences.lrn.error.empty", _.trim.nonEmpty)
-      .verifying("supplementary.consignmentReferences.lrn.error.length", noLongerThan(lrnMaxLength))
+      .verifying("supplementary.consignmentReferences.lrn.error.length", hasSpecificLength(lrnMaxLength))
```
PatrykRudnicki: Are you sure that `lrn` must have a specific length?

app/forms/supplementary/ProcedureCodes.scala
```diff
@@ -0,0 +1,65 @@
...
+}
+
+object ProcedureCodes {
+  implicit val format = Json.format[ProcedureCodes]
+
+  private val procedureCodeLength = 4
+  private val additionalProcedureCodeLength = 3
+  val mapping = Forms.mapping(
+    "procedureCode" -> text()
+    .verifying("supplementary.procedureCodes.procedureCode.error.empty", _.trim.nonEmpty)
```
PatrykRudnicki: We can create `nonEmpty` method inside our validation helper.

app/forms/supplementary/ProcedureCodes.scala
```diff
@@ -0,0 +1,65 @@
...
+import play.api.data.{Form, Forms}
+import play.api.libs.json.Json
+import utils.validators.FormFieldValidator._
+
+case class ProcedureCodes(
+  procedureCode: String,      // max 4 alphanumeric characters
+  additionalProcedureCodes: Seq[String]     // max 99 codes, each is max 3 alphanumeric characters
+) {
+
+  def toMetadataProperties(): Map[String, String] = {
```
PatrykRudnicki: Probably we should move all methods `toMetadataProperties` to class, only address needs to have it in a companion object, we can add it to our TODO list

app/forms/supplementary/ProcedureCodes.scala
```diff
@@ -0,0 +1,65 @@
...
+  val mapping = Forms.mapping(
+    "procedureCode" -> text()
+    .verifying("supplementary.procedureCodes.procedureCode.error.empty", _.trim.nonEmpty)
+    .verifying("supplementary.procedureCodes.procedureCode.error.length", isEmpty or hasSpecificLength(procedureCodeLength))
+    .verifying("supplementary.procedureCodes.procedureCode.error.specialCharacters", isAlphanumeric),
+  "additionalProcedureCodes" -> seq(text()
+    .verifying("supplementary.procedureCodes.additionalProcedureCode.error.length", isEmpty or hasSpecificLength(additionalProcedureCodeLength))
+    .verifying("supplementary.procedureCodes.additionalProcedureCode.error.specialCharacters", isAlphanumeric)
+  )
+    .verifying("supplementary.procedureCodes.additionalProcedureCode.error.singleEmpty", _.exists(_.trim.nonEmpty))
```
PatrykRudnicki: The same here about `nonEmpty`. Do you need to use exists and check this condition on the sequence? Or we can move it to a specific element like above for length and special character

MaciejRewera: It has to be that way, because it is checking if there is **at least one** non empty element.

PR #72, opened 2019-01-16
CEDS-876 one field removed

Jodlowy:


app/forms/supplementary/WarehouseIdentification.scala
```diff
@@ -21,25 +21,24 @@ import play.api.data.{Form, Forms}
...
 object WarehouseIdentification {
   implicit val format = Json.format[WarehouseIdentification]
 
   val formId = "IdentificationOfWarehouse"
 
-  val mapping = Forms.mapping(
-    "typeCode" -> optional(text().verifying("supplementary.warehouse.typeCode.error", noLongerThan(35))),
+  val mapping = Forms.mapping( 
     "identificationNumber" -> optional(
       text().verifying("supplementary.warehouse.identificationNumber.error", noLongerThan(35))
```
PatrykRudnicki: So based on preparing data, this field must have minimum two characters, first is for ID and rest of the string is for type code. What do you think about adding validation for minimum 2 chars? We have it inside optional so it can't affect the optional field.

PR #76, opened 2019-01-17
CEDS-872 Fixes for exporter ID page and whole journey

PatrykRudnicki:


app/controllers/ChoiceController.scala
```diff
@@ -39,6 +39,7 @@ class ChoiceController @Inject()(
   errorHandler: ErrorHandler
 )(implicit ec: ExecutionContext)
     extends FrontendController with I18nSupport {
+  import forms.Choice.AllowedChoiceValues._
```
MaciejRewera: This import can be moved to other imports section, outside the controller class.

PR #95, opened 2019-01-24
Tests from Field form validator

PatrykRudnicki:


app/utils/validators/FormFieldValidator.scala
```diff
@@ -68,10 +83,11 @@ object FormFieldValidator {
       case _              => false
   }
 
-  val isAllCapitalLetter: String => Boolean = (input: String) => input.matches(allCapitalLettersRegex)
-
   val isDecimal: String => Boolean = (input: String) => input.matches(decimalRegex)
```
MaciejRewera: The method `isDecimalNoLongerThan` is not implemented corretly, according to its name at least.
For now, it is checking if the String contains no more digits than stated. But it may contain hundreads of letters.

test/util/validators/FormFieldValidatorSpec.scala
```diff
@@ -179,6 +361,220 @@ class FormFieldValidatorSpec extends WordSpec with MustMatchers {
...
+      }
+    }
+
+    "return true" when {
+      "string contains only digits and dot" in {
+        isDecimal("123456.123456") must be(true)
+      }
+    }
+  }
+
```
MaciejRewera: Added comment on this method in its definition place. We cannot proceed with the current form.

app/utils/validators/FormFieldValidator.scala
```diff
@@ -20,21 +20,31 @@ import scala.util.{Success, Try}
...
-  private val numericRegexValue = "[0-9]*"
-  private val alphabeticRegexValue = "[a-zA-Z]*"
-  private val alphanumericRegexValue = "[a-zA-Z0-9]*"
-  private val firstCapitalLetter = "[A-Z]{1}(.*)"
-  private val zerosOnlyRegexValue = "[0]*"
-  private val allCapitalLettersRegex = "[A-Z]*"
+  private val zerosOnlyRegexValue = "[0]+"
   private val noMoreDecimalPlacesThanRegexValue: Int => String = (decimalPlaces: Int) =>
     s"^([0-9]*)([\\.]{0,1}[0-9]{0,$decimalPlaces})$$"
   private val decimalRegex = "^-?[0-9]+[\\.]([0-9]+)?$"
```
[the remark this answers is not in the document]
PatrykRudnicki: Done

PR #96, opened 2019-01-28
CEDS-914 New Cancel declaration screen

PatrykRudnicki:


app/controllers/CancelDeclarationController.scala
```diff
@@ -49,103 +47,37 @@ class CancelDeclarationController @Inject()(
...
-                        message = messagesApi("global.error.message")
-                      )
-                    )
+          val metadata = form.createCancellationMetadata(request.user.eori)
+          exportsMetrics.startTimer(cancelMetric)
+          customsDeclarationsConnector.submitCancellation(metadata).flatMap {
+            case CustomsDeclarationsResponse(ACCEPTED, Some(_)) =>
+              Future.successful(Ok(cancellation_confirmation_page(appConfig)))
+            case error =>
+              exportsMetrics.incrementCounter(cancelMetric)
```
MaciejRewera: Metrics should be incremented for the Accepted response as well.

app/views/components/input_textarea.scala.html
```diff
@@ -23,7 +23,7 @@
     hint: Option[String] = None
 )(implicit messages: Messages)
 
-<div class="form-group @if(field.hasErrors){form-group-error}">
+<div class="form-field @if(field.hasErrors){form-field--error}">
```
MaciejRewera: Please, double check if the name of the class contains double or single dash ("--").

PatrykRudnicki: Double :)

PR #99, opened 2019-01-30
CEDS-930 Update declaration status during cancel and handle cancellation

PatrykRudnicki:


test/base/MockConnectors.scala
```diff
@@ -103,4 +98,7 @@ trait MockConnectors extends MockitoSugar {
     when(mockNrsConnector.submitNonRepudiation(any())(any(), any()))
       .thenReturn(Future.successful(NrsSubmissionResponse("submissionId1")))
 
+  def successfullyCancelDeclarationResponse(response: Boolean) =
```
MaciejRewera: I'd rather slightly change the name to 'successfulCancelDeclarationResponse'.
Current name suggests different behaviour.

test/controllers/CancelDeclarationControllerSpec.scala
```diff
@@ -37,7 +37,6 @@ class CancelDeclarationControllerSpec extends CustomExportsBaseSpec {
   "CancelDeclarationController" should {
     "return 200 with a success" in {
       authorizedUser()
```
MaciejRewera: There is authorizedUser() invocation in every test, so I'd suggest to use `BeforeAndAfter` and `before {...}` clause.

PR #102, opened 2019-02-06
CEDS-937 toMetadataProperties testing

MaciejRewera:


app/forms/supplementary/AdditionalDeclarationType.scala
```diff
@@ -0,0 +1,53 @@
...
+
+package forms.supplementary
+
+import play.api.data.Forms.text
+import play.api.data.{Form, Forms, Mapping}
+import play.api.libs.json.Json
+
+case class AdditionalDeclarationType(
+  value: String // 1 upper case alphabetic character
+)
```
PatrykRudnicki: Do you need to have a method `toMetadataProperties`?

test/controllers/supplementary/ConsignmentReferencesControllerSpec.scala
```diff
@@ -85,12 +85,12 @@ class ConsignmentReferencesControllerSpec extends CustomExportsBaseSpec with Bef
...
 
     "populate the form fields with data from cache" in {
-      val result = displayConsignmentReferencesPageTestScenario(
-        Some(ConsignmentReferences(ducr = Some(Ducr(exemplaryDucr)), lrn = "1234ABCD"))
-      )
+      val result = displayConsignmentReferencesPageTestScenario(Some(correctConsignmentReferences))
 
       contentAsString(result).replaceAll(" ", "") must include("name=\"ducr.ducr\"\nvalue=\"" + exemplaryDucr + "\"")
-      contentAsString(result).replaceAll(" ", "") must include("name=\"lrn\"\nvalue=\"1234ABCD\"")
+      contentAsString(result).replaceAll(" ", "") must include(
```
PatrykRudnicki: If you have more than one `contentAsString` is better to extract it to variable

app/forms/supplementary/AdditionalDeclarationType.scala
MaciejRewera: Not here. AdditionalDeclarationType together with DispatchLocation are parts of DeclarationType. Alone, they have no mapping.
DeclarationType only is going to be converted to properties.

test/controllers/supplementary/ConsignmentReferencesControllerSpec.scala
MaciejRewera: Done

PR #103, opened 2019-02-07
CEDS-934: Product owner feedback changes

NC-1234:


app/forms/supplementary/RepresentativeDetails.scala
```diff
@@ -49,7 +49,7 @@ object RepresentativeDetails {
   implicit val format = Json.format[RepresentativeDetails]
 
   private val representativeStatusCodeAllowedValues =
-    Set(StatusCodes.DirectRepresentative, StatusCodes.IndirectRepresentative)
+    Set(StatusCodes.Declarant,StatusCodes.DirectRepresentative, StatusCodes.IndirectRepresentative)
```
PatrykRudnicki: Inside this object, you can import `StatusCodes` model
`import StatusCodes._` and after that, your Set can look like `Set(Declarant, DirectRepresentative, IndirectRepresentative)`, but it's not a problem

app/views/supplementary/previous_documents.scala.html
```diff
@@ -28,15 +28,15 @@
     appConfig = appConfig
 ) {
     @helpers.form(PreviousDocumentsController.savePreviousDocuments(), 'autoComplete -> "off") {
-        @components.back_link("/customs-declare-exports/declaration/supplementary/package-information")
+        @components.back_link("/customs-declare-exports/declaration/supplementary/transaction-type")
 
         @components.error_summary(form.errors)
 
-        @components.page_title(Some("supplementary.previousDocuments.title"))
+        @components.page_title(Some("supplementary.previousDocuments.title"),Some("For example, 8GB-123456789101-SHIP1"))
```
PatrykRudnicki: Please move `"For example, 8GB-123456789101-SHIP1"` to messages and put key here

test/controllers/ChoiceControllerSpec.scala
```diff
@@ -91,15 +91,15 @@ class ChoiceControllerSpec extends CustomExportsBaseSpec with BeforeAndAfter {
         .cache[Choice](any(), ArgumentMatchers.eq(Choice.choiceId), any())(any(), any(), any())
     }
 
-    "redirect to role page when simplified declaration chosen" in {
+    "redirect to dispatch-location page when simplified declaration chosen" in {
```
PatrykRudnicki: if you can please remove `-` from this place. It's test description, so you don't need to use specific naming like view or something like this. Like one problem above, it's not important to change it

PR #104, opened 2019-02-08
CEDS-911 Add possibility to add more codes

PatrykRudnicki:


app/controllers/supplementary/ProcedureCodesPageController.scala
```diff
@@ -39,29 +41,176 @@ class ProcedureCodesPageController @Inject()(
...
-      .bindFromRequest()
-      .fold(
-        (formWithErrors: Form[ProcedureCodes]) =>
-          Future.successful(BadRequest(procedure_codes(appConfig, formWithErrors))),
-        validProcedureCodes =>
-          customsCacheService
-            .cache[ProcedureCodes](supplementaryDeclarationCacheId, ProcedureCodes.id, validProcedureCodes)
-            .map { _ =>
-              Redirect(controllers.supplementary.routes.SupervisingCustomsOfficeController.displayForm())
+    val bindedForm = form.bindFromRequest()
```
MaciejRewera: It should be `boundForm`

app/controllers/supplementary/ProcedureCodesPageController.scala
```diff
@@ -39,29 +41,176 @@ class ProcedureCodesPageController @Inject()(
...
+            customsCacheService.cache[ProcedureCodesData](cacheId, formId, updatedCache).map { _ =>
+              Redirect(controllers.supplementary.routes.SupervisingCustomsOfficeController.displayForm())
+            }
+        }
+    }
+  //scalastyle:on methodLength
+
+  private def retrieveProcedureCode(action: String): String = action.dropWhile(_ != ':').drop(1)
+
+  private def errorPage()(implicit request: Request[_]): Future[Result] =
```
MaciejRewera: This and the next methods names are slightly misleading. Please, add a verb to it, like 'display'.

app/forms/supplementary/ProcedureCode.scala
```diff
@@ -17,22 +17,58 @@
...
+        .verifying("supplementary.procedureCodes.additionalProcedureCode.error.specialCharacters", isAlphanumeric)
+    )
+  )(ProcedureCode.apply)(ProcedureCode.unapply)
+
+  val formId = "ProcedureCodes"
+
+  def form(): Form[ProcedureCode] = Form(mapping)
+}
+
+case class ProcedureCodesData(procedureCode: Option[String], additionalProcedureCodes: Seq[String])
```
MaciejRewera: There is no need to have both case classes in the same file. I'd rather extract it to separate one.

app/forms/supplementary/ProcedureCode.scala
```diff
@@ -41,32 +77,14 @@ case class ProcedureCodes(
...
-    "additionalProcedureCodes" -> seq(
-      text()
-        .verifying(
-          "supplementary.procedureCodes.additionalProcedureCode.error.length",
-          isEmpty or hasSpecificLength(additionalProcedureCodeLength)
-        )
-        .verifying("supplementary.procedureCodes.additionalProcedureCode.error.specialCharacters", isAlphanumeric)
-    ).verifying("supplementary.procedureCodes.additionalProcedureCode.error.singleEmpty", _.exists(_.trim.nonEmpty))
-  )(ProcedureCodes.apply)(ProcedureCodes.unapply)
+  def contains(code: String): Boolean = additionalProcedureCodes.contains(code)
```
MaciejRewera: This method's name is also not fully explanatory. At first I thought it would also look after Procedure Code.

app/forms/supplementary/ProcedureCode.scala
```diff
@@ -17,22 +17,58 @@
...
+      text()
+        .verifying(
+          "supplementary.procedureCodes.additionalProcedureCode.error.length",
+          isEmpty or hasSpecificLength(additionalProcedureCodeLength)
+        )
+        .verifying("supplementary.procedureCodes.additionalProcedureCode.error.specialCharacters", isAlphanumeric)
+    )
+  )(ProcedureCode.apply)(ProcedureCode.unapply)
+
+  val formId = "ProcedureCodes"
```
MaciejRewera: Please, change to the same ID as class name.

app/forms/supplementary/ProcedureCode.scala
PatrykRudnicki: Yeah, but those classes are related to each other, so I think it should be in the same place

PR #110, opened 2019-02-13
CEDS-943 Add possibility to handle more than one holder

PatrykRudnicki:


app/controllers/SimpleDeclarationController.scala
```diff
@@ -95,7 +95,7 @@ class SimpleDeclarationController @Inject()(
       case _          => Ok(simple_declaration(appConfig, form, allCountries))
     }
   }
-
+  //scalastyle:off method.length
```
MaciejRewera: Maybe create sub-methods instead?

app/controllers/supplementary/DeclarationHolderController.scala
```diff
@@ -33,29 +35,155 @@ class DeclarationHolderController @Inject()(
...
+        )
+    }
+  }
+
+  private def addAnotherHolderHandler(
+    userInput: DeclarationHolder,
+    cachedData: DeclarationHoldersData
+  )(implicit request: Request[_], hc: HeaderCarrier): Future[Result] =
+    (userInput, cachedData.holders) match {
+      case (_, holders) if holders.length >= 99 =>
```
MaciejRewera: I'd advise to create a constant for this "99" value. Somewhere where it could be accessible, for e.g. in DeclarationHolder object?

app/controllers/supplementary/DeclarationHolderController.scala
```diff
@@ -33,29 +35,155 @@ class DeclarationHolderController @Inject()(
...
+              case "Add"                             => addAnotherHolderHandler(validForm, cache)
+              case "Save and continue"               => saveAndContinueHandler(validForm, cache)
+              case value if value.contains("Remove") => removeHolderHandler(retrieveHolder(actionType), cache)
+              case _                                 => errorHandler.displayErrorPage()
+          }
+        )
+    }
+  }
+
+  private def addAnotherHolderHandler(
```
MaciejRewera: I'd suggest a different names for methods. Handler is better for a class but here it is not required to be readable. 'addHolder' should be enough.

app/utils/validators/FormFieldValidator.scala
```diff
@@ -55,6 +55,9 @@ object FormFieldValidator {
 
   val hasSpecificLength: Int => String => Boolean = (length: Int) => (input: String) => input.length == length
 
+  val lengthRange: Int => Int => String => Boolean = (min: Int) =>
```
MaciejRewera: `lengthInRange` maybe?

PR #119, opened 2019-02-19
CEDS-941 confirmation page

MaciejRewera:


app/controllers/supplementary/SummaryPageController.scala
```diff
@@ -52,7 +52,7 @@ class SummaryPageController @Inject()(
   def displayPage(): Action[AnyContent] = authenticate.async { implicit request =>
     customsCacheService.fetch(supplementaryCacheId).map {
       case Some(cacheMap) => Ok(summary_page(appConfig, SupplementaryDeclarationData(cacheMap)))
-      case None           => Ok(summary_page(appConfig, SupplementaryDeclarationData()))
+      case None           => BadRequest(summary_page_no_data(appConfig))
```
PatrykRudnicki: I'm not sure about BadRequest here. It's not an error, but an empty summary page, so `summary_page_no_data` should be enough

app/controllers/supplementary/SummaryPageController.scala
```diff
@@ -73,7 +73,13 @@ class SummaryPageController @Inject()(
               .saveSubmissionResponse(submission)
               .flatMap { _ =>
                 exportsMetrics.incrementCounter(submissionMetric)
-                customsCacheService.remove(supplementaryCacheId).map(_ => Ok("Supplementary Declaration submitted"))
+                customsCacheService
+                  .remove(supplementaryCacheId)
+                  .map(
+                    _ =>
+                      Redirect(controllers.supplementary.routes.ConfirmationPageController.displayPage())
+                        .flashing(prepareFlashScope(lrn.getOrElse(""), conversationId))
```
PatrykRudnicki: I'm not sure about `.getOrElse`... I'm thinking about some exception when lrn is empty. As I know this field cannot be empty.

test/controllers/supplementary/ConfirmationPageControllerSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+import play.api.test.Helpers._
+
+class ConfirmationPageControllerSpec extends CustomExportsBaseSpec {
+
+  private val lrn = "1234567890"
+  private val conversationId = "12QW-34ER-56TY-78UI-90OP"
+
+  override protected def getRequest(
+    uri: String,
+    headers: Map[String, String] = Map.empty
```
PatrykRudnicki: You don't use `headers` here. I'm not sure that we need to override `getRequest` method. We can create the method `getRequestWithFlash` or something like this and use `getRequest` method from `CustomExportsBaseSpec`.

I'm not sure that this case is good to use override not create a new method

test/controllers/supplementary/ConfirmationPageControllerSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+  ): FakeRequest[AnyContentAsEmpty.type] =
+    super
+      .getRequest(uri)
+      .withFlash("LRN" -> lrn, "ConversationId" -> conversationId)
+
+  private trait test {
```
PatrykRudnicki: Trait name should start with capital letter

test/controllers/supplementary/SummaryPageControllerSpec.scala
```diff
@@ -48,132 +49,151 @@ class SummaryPageControllerSpec extends CustomExportsBaseSpec {
...
-    "return 200 code" in new test {
-      val result = route(app, getRequest(summaryPageUri)).get
-      status(result) must be(OK)
-    }
+  "Summary page on displayPage" when {
 
-    "display \"back\" button that links to Documents Produced page" in new test {
-      val resultAsString = contentAsString(route(app, getRequest(summaryPageUri)).get)
+    "there is data in cache for supplementary declaration" should {
+      "return 200 code" in new test {
```
PatrykRudnicki: Please use capital letter for `test` trait

PR #120, opened 2019-02-20
Fix for empty fields in cache

PatrykRudnicki:


app/models/declaration/supplementary/SupplementaryDeclarationData.scala
```diff
@@ -36,7 +36,7 @@ case class SupplementaryDeclarationData(
   override def toMetadataProperties(): Map[String, String] =
     this.toMap.values.foldLeft(Map("declaration.functionCode" -> suppDecFunctionCode)) { (map, convertable) =>
       map ++ convertable.toMetadataProperties()
-    }
+    }.filter(_._2 != "")
```
MaciejRewera: String class has nonEmpty() method as well. I'd use that here.

PR #122, opened 2019-02-22
CEDS-939 Notification status

PatrykRudnicki:


app/models/Status.scala
```diff
@@ -17,38 +17,107 @@
...
 
   implicit object StatusFormat extends Format[Status] {
     def reads(status: JsValue): JsResult[Status] = status match {
-      case JsString("Accepted")  => JsSuccess(Accepted)
-      case JsString("Arrived")   => JsSuccess(Arrived)
-      case JsString("Departed")  => JsSuccess(Departed)
-      case JsString("PreLodged") => JsSuccess(PreLodged)
-      case JsString("Query")     => JsSuccess(Query)
-      case _                     => JsError("Incorrect value")
+      case JsString("01")   => JsSuccess(Accepted)
```
raghu1978: why Jsstring?

app/models/Status.scala
```diff
@@ -17,38 +17,107 @@
...
 
-case object Query extends Status
+case object Released extends Status
+
+case object Cleared extends Status
+
+case object Cancelled extends Status
+
+case object CustomsPositionGranted extends Status {
+  override def toString(): String = "Customs Position Granted"
```
raghu1978: move these into ENUM please

test/controllers/supplementary/ConfirmationPageControllerSpec.scala
```diff
@@ -53,24 +56,47 @@ class ConfirmationPageControllerSpec extends CustomExportsBaseSpec {
       resultAsString must include(messages("supplementary.confirmation.explanation"))
     }
 
-    "display LRN from flash" in new Test {
```
raghu1978: are we using flash? shouldn't use the scope on MDTP please use cache

PatrykRudnicki: It's actually implemented in previous task, not related with actual one. @MaciejRewera check this

app/models/Status.scala
PatrykRudnicki: Why?

app/models/Status.scala
PatrykRudnicki: so we're sending this as Json, the values will be JsString, code inside will be the same like in notifications

test/controllers/supplementary/ConfirmationPageControllerSpec.scala
MaciejRewera: If we shouldn't, then I will change it to cache in separate PR.

app/models/Status.scala
PatrykRudnicki: https://medium.com/@yuriigorbylov/scala-enumerations-hell-5bdba2c1216

If you can I can only move it to `java.util.Enum`, but I don't see any pros of it

PR #123, opened 2019-02-26
Multiplicity enum switch

MaciejRewera:


app/views/components/add_save_button.scala.html
```diff
@@ -14,12 +14,14 @@
...
+
 @(
     saveKey: String = "site.save_and_continue",
     addKey: String = "site.add"
 )(implicit messages: Messages)
 
 <div class="section align='right'">
-    <button id="add" name="action" class="button" value="@messages(addKey)">@messages(addKey)</button>
-    <button id="submit" name="action" class="button" value="@messages(saveKey)">@messages(saveKey)</button>
+    <button id="add" class="button" name="@Add.toString" value="">@messages(addKey)</button>
```
PatrykRudnicki: You can remove `value=""`. You don't use it.

app/views/components/add_save_button.scala.html
```diff
@@ -14,12 +14,14 @@
...
+    <button id="submit" class="button" name="@SaveAndContinue.toString" value="">@messages(saveKey)</button>
```
PatrykRudnicki: Same here

app/views/components/add_save_button.scala.html
MaciejRewera: You're right

PR #125, opened 2019-02-27
CEDS-949 Add multiplicity to 2/2 AdditionalInformation

PauloPortugal:
* User should be able to enter up to 99 entries as per existing
   validations for this data element
 * User should not be able to enter the same value more than once
 * This data element is made up of 2 fields. Both fields must be
   populated for user to add the entry
 * The entries should be displayed as a list and show that the two
   fields are associated. E.g. 'field1-field2'

### DIFF
--- a/app/controllers/supplementary/AdditionalInformationController.scala
+++ b/app/controllers/supplementary/AdditionalInformationController.scala
@@ -18,13 +18,19 @@ package controllers.supplementary
 import config.AppConfig
 import controllers.actions.AuthAction
 import controllers.util.CacheIdGenerator.supplementaryCacheId
+import controllers.util.{Add, FormAction, Remove, SaveAndContinue}
 import forms.supplementary.AdditionalInformation
-import forms.supplementary.AdditionalInformation._
+import forms.supplementary.AdditionalInformation.form
+import handlers.ErrorHandler
 import javax.inject.Inject
-import play.api.data.Form
+import models.declaration.supplementary.AdditionalInformationData
+import models.declaration.supplementary.AdditionalInformationData.{formId, maxNumberOfItems}
+import models.requests.AuthenticatedRequest
+import play.api.data.{Form, FormError}
 import play.api.i18n.{I18nSupport, MessagesApi}
-import play.api.mvc.{Action, AnyContent}
+import play.api.mvc.{Action, AnyContent, Request, Result}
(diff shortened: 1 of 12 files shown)


app/controllers/supplementary/AdditionalInformationController.scala
```diff
@@ -34,27 +39,160 @@ class AdditionalInformationController @Inject()(
...
+    }
+  }
+
+  private def saveAndContinue(
+    userInput: AdditionalInformation,
+    cachedData: AdditionalInformationData
+  )(implicit request: AuthenticatedRequest[_], hc: HeaderCarrier): Future[Result] =
+    (userInput, cachedData.items) match {
+      case (item, Seq()) => handleSaveAndContinueEmptyCache(userInput, item)
+      case (item, items) => handleSaveAndContinueCache(userInput, item, items)
```
PatrykRudnicki: `userInput` and `item` are totally the same, but different name, something goes wrong here.  Please remove one of them

PauloPortugal: Thank you for spotting that @PatrykRudnicki .Fixed!

app/controllers/supplementary/AdditionalInformationController.scala
```diff
@@ -34,27 +39,158 @@ class AdditionalInformationController @Inject()(
...
+      .map(_.getOrElse(AdditionalInformationData(Seq())))
+
+    cachedData.flatMap { cache =>
+      boundForm
+        .fold(
+          (formWithErrors: Form[AdditionalInformation]) =>
+            Future.successful(BadRequest(additional_information(appConfig, formWithErrors, cache.items))),
+          validForm =>
+            actionType match {
+              case "Add"                             => addItem(validForm, cache)
```
PatrykRudnicki: @MaciejRewera changed the approach of using those cases. Please rebase or update your branch and run tests after it, probably it will fail

Look here and below for action type match: https://github.com/hmrc/customs-declare-exports-frontend/blob/master/app/controllers/supplementary/DeclarationHolderController.scala#L60

PauloPortugal: Thank you @PatrykRudnicki and @MaciejRewera! This has been addressed, branch rebased  and commit amended.

PR #129, opened 2019-03-05
CEDS-949 Add multiplicity to 2/3 DocumentsProduced

PauloPortugal:
* User should be able to enter up to 99 entries as per existing
   validations for this data element
 * User should not be able to enter the same value more than once
 * This data element is made up of 6 fields
 * The entries should be displayed as a list and show that the 6
   fields are associated

### DIFF
--- a/app/controllers/supplementary/DocumentsProducedController.scala
+++ b/app/controllers/supplementary/DocumentsProducedController.scala
@@ -18,13 +18,22 @@ package controllers.supplementary
 
 import config.AppConfig
 import controllers.actions.AuthAction
+import controllers.supplementary.routes.{DocumentsProducedController, SummaryPageController}
 import controllers.util.CacheIdGenerator.supplementaryCacheId
+import controllers.util.{Add, FormAction, Remove, SaveAndContinue}
 import forms.supplementary.DocumentsProduced
+import forms.supplementary.DocumentsProduced.form
+import handlers.ErrorHandler
 import javax.inject.Inject
-import play.api.data.Form
+import models.declaration.supplementary.DocumentsProducedData
+import models.declaration.supplementary.DocumentsProducedData.{formId, maxNumberOfItems}
+import models.requests.AuthenticatedRequest
+import play.api.data.{Form, FormError}
 import play.api.i18n.{I18nSupport, MessagesApi}
-import play.api.mvc.{Action, AnyContent}
+import play.api.libs.json.{JsValue, Json}
+import play.api.mvc.{Action, AnyContent, Request, Result}
 import services.CustomsCacheService
+import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import views.html.supplementary.documents_produced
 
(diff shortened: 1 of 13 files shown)


app/controllers/supplementary/DocumentsProducedController.scala
```diff
@@ -34,29 +41,130 @@ class DocumentsProducedController @Inject()(
...
+    implicit request: AuthenticatedRequest[_]
+  ) =
+    document match {
+      case _ if documents.length >= maxNumberOfItems =>
+        handleErrorPage(Seq(("", "supplementary.addDocument.maximumAmount.error")), document, documents)
+
+      case _ if documents.contains(document) =>
+        handleErrorPage(Seq(("", "supplementary.addDocument.duplicated")), document, documents)
+
+      case _ if document.isDefined =>
```
PatrykRudnicki: This and next case are very similar, my propose is to use

`val updatedCache = if(document.isDefined) DocumentsProducedData(documents :+ document) else DocumentsProducedData(documents)`

To avoid having to cases, in both of them we will make a redirect, in one case it'll be one unnecessary call to the database. You can improve my propose and based on this cache or don't cache data.

app/controllers/supplementary/DocumentsProducedController.scala
```diff
@@ -34,29 +41,130 @@ class DocumentsProducedController @Inject()(
...
+          Redirect(controllers.supplementary.routes.SummaryPageController.displayPage())
+        }
+
+      case _ if !document.isDefined =>
+        Future.successful(Redirect(controllers.supplementary.routes.SummaryPageController.displayPage()))
+    }
+
+  private def handleSaveAndContinueEmptyCache(document: DocumentsProduced)(implicit request: AuthenticatedRequest[_]) =
+    document match {
+      case _ if document.isDefined =>
```
PatrykRudnicki: The same like above, here when the page is optional you don't need to use pattern matching, always will be redirected. Add only cache data when document is defined

PatrykRudnicki: The golden rule for me don't use pattern matching for Booleans

app/controllers/supplementary/ProcedureCodesPageController.scala
```diff
@@ -188,7 +188,6 @@ class ProcedureCodesPageController @Inject()(
             }
         }
     }
-  //scalastyle:on methodLength
```
PatrykRudnicki: If you remove turn on scalastyle you need to remove turn off too

app/forms/supplementary/DocumentsProduced.scala
```diff
@@ -80,4 +71,19 @@ object DocumentsProduced {
...
+    val fields = value.split('-')
+
+    if (fields.length == 0) DocumentsProduced(None, None, None, None, None, None)
+    else
+      DocumentsProduced(
+        Some(fields(0)),
+        Some(fields(1)),
+        Some(fields(2)),
+        Some(fields(3)),
+        Some(fields(4)),
```
PatrykRudnicki: When you put only 3 fields probably you will have error here with retrieving 3, 4 or 5 field

app/controllers/supplementary/DocumentsProducedController.scala
```diff
@@ -34,29 +42,129 @@ class DocumentsProducedController @Inject()(
...
+
+      case _ if !document.isDefined =>
+        Future.successful(Redirect(controllers.supplementary.routes.SummaryPageController.displayPage()))
+    }
+
+  private def addItem(
+    userInput: DocumentsProduced,
+    cachedData: DocumentsProducedData
+  )(implicit request: AuthenticatedRequest[_], hc: HeaderCarrier): Future[Result] =
+    (userInput, cachedData.documents) match {
```
PatrykRudnicki: missing case when document is not defined, different solution in previous comment

app/controllers/supplementary/DocumentsProducedController.scala
PauloPortugal: I think we should reuse the code there. 
In the case the `! document.isDefined`,AKA _(not defined)_, we do not need to do a DB call.

app/forms/supplementary/DocumentsProduced.scala
```diff
@@ -29,32 +28,27 @@ case class DocumentsProduced(
...
-        documentQuantity.getOrElse("")
-    )
+  def toJson: JsValue = Json.toJson(this)
 }
 
 object DocumentsProduced {
+
   implicit val format = Json.format[DocumentsProduced]
 
+  implicit val reads = Json.reads[DocumentsProduced]
```
PatrykRudnicki: Hmm, why do you need `implicit val reads` when you have above `implicit val format`? Format contains reads and writes. It can be some corner case that scala cannot use only format, but I'm not sure about it

app/controllers/supplementary/DocumentsProducedController.scala
PauloPortugal: I am all about golden rules!!! Much better than silver, although copper is becoming quite fashionable these days.

app/controllers/supplementary/DocumentsProducedController.scala
PatrykRudnicki: yeah, so I think that we should use solution to not do db call when you don't need it

app/views/supplementary/summary/additional_documentation_section.scala.html
```diff
@@ -14,34 +14,33 @@
...
 @import models.viewmodels.HtmlTableRow
-
-@(additionalDocumentation: Option[DocumentsProduced])(implicit messages: Messages)
+@(additionalDocumentationData: Option[DocumentsProducedData])(implicit messages: Messages)
 
 @components.summary_list(Some(messages("supplementary.summary.additionalDocumentation.header"))) {
     @components.table_row_no_change_link(HtmlTableRow(
         label = messages("supplementary.summary.additionalDocumentation.documentTypeCode"),
-        value = additionalDocumentation.flatMap(_.documentTypeCode)
+        value = additionalDocumentationData.map(_.documents.map(_.documentTypeCode))
```
PatrykRudnicki: The first `map` should be `flatMap`.

It's difficult to say it based on the github code, but... maps through the options must return Option[Option[???]]

app/controllers/supplementary/DocumentsProducedController.scala
PauloPortugal:
```
if (document.isDefined)
      customsCacheService
        .cache[DocumentsProducedData](supplementaryCacheId, formId, DocumentsProducedData(documents :+ document))
    Future.successful(Redirect(controllers.supplementary.routes.SummaryPageController.displayPage()))
```

This is my suggestion, 
One of my golden rules is to avoid `else` keyword. So the common code always gets run and the exception is caught by the if statement. If also read better :)
What do you think?

app/controllers/supplementary/DocumentsProducedController.scala
PauloPortugal: I have amended this. Please review and resolve both conversations if you are happy ;)

app/controllers/supplementary/ProcedureCodesPageController.scala
PauloPortugal: Yes I made this by _"mistake"_ let me put this back on, but the next time I see this code I am going to refactor the methods/functions to not be so long. :)

app/controllers/supplementary/DocumentsProducedController.scala
PatrykRudnicki: Yep, it's really good approach, but here when you have call to database we should do like this, so probably the easiest one will be with stay with cases, because now the call to database will be started, but you're not sure that it'll finish with successful before redirection

app/views/supplementary/summary/additional_documentation_section.scala.html
```diff
@@ -14,34 +14,33 @@
...
 @components.summary_list(Some(messages("supplementary.summary.additionalDocumentation.header"))) {
     @components.table_row_no_change_link(HtmlTableRow(
         label = messages("supplementary.summary.additionalDocumentation.documentTypeCode"),
-        value = additionalDocumentation.flatMap(_.documentTypeCode)
+        value = additionalDocumentationData.flatMap(_.documents.map(_.documentTypeCode))
     ))
     @components.table_row_no_change_link(HtmlTableRow(
         label = messages("supplementary.summary.additionalDocumentation.documentId"),
-        value = additionalDocumentation.flatMap(_.documentIdentifier)
+        value = additionalDocumentationData.map(_.documents.map(_.documentIdentifier))
```
PatrykRudnicki: This comment about `flatMap` is for whole this file. Here you use a `map` and the same in lines: 32, 36, 40, 44

app/controllers/supplementary/DocumentsProducedController.scala
PauloPortugal: I have addressed your concern about the `Future` and the `redirect` problem. Well spotted!

app/views/supplementary/summary/additional_documentation_section.scala.html
PauloPortugal: If I change it to a `FlatMap` I actually have a compilation error. It works in this scenario :)

PatrykRudnicki: I check the type of this element. It's `Option[Seq[Option[String]` whole this: `additionalDocumentationData.map(_.documents.map(_.documentTypeCode))`

We're handling this because HtmlTableRow can take `value: Option[_]`. So flatMap is wrong, just for your information.

PR #132, opened 2019-03-06
CEDS-944 CEDS-945 Item Type multiplicity

MaciejRewera:


PauloPortugal:
I really like your commit hygiene :)

app/controllers/supplementary/ItemTypePageController.scala
```diff
@@ -42,21 +46,139 @@ class ItemTypePageController @Inject()(
...
+    ItemType(
+      combinedNomenclatureCode = itemTypeInput.combinedNomenclatureCode,
+      taricAdditionalCodes = itemTypeCache.taricAdditionalCodes ++ itemTypeInput.taricAdditionalCodes,
+      nationalAdditionalCodes = itemTypeCache.nationalAdditionalCodes ++ itemTypeInput.nationalAdditionalCodes,
+      descriptionOfGoods = itemTypeInput.descriptionOfGoods,
+      cusCode = itemTypeInput.cusCode,
+      statisticalValue = itemTypeInput.statisticalValue
+    )
+
+  private def validateItemType(itemType: ItemType)(implicit request: Request[AnyContent]): Future[ValidationResult] =
```
PatrykRudnicki: You can get rid of this method or pass action type. You use this method inside `handleAddition` and `handleSaveAndContinue` so you know what type you have. Based on it you don't need to use `getOrElse`

app/forms/supplementary/ItemType.scala
```diff
@@ -53,74 +63,58 @@ case class ItemType(
...
-        "supplementary.itemType.description.error.length",
-        isEmpty or noLongerThan(descriptionOfGoodsMaxLength)
-      ),
-    "cusCode" -> optional(
-      text()
-        .verifying("supplementary.itemType.cusCode.error.length", hasSpecificLength(cusCodeLength))
-        .verifying("supplementary.itemType.cusCode.error.specialCharacters", isAlphanumeric)
-    ),
+    "combinedNomenclatureCode" -> text(),
+    taricAdditionalCodesKey -> default(seq(text()), Seq.empty),
```
PatrykRudnicki: Why this? `default(seq(text()), Seq.empty)`

app/controllers/supplementary/ItemTypePageController.scala
MaciejRewera: Removed

app/forms/supplementary/ItemType.scala
MaciejRewera: Because without default() we would bind empty string into the sequence.

PR #133, opened 2019-03-08
General approach of handling forms - implemented on AdditionalInformationPage

PatrykRudnicki:
DO NOT MERGE IT

I want to share the idea of the general approach. To presentation, I choose Additional Information page for now only Add method works, but please focus and check MultiplyItemHelper object.

Amount of code it's good shown inside PreviousDocumentsController and way how to handle it.

AdditionalInformation page has this solution but needs more changes to fix everything.

I'm waiting for feedback

### DIFF
--- a/app/controllers/supplementary/AdditionalInformationController.scala
+++ b/app/controllers/supplementary/AdditionalInformationController.scala
@@ -15,22 +15,21 @@
  */
 
 package controllers.supplementary
+
 import config.AppConfig
 import controllers.actions.AuthAction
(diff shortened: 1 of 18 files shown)


PauloPortugal:
@PatrykRudnicki please `git commit --amend` to fix you typo from `CEDS-974 Commit to show general appraoch` to `CEDS-974 Commit to show general approach`

PatrykRudnicki:
Thanks! So I'll continue with this approach and try to make it better :)

PR #135, opened 2019-03-12
CEDS-975 - 3/37 - Additional supply chain actors - header level multiplicity

mattclark-zerogravit:
* added ability to add and remove additional actors / parties from the supply chain actors page.
* updated the summary page so the additional party details are included


app/controllers/supplementary/DeclarationAdditionalActorsController.scala
```diff
@@ -34,31 +43,123 @@ class DeclarationAdditionalActorsController @Inject()(
...
-      .bindFromRequest()
-      .fold(
-        (formWithErrors: Form[DeclarationAdditionalActors]) =>
-          Future.successful(BadRequest(declaration_additional_actors(appConfig, formWithErrors))),
-        form =>
-          customsCacheService.cache[DeclarationAdditionalActors](supplementaryCacheId, formId, form).map { _ =>
-            Redirect(controllers.supplementary.routes.DeclarationHolderController.displayForm())
-        }
-      )
+  def saveAndContinue(
```
MaciejRewera: Please, change the order of methods, so that the caller is before callee.
And some of methods does not have to be public.

app/forms/supplementary/DeclarationAdditionalActors.scala
```diff
@@ -16,23 +16,23 @@
...
+  implicit val writes = Json.writes[DeclarationAdditionalActors]
 
-  override def toMetadataProperties(): Map[String, String] =
-    Map(
-      "declaration.goodsShipment.aeoMutualRecognitionParties[0].id" -> eori.getOrElse(""),
-      "declaration.goodsShipment.aeoMutualRecognitionParties[0].roleCode" -> partyType.getOrElse("")
-    )
+  def isDefined: Boolean = eori.isDefined && partyType.isDefined
+
+  def toJson: JsValue = Json.toJson(this)
```
MaciejRewera: This is briliant. I mean, it's a shame I did not think about it earlier. It would save me some trouble.

app/views/supplementary/declaration_additional_actors.scala.html
```diff
@@ -16,32 +16,55 @@
...
+                        <td scope="row">@actor.eori</td>
+                        <td>@actor.partyType</td>
+                        <td><button class="button" name="@Remove.toString" value="@actor.toJson">@messages("site.remove")</button></td>
+                    </tr>
+                }
+                </tbody>
+            </table>
+        }
+
         @components.error_summary(form.errors)
```
MaciejRewera: I would expect `error_summary` to be placed before the table, but if it's required then I'm fine with that.

app/views/supplementary/declaration_additional_actors.scala.html
```diff
@@ -16,32 +16,55 @@
...
 
 @main_template(
     title = messages("supplementary.additionalActors.title"),
     appConfig = appConfig
 ) {
-    @helpers.form(DeclarationAdditionalActorsController.saveAdditionalActors(), 'autoComplete -> "off") {
+    @helpers.form(DeclarationAdditionalActorsController.saveForm(), 'autoComplete -> "off") {
         @components.back_link("/customs-declare-exports/declaration/supplementary/representative-details")
 
+        @if(actors.nonEmpty) {
```
MaciejRewera: Can we have some kind of a generic component for the table? It would be handy later. There is something similar: `single_value_elements_table`, but it won't handle case when there is more fields.

app/views/supplementary/summary/parties_section.scala.html
```diff
@@ -63,28 +63,44 @@
...
         value = parties
                 .flatMap(_.declarationHoldersData)
                 .flatMap(data => Some(data.holders))
                 .getOrElse(Seq())
-                .map(_.eori)
+                .map(_.authorisationTypeCode)
     ))
+
+    @components.summary_list(Some(messages("supplementary.summary.parties.additionalParties.header"))) {
+        @if(parties.map(_.declarationAdditionalActorsData.map(_.actors.nonEmpty))) {
```
MaciejRewera: Same here (regarding generic component for data presentation).

test/controllers/supplementary/DeclarationAdditionalActorsControllerSpec.scala
```diff
@@ -44,41 +57,229 @@ class DeclarationAdditionalActorsControllerSpec extends CustomExportsBaseSpec {
...
+      val result = route(app, getRequest(uri)).get
+      val stringResult = contentAsString(result)
+
+      status(result) must be(OK)
+      stringResult must include(messages("site.back"))
+      stringResult must include(messages("/declaration/supplementary/representative-details"))
+    }
+  }
+
+  private def testHappyPathsScenarios(
```
MaciejRewera: 👍

test/forms/supplementary/DeclarationAdditionalActorsSpec.scala
```diff
@@ -17,38 +17,50 @@
...
-import play.api.libs.json.{JsObject, JsString, JsValue}
+import play.api.libs.json.{JsArray, JsObject, JsString, JsValue}
 
 class DeclarationAdditionalActorsSpec extends WordSpec with MustMatchers {
   import DeclarationAdditionalActorsSpec._
 
   "Method toMetadataProperties" should {
     "return proper Metadata Properties" in {
-      val additionalActors = correctAdditionalActors
+      val additionalActorsData = correctAdditionalActorsData
```
MaciejRewera: Since we may have multiple values in it, could you test against this scenario? To make sure we can handle more values and indexes.

app/controllers/supplementary/DeclarationAdditionalActorsController.scala
```diff
@@ -34,31 +43,126 @@ class DeclarationAdditionalActorsController @Inject()(
...
+  errorHandler: ErrorHandler,
   customsCacheService: CustomsCacheService
 )(implicit ec: ExecutionContext)
     extends FrontendController with I18nSupport {
 
-  import DeclarationAdditionalActors._
-
-  val additionalActorsForm = form()
+  val exceedMaximumNumberError = "supplementary.additionalActors.maximumAmount.error"
+  val duplicateActorError = "supplementary.additionalActors.duplicated.error"
```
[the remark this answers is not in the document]
PauloPortugal: Good point

app/forms/supplementary/DeclarationAdditionalActors.scala
PauloPortugal: This was the thing I was telling you about!!! :) 
When I reviewed your MR I mention this.

app/views/supplementary/declaration_additional_actors.scala.html
PauloPortugal: @MaciejRewera There was no requirement, we are just following a pattern.

app/views/supplementary/declaration_additional_actors.scala.html
PauloPortugal: Actually that would be a good idea, but also I have no idea how to do it and how can it be generic.

app/controllers/supplementary/DeclarationAdditionalActorsController.scala
mattclark-zerogravit: 👍

PauloPortugal:
PR looks really good :stuck_out_tongue: 
I am a bit of a sucker for git hygiene. So I can see that we have a `merge` form master instead of a `rebase`. There is a good article https://www.atlassian.com/git/tutorials/merging-vs-rebasing about the differences between them that can illustrate the advantages of having a rebase.
In a nutshell git commits are ordered based on timestamps. If two developers code at the same time, i.e, commit at the same time and do a merge instead of a rebase their commits can be tangled. If at least one does a `rebase` this will make sure their commits are all grouped together. This is good in environment deployments where you want to role back a feature introduced by some developer and you do not need to roll anything else. 

But like I said I am a sucker for it and I do follow the Linux kernel rules loosely and the 7 golden commit standards.

But I digress. You are a joy to pair with and the PR is a :star:

app/views/supplementary/declaration_additional_actors.scala.html
mattclark-zerogravit: I think we need to address this across the service once we have the full journey.

app/views/supplementary/declaration_additional_actors.scala.html
mattclark-zerogravit: Will review why this is here and move if necessary

mattclark-zerogravit: Have moved

app/forms/supplementary/DeclarationAdditionalActors.scala
```diff
@@ -16,23 +16,23 @@
...
 import play.api.data.Forms.{optional, text}
 import play.api.data.{Form, Forms}
-import play.api.libs.json.Json
+import play.api.libs.json.{JsValue, Json}
 import uk.gov.voa.play.form.ConditionalMappings.mandatoryIfNot
 
-case class DeclarationAdditionalActors(eori: Option[String], partyType: Option[String])
-    extends MetadataPropertiesConvertable {
+case class DeclarationAdditionalActors(eori: Option[String], partyType: Option[String]) {
+  implicit val writes = Json.writes[DeclarationAdditionalActors]
```
PatrykRudnicki: I'm curious why you need implicit writes when you have defined implicit format in companion object

app/forms/supplementary/DeclarationAdditionalActors.scala
```diff
@@ -16,23 +16,23 @@
...
-      "declaration.goodsShipment.aeoMutualRecognitionParties[0].roleCode" -> partyType.getOrElse("")
-    )
+  def isDefined: Boolean = eori.isDefined && partyType.isDefined
+
+  def toJson: JsValue = Json.toJson(this)
 }
 
 object DeclarationAdditionalActors {
+
+  def fromJson(value: JsValue): DeclarationAdditionalActors = Json.fromJson(value).get
```
PatrykRudnicki: I don't like `get`. Probably it's `Option` so to avoid problems with empty data my propose is to use `getOrElse("")`

build.sbt
```diff
@@ -59,15 +59,12 @@ lazy val scoverageSettings: Seq[Setting[_]] = Seq(
...
-    ,"forms\\..*"
-    ,"views\\..*"
-    ,"domain\\..*"
     ,"metrics\\..*"
     ,"features\\..*"
+    ,"test\\..*"
     ,".*(BuildInfo|Routes|Options).*"
   ).mkString(";"),
-  coverageMinimum := 70,
+  coverageMinimum := 83,
```
PatrykRudnicki: 👍  Did you choose an actual value? :D

app/forms/supplementary/DeclarationAdditionalActors.scala
mattclark-zerogravit: Because the Json.toJson needs implicit writes

build.sbt
mattclark-zerogravit: just bumped it up to what it was passing at so people are encouraged to not let this keep slipping down to 70%

PR #136, opened 2019-03-12
Ceds 946

harishkancharla:


PauloPortugal:
@harishkancharla `CEDS-946: Rebase and resolve conflicts` this commit is not necessary if you have done a `rebase`. We also should not have commits for merge, but then again I can see that this is something we are not suggesting

PauloPortugal:
It also has 14 commits which is quite difficult to read through them. Normally commits tell a story, here is a bit of a mess, as every single change was committed instead of grouping them logically by behaviour or features.
This can be avoided doing `git rebase -i HEAD~?` to organise commits or `git commit --amend` to amend a previous commit.

app/views/supplementary/package_information.scala.html
```diff
@@ -19,19 +19,51 @@
...
+                <thead>
+                <tr>
+                    <th scope="col">@messages("supplementary.packageInformation.typesOfPackages")</th>
+                    <th scope="col">@messages("supplementary.packageInformation.numberOfPackages")</th>
+                    <th scope="col">@messages("supplementary.packageInformation.shippingMarks")</th>
+                    <th scope="col">@messages("supplementary.packageInformation.remove")</th>
+                </tr>
+                </thead>
+                <tbody>
+                @packages.zipWithIndex.map { case (item, index) =>
```
PauloPortugal: :+1: We should use this approach from now on

project/AppDependencies.scala
```diff
@@ -34,7 +34,8 @@ object AppDependencies {
     "org.jsoup" % "jsoup" % "1.10.3" % "test",
     "com.typesafe.play" %% "play-test" % PlayVersion.current % "test",
     "org.mockito" % "mockito-core" % "2.13.0" % "test",
-    "org.scalacheck" %% "scalacheck" % "1.13.4" % "test"
+    "org.scalacheck" %% "scalacheck" % "1.14.0" % "test"
```
PauloPortugal: :+1:

app/controllers/supplementary/PackageInformationController.scala
```diff
@@ -31,31 +35,85 @@ import views.html.supplementary.package_information
...
+        cacheService
+          .cache[Seq[PackageInformation]](supplementaryCacheId(), formId, updatedPackagings)
+          .map(_ => Redirect(routes.PackageInformationController.displayForm()))
+      }
+      case _ => errorHandler.displayErrorPage()
     }
+
+  def continue()(implicit request: AuthenticatedRequest[AnyContent], packagings: Seq[PackageInformation]) = {
+    val payload = form.bindFromRequest()
+    if (payload.data.filter(_._2.size > 0).size > 1) badRequest(payload, USE_ADD)
```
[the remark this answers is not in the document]
raghu1978: this is the approach i would want every multiple items screen to adhere to.

app/controllers/supplementary/PackageInformationController.scala
```diff
@@ -31,31 +35,85 @@ import views.html.supplementary.package_information
...
+      }
+  }
+
+  def remove(
+    id: Option[String]
+  )(implicit authRequest: AuthenticatedRequest[AnyContent], packagings: Seq[PackageInformation]) =
+    id match {
+      case Some(id) => {
+        val updatedPackagings =
+          packagings.zipWithIndex.filterNot(_._2.toString == id).map(_._1)
```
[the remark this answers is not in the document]
raghu1978: its a tuple and would have the object and id in the tuple, i am using collection methods and there are no custom functions

app/controllers/supplementary/PackageInformationController.scala
```diff
@@ -31,31 +35,85 @@ import views.html.supplementary.package_information
...
+          isAdditionValid[PackageInformation](validForm).fold(
+            cacheService
+              .cache[Seq[PackageInformation]](supplementaryCacheId(), formId, (packagings :+ validForm))
+              .map(_ => Redirect(routes.PackageInformationController.displayForm()))
+          )(badRequest(form.fill(validForm), _))
         }
       )
-  }
+
+  private def isAdditionValid[A](item: A)(implicit cachedItems: Seq[A]): Option[String] =
```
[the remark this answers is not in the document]
raghu1978: cheers

app/controllers/supplementary/CommodityMeasureController.scala
```diff
@@ -0,0 +1,65 @@
...
+  def displayForm(): Action[AnyContent] = authenticate.async { implicit request =>
+    cacheService
+      .fetchAndGetEntry[Seq[PackageInformation]](supplementaryCacheId, formId)
+      .flatMap {
+        case Some(packages) =>
+          cacheService.fetchAndGetEntry[CommodityMeasure](supplementaryCacheId, commodityFormId).map {
+            case Some(data) => Ok(goods_measure(form.fill(data)))
+            case _          => Ok(goods_measure(form))
+          }
+        case _ => Future.successful(BadRequest(goods_measure(form.withGlobalError(ADD_ONE))))
```
PatrykRudnicki: Global error inside `displayForm`?

app/controllers/supplementary/PackageInformationController.scala
```diff
@@ -31,31 +35,85 @@ import views.html.supplementary.package_information
...
 
 class PackageInformationController @Inject()(
-  appConfig: AppConfig,
-  override val messagesApi: MessagesApi,
   authenticate: AuthAction,
-  customsCacheService: CustomsCacheService
-)(implicit ec: ExecutionContext)
+  errorHandler: ErrorHandler,
+  cacheService: CustomsCacheService
+)(implicit ec: ExecutionContext, appConfig: AppConfig, override val messagesApi: MessagesApi)
```
PatrykRudnicki: 👍

app/controllers/supplementary/PackageInformationController.scala
```diff
@@ -31,31 +35,85 @@ import views.html.supplementary.package_information
...
+      .map(items => Ok(package_information(form, items.getOrElse(Seq.empty))))
+
+  }
+
+  def submitForm(): Action[AnyContent] = authenticate.async { implicit authRequest =>
+    val actionTypeOpt = authRequest.body.asFormUrlEncoded.map(FormAction.fromUrlEncoded(_))
+    cacheService
+      .fetchAndGetEntry[Seq[PackageInformation]](supplementaryCacheId, formId)
+      .flatMap { data =>
+        implicit val packagings = data.getOrElse(Seq.empty)
```
PatrykRudnicki: I'm not sure about implicit data for packagings, but it's something about good practices only, so we can stay with this

app/views/supplementary/package_information.scala.html
raghu1978: this needs to be moved into a component so every screen can use, we will add a tech task to do that

app/controllers/supplementary/PackageInformationController.scala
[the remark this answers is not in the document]
PatrykRudnicki: I agree with Paulo, we can use filter and case to use naming for it, but in this case Raghu use zipWithIndex, so it's readable that this `_._2` will be index and after it `_._1` will be come back to packages.

Please to think about how to make it better, but for me, it's okay for now.

app/controllers/supplementary/PackageInformationController.scala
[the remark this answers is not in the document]
PatrykRudnicki: So @raghu1978 it'll be a problem, we decided to stay with the current implementation: `Save and continue` we had few meetings about it. I agree with Paulo, I want to have Save and Continue here.

Based on the implementation instead of using `_._2` `> 1` where you can't know from this code what do you want to filter etc to extract this logic.

```
private def isFormEmpty[A](form: Form[A]): Boolean =
    retrieveData(form).filter { case (_, value) => value.nonEmpty }.isEmpty

  private def retrieveData[A](form: Form[A]): Map[String, String] =
    form.data.filter { case (name, _) => name != "csrfToken" }
```

First question why `1`? For me it's not readable

app/models/viewmodels/HtmlTableRow.scala
```diff
@@ -24,7 +24,7 @@ object HtmlTableRow {
 
   def apply(label: String, value: String): HtmlTableRow = new HtmlTableRow(label, Seq(Some(value)))
 
-  def apply(label: String, value: Option[_]): HtmlTableRow = new HtmlTableRow(label, adjust(value))
+  //def apply(label: String, value: Option[_]): HtmlTableRow = new HtmlTableRow(label, adjust(value))
```
PatrykRudnicki: Please remove this comment

app/controllers/supplementary/PackageInformationController.scala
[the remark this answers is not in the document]
raghu1978: ok will change this

app/controllers/supplementary/PackageInformationController.scala
raghu1978: its making the packages available for all methods , if we have option there would be further retrieval of option in every method.

app/models/viewmodels/HtmlTableRow.scala
raghu1978: this is not my change but will update

app/views/supplementary/package_information.scala.html
PauloPortugal: Yes please!!!! :heart:

PR #137, opened 2019-03-12
Ceds fix link

NC-1234:


app/views/supplementary/confirmation_page.scala.html
```diff
@@ -35,7 +35,7 @@ <h1 class="heading-medium">@messages("supplementary.confirmation.whatHappensNext
 
     <p>
         @messages("supplementary.confirmation.explanation")
-        <a href="/customs-declare-exports/notifications/@{flash.get("ConversationId").getOrElse("")}">
+        <a href="/customs-declare-exports/submissions">
```
PatrykRudnicki: You removed only usage in this file of flash `conversationId` please remove passing it inside the controller

PR #138, opened 2019-03-12
CEDS-1002 Review Frontend tests

leszeklorek:


PatrykRudnicki:
Please solve conflicts...

ExporterDetailsPageControllerSpec l.33 change mapping to messages keys
CancelDeclarationControllerSpec l.44 not sure that we need to use \“Submit\” instead of Submit
I don't like parts like this: `getElementByCss(view, “form>div:nth-child(3)>label>span”)`

PR #145, opened 2019-03-14
CEDS-1006 Fixed error handling for radio-button elements

MaciejRewera:
This fix was introduced by non-meaningful error messages when user does
not select any option for radio button element, but this information is
mandatory.


app/forms/Choice.scala
```diff
@@ -16,22 +16,28 @@
...
 
 case class Choice(choice: String)
 
 object Choice {
   implicit val format = Json.format[Choice]
 
+  val choiceId = "Choice"
+
   import AllowedChoiceValues._
   private val correctChoice = Set(SupplementaryDec, StandardDec, Arrival, Departure, CancelDec)
```
PatrykRudnicki: If you can please change name to `correctChoices`

MaciejRewera: Yes, I can.

PR #149, opened 2019-03-14
CEDS-976 Add multiple transport containers

PauloPortugal:
**THIS IS WORK IN PROGRESS. DO NOT MERGE.** 

* **Remove the 'Container ID' input field from the `/transport-information` page**
This removes the text input field from the Transport Information page.
This was requested by @raghu1978. From now on, pages with single and
multiple items will be separated. Multiple items will be asked in a
separate page.

* **Route user to add-transport-containers page**
If the user has selected that he is using any container, then the
next page being presented will the /add-transport-caontainers where
we present the add to list pattern and the user can enter multiple
container ids.
If the user does not have any container to enter then the journey will
proceed to the /total-numbers-of-items page.

* **Provide ability to add multiple containers**
This is loosely following the pattern already in place

* **Grey out the 'Remove' and 'Add' buttons**
This is part of our progressive enhacemente on accessibility concerns.
The HMRC Front-End team will have to iterate on this as this might not
be the final solution, but it removes the confusion of having several
call to action green buttons on the same page.

### DIFF
--- a/app/controllers/supplementary/TransportInformationContainersPageController.scala
+++ b/app/controllers/supplementary/TransportInformationContainersPageController.scala
@@ -18,40 +18,114 @@ package controllers.supplementary
 
 import config.AppConfig
 import controllers.actions.AuthAction
+import controllers.supplementary.routes.{TotalNumberOfItemsController, TransportInformationContainersPageController}
 import controllers.util.CacheIdGenerator.supplementaryCacheId
+import controllers.util.MultipleItemsHelper.{add, remove, saveAndContinue}
+import controllers.util.{Add, FormAction, Remove, SaveAndContinue}
+import forms.supplementary.TransportInformationContainer
 import forms.supplementary.TransportInformationContainer.form
 import handlers.ErrorHandler
 import javax.inject.Inject
 import models.declaration.supplementary.TransportInformationContainerData
-import models.declaration.supplementary.TransportInformationContainerData.formId
+import models.declaration.supplementary.TransportInformationContainerData.{id, maxNumberOfItems}
+import models.requests.AuthenticatedRequest
+import play.api.data.Form
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.mvc.{Action, AnyContent}
 import services.CustomsCacheService
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import views.html.supplementary.add_transport_containers
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
 
 class TransportInformationContainersPageController @Inject()(
-  appConfig: AppConfig,
   override val messagesApi: MessagesApi,
   authenticate: AuthAction,
   errorHandler: ErrorHandler,
   customsCacheService: CustomsCacheService
-)(implicit ec: ExecutionContext)
+)(implicit ec: ExecutionContext, appConfig: AppConfig)
     extends FrontendController with I18nSupport {
 
   implicit val countries = services.Countries.allCountries
 
   def displayPage(): Action[AnyContent] = authenticate.async { implicit request =>
     customsCacheService
-      .fetchAndGetEntry[TransportInformationContainerData](supplementaryCacheId, formId)
+      .fetchAndGetEntry[TransportInformationContainerData](supplementaryCacheId, id)
       .map {
-        case Some(data) => Ok(add_transport_containers(appConfig, form, data.containers))
-        case _          => Ok(add_transport_containers(appConfig, form, Seq()))
+        case Some(data) => Ok(add_transport_containers(form, data.containers))
+        case _          => Ok(add_transport_containers(form, Seq()))
       }
   }
 
-  def submitTransportInformation(): Action[AnyContent] = ???
+  def handlePost(): Action[AnyContent] = authenticate.async { implicit request =>
+    val boundForm = form.bindFromRequest()
 
+    val actionTypeOpt = request.body.asFormUrlEncoded.map(FormAction.fromUrlEncoded(_))
+
+    val cachedData = customsCacheService
+      .fetchAndGetEntry[TransportInformationContainerData](supplementaryCacheId, id)
+      .map(_.getOrElse(TransportInformationContainerData(Seq())))
+
+    cachedData.flatMap { cache =>
+      actionTypeOpt match {
+        case Some(Add) => addContainer(boundForm, maxNumberOfItems, cache)
+
+        case Some(Remove(ids)) => removeContainer(cache, ids)
+
+        case Some(SaveAndContinue) => saveContainer(boundForm, maxNumberOfItems, cache)
+
+        case _ => errorHandler.displayErrorPage()
+      }
+    }
+  }
+
+  private def saveContainer(
+    boundForm: Form[TransportInformationContainer],
+    elementLimit: Int,
+    cache: TransportInformationContainerData
+  )(implicit request: AuthenticatedRequest[_], appConfig: AppConfig) =
+    saveAndContinue(boundForm, cache.containers, true, elementLimit).fold(
+      formWithErrors => Future.successful(BadRequest(add_transport_containers(formWithErrors, cache.containers))),
+      updatedCache =>
+        if (updatedCache != cache.containers)
+          customsCacheService
+            .cache[TransportInformationContainerData](
+              supplementaryCacheId,
+              id,
+              TransportInformationContainerData(updatedCache)
+            )
+            .map(_ => Redirect(TotalNumberOfItemsController.displayForm()))
+        else Future.successful(Redirect(TotalNumberOfItemsController.displayForm()))
+    )
+
+  private def removeContainer(cache: TransportInformationContainerData, ids: Seq[String])(
+    implicit request: AuthenticatedRequest[_]
+  ) = {
+    val updatedCache = remove(ids.headOption, cache.containers)
+
+    customsCacheService
+      .cache[TransportInformationContainerData](
+        supplementaryCacheId,
+        id,
+        TransportInformationContainerData(updatedCache)
+      )
+      .map(_ => Redirect(TransportInformationContainersPageController.displayPage()))
+  }
+
+  private def addContainer(
+    boundForm: Form[TransportInformationContainer],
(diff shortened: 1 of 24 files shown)


app/controllers/supplementary/TransportInformationContainersPageController.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    val actionTypeOpt = request.body.asFormUrlEncoded.map(FormAction.fromUrlEncoded(_))
+
+    val cachedData = customsCacheService
+      .fetchAndGetEntry[TransportInformationContainerData](supplementaryCacheId, id)
+      .map(_.getOrElse(TransportInformationContainerData(Seq())))
+
+    cachedData.flatMap { cache =>
+      actionTypeOpt match {
+        case Some(Add) => addContainer(boundForm, maxNumberOfItems, cache)
```
raghu1978: you can make the cache Implicit

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+        val body = Seq(("id", ""), saveAndContinueActionURLEncoded)
+
+        val result = route(app, postRequestFormUrlEncoded(uri, body: _*)).get
+        val stringResult = contentAsString(result)
+
+        status(result) must be(BAD_REQUEST)
+        stringResult must include(messages("supplementary.continue.mandatory"))
+      }
+
+      "try to add a longer id" in {
```
raghu1978: please move form based tests into forms and if possible use generators to test all permutation data.

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+
+        val body = Seq(("id", createRandomString(20)), addActionURLEncoded)
+        val result = route(app, postRequestFormUrlEncoded(uri, body: _*)).get
+
+        status(result) must be(BAD_REQUEST)
+        contentAsString(result) must include(messages("supplementary.transportInfo.containerId.error"))
+      }
+
+      "try to add non-alphanumeric characters" in {
```
raghu1978: applies same as above comment

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+        withCaching[TransportInformationContainerData](Some(cacheWithMaximumAmountOfHolders), id)
+
+        val body = Seq(("id", "M1l3s"), addActionURLEncoded)
+        val result = route(app, postRequestFormUrlEncoded(uri, body: _*)).get
+
+        status(result) must be(BAD_REQUEST)
+        contentAsString(result) must include(messages("supplementary.limit"))
+      }
+
+      "try to save more than 9999 containers" in {
```
raghu1978: what's the difference between this and the above test in 182?

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+
+      status(result) must be(SEE_OTHER)
+      header.headers.get("Location") must be(
+        Some("/customs-declare-exports/declaration/supplementary/total-numbers-of-items")
+      )
+    }
+  }
+}
+
+object TransportInformationContainersPageControllerSpec {
```
raghu1978: why do you need an object creation in here?

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
PatrykRudnicki: First we need to discuss that we want to use generators. We don't have decision about it and you can . find some people that don't want to use it.

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
PatrykRudnicki: Companion object is a good place to store some constant data. We started using it for tests, so whole constants you will have in one place.

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
raghu1978: regardless the tests needs to go into forms specific specs

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
raghu1978: might be worth companion objects for tests at one place we will to be aware of the re-use of they are really created for that purpose.

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
PauloPortugal: One `adds` and the other `saves and continues`

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
PauloPortugal: If they live in the same class then we should be able to check if some has been created already.

app/controllers/supplementary/TransportInformationContainersPageController.scala
PauloPortugal: I actually like having the `cachedData` as an explicity variable as this is business logic. 
I like the use of `implicit` for `currying` or to pass boiler plate parameters (Play's `request`, HMRC's middleware, app config, etc), something that does not immediately translates as business logic, but more as app/infrastructure dependencies, something that a developer should not need to worry as such, as they are always there regardless of the logic needs/requirements.

PauloPortugal: @raghu1978 What do you think?

raghu1978: in this case it is implicit that you need the collection and to pass it across you still are aware in the code as every method has that as an implicit parmater, there is no difference in how we pass header carrier and execution context

PauloPortugal: For example https://github.com/hmrc/customs-declare-exports-frontend/blob/master/test/generators/Generators.scala#L107 this is something I dislike.
1. It is something that is only going to be used by the one `form`  but it is instead in a generic Generators class.
2. It is `implicit` when there is not need to such and makes life of the tester/developer a bit harder when could have well been defined in the test method or as a function on the companion spec object.

PauloPortugal: >  there is no difference in how we pass header carrier and execution context
None at all, the difference is not how we pass it, but how we use them. The `request`, `appConfig`, `headerCarrier`, `format`, `executionContext` it makes some sense to pass them as implicit variables (the same with Authorization/Middleware libraries/functions or as I said before `currying`), although for example I would think twice about passing `countries` as an implicit variable.

The somewhat obsession of writing less code, does not translate into more practical or readable code, it just becomes more _enigmatic_. The `cachedData` is a variable that holds a state that is  useuful to _Transport Containers_ business logic, so I would like to continue to explicitly using it.

Another good justification to use implicit variables could be because the signature of the function is now too verbose or too long which is also not the case.

In EffectiveScala they seems to corroborate with me http://twitter.github.io/effectivescala/#Types%20and%20Generics-Implicits and I have also brought up this discussion in the #community-scala and people do not strongly disagree with it.

I hope that this `implicit` use does not block the PR.

PatrykRudnicki: Cache shouldn't be passed as implicit. I told you this in your last PR or during talking with you. Countries is something that always is constant. In my opinion it can be passed as implicit. Cache shouldn't. It's my opinion.

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
PatrykRudnicki: Companion object should be close to the class. We shouldn't move it from this file. It's not an object to use by all classes. It must contain only stuff for this test.

test/controllers/supplementary/TransportInformationContainersPageControllerSpec.scala
PauloPortugal: @PatrykRudnicki has a good point on the generators discussion. 
Before I say _yes_ or _no_ to something I normally like to experiment with it.
So I have introduced Generators following loosely what @raghu1978  has made, with the following changes:
1. Unfortunately the `Arbitrary` library expects implicit variables, so I have defined the function on the test method, this way it becomes clearer what we are doing, instead of making our `generator.Generators` a dumping ground of our CDS generative functions.
2. I _have tried_ to make a sensible use of them. After pairing with @MaciejRewera I actually got one previously green test to to red when it shouldn't, i.e, the data generated was incorrect. We are now comfortable that we have fixed this issue :)
3. I actually found the use of Generators interesting but over-complicated for my scenario.

app/controllers/supplementary/TransportInformationPageController.scala
```diff
@@ -46,21 +48,26 @@ class TransportInformationPageController @Inject()(
...
-        case Some(data) => Ok(transport_information(appConfig, TransportInformation.form.fill(data)))
-        case _          => Ok(transport_information(appConfig, TransportInformation.form))
+        case Some(data) => Ok(transport_information(appConfig, form.fill(data)))
+        case _          => Ok(transport_information(appConfig, form))
       }
   }
 
   def submitTransportInformation(): Action[AnyContent] = authenticate.async { implicit request =>
-    TransportInformation.form
+    form
```
PatrykRudnicki: 👍

app/models/declaration/supplementary/TransportInformationContainerData.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def containsItem(container: TransportInformationContainer): Boolean = containers.contains(container)
+}
+
+object TransportInformationContainerData {
+  implicit val format = Json.format[TransportInformationContainerData]
+
+  val id = "TransportInformationContainerData"
+
+  val maxNumberOfItems = 9999
```
PatrykRudnicki: 😂

PR #151, opened 2019-03-14
CEDS-992 Add mandatory info validation before displaying Summary page

MaciejRewera:
If the user does not provide information we consider mandatory, they
should not be able to access Summary page. Currently it is LRN only.

### DIFF
--- a/app/controllers/supplementary/SummaryPageController.scala
+++ b/app/controllers/supplementary/SummaryPageController.scala
@@ -20,6 +20,7 @@ import config.AppConfig
 import connectors.{CustomsDeclarationsConnector, CustomsDeclareExportsConnector}
 import controllers.actions.AuthAction
 import controllers.util.CacheIdGenerator.supplementaryCacheId
+import forms.supplementary.ConsignmentReferences
 import handlers.ErrorHandler
 import javax.inject.Inject
 import metrics.ExportsMetrics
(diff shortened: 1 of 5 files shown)


app/controllers/supplementary/SummaryPageController.scala
```diff
@@ -51,11 +53,16 @@ class SummaryPageController @Inject()(
 
   def displayPage(): Action[AnyContent] = authenticate.async { implicit request =>
     customsCacheService.fetch(supplementaryCacheId).map {
-      case Some(cacheMap) => Ok(summary_page(appConfig, SupplementaryDeclarationData(cacheMap)))
-      case None           => Ok(summary_page_no_data(appConfig))
+      case Some(cacheMap) if containsMandatoryData(cacheMap) =>
+        Ok(summary_page(appConfig, SupplementaryDeclarationData(cacheMap)))
+      case Some(_) => Ok(summary_page_no_data(appConfig))
+      case None    => Ok(summary_page_no_data(appConfig))
```
PatrykRudnicki: Instead of two cases you can use one with `_`.

We start moving appConfig as implicit one, so if you want you can move it too, using views are shorter :)

MaciejRewera: I've changed both.

app/controllers/supplementary/SummaryPageController.scala
```diff
@@ -49,13 +51,18 @@ class SummaryPageController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController with I18nSupport {
 
+  implicit val appConfigImpl: AppConfig = appConfig
```
PatrykRudnicki: I mean to make this app conf implcit like ec

`(implicit ec: ExecutionContext, appConfig: AppConfig)`

PatrykRudnicki: And remove from 42 line

PR #156, opened 2019-03-20
Ceds 919

raghu1978:
CEDS-919 looping of item level multiple items and moving screen navigation, adding an item journey using temporary cache id.


app/controllers/supplementary/CommodityMeasureController.scala
```diff
@@ -41,10 +41,10 @@ class CommodityMeasureController @Inject()(
 
   def displayForm(): Action[AnyContent] = authenticate.async { implicit request =>
     cacheService
-      .fetchAndGetEntry[Seq[PackageInformation]](supplementaryCacheId, formId)
+      .fetchAndGetEntry[Seq[PackageInformation]](goodsItemCacheId, formId)
       .flatMap {
-        case Some(_) =>
-          cacheService.fetchAndGetEntry[CommodityMeasure](supplementaryCacheId, commodityFormId).map {
+        case Some(packages) =>
```
PatrykRudnicki: You don't use packages here. Please use `_`

app/controllers/supplementary/DocumentsProducedController.scala
```diff
@@ -141,7 +155,7 @@ class DocumentsProducedController @Inject()(
     if (cachedData.containsItem(docToRemove)) {
       val updatedCache = cachedData.copy(documents = cachedData.documents.filterNot(_ == docToRemove))
 
-      customsCacheService.cache[DocumentsProducedData](supplementaryCacheId, formId, updatedCache).map { _ =>
+      customsCacheService.cache[DocumentsProducedData](goodsItemCacheId, formId, updatedCache).map { _ =>
```
PatrykRudnicki: In every place you use `customsCacheService` but in add goods item you use this `items` one. Probably everything in this file is okay, but I want to understand how data from this screen is handled? I mean for me when we discussed about separate cache I thought we will be storing all item data in separate cache and when use finished adding item we will add this to the main cache clear item cache and give a possibility to the user to go through the journey once again.

app/controllers/supplementary/ItemsSummaryController.scala
```diff
@@ -0,0 +1,46 @@
...
+class ItemsSummaryController @Inject()(
+  authenticate: AuthAction,
+  errorHandler: ErrorHandler,
+  cacheService: CustomsCacheService
+)(implicit ec: ExecutionContext, appConfig: AppConfig, override val messagesApi: MessagesApi)
+    extends FrontendController with I18nSupport {
+
+  def displayForm(): Action[AnyContent] = authenticate.async { implicit request =>
+    cacheService
+      .fetchAndGetEntry[Seq[GovernmentAgencyGoodsItem]](supplementaryCacheId, itemsId)
```
PatrykRudnicki: My propose is to wrap this type `Seq[GovernmentAgencyGoodsItem]` inside some class. Don't use Sequence for cache. But it's my personal thought.

app/models/DeclarationFormats.scala
```diff
@@ -71,4 +72,6 @@ object DeclarationFormats {
   implicit val loadingLocationFormats = Json.format[LoadingLocation]
 
   implicit val warehouseFormats = Json.format[Warehouse]
-}
+
+  implicit val governmentAgencyGoodsItemFormats = Jsonx.formatCaseClass[GovernmentAgencyGoodsItem]
```
PatrykRudnicki: Why you need this? I mean whole `Jsonx`?

project/AppDependencies.scala
```diff
@@ -19,7 +19,8 @@ object AppDependencies {
     "uk.gov.hmrc" %% "play-language" % "3.4.0",
     "com.thoughtworks.xstream" % "xstream" % "1.4.10",
     "uk.gov.hmrc" %% "wco-dec" % "0.23.0",
-    "com.fasterxml.jackson.core" % "jackson-core" % jacksonVersion,
+    "ai.x"         %% "play-json-extensions" % "0.9.0",
```
PatrykRudnicki: In my opinion we shouldn't use this and remove whole jackson dependencies.

app/controllers/supplementary/DocumentsProducedController.scala
raghu1978: The goodsItemCacheId is the temp cache for adding item elements. if you looks at itemsCache you will get an understanding

app/controllers/supplementary/ItemsSummaryController.scala
raghu1978: we have already complicated these with too many classes and these are stored in database without an issue we should have it as is.

app/models/DeclarationFormats.scala
raghu1978: this is the case class which has 21+ fields and  I sued the library to handle implicit Json formats

project/AppDependencies.scala
raghu1978: the library allows has to simplify son formats , I tried to not use and it is not that easy to get away from wco-dec as of now

app/services/ItemsCachingService.scala
```diff
@@ -0,0 +1,178 @@
...
+            )
+        )
+      )
+
+  def additionalInfo(cachedData: CacheMap) =
+    cachedData
+      .getEntry[AdditionalInformationData](AdditionalInformationData.formId)
+      .map(_.items.map(info => AdditionalInformation(Some(info.code), Some(info.description))))
+
+  def documents(cachedData: CacheMap) =
```
PatrykRudnicki: Please add return type to all methods in this file

app/services/ItemsCachingService.scala
```diff
@@ -0,0 +1,178 @@
...
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.http.cache.client.CacheMap
+import uk.gov.hmrc.wco.dec._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class ItemsCachingService @Inject()(cacheService: CustomsCacheService)(appConfig: AppConfig) {
+
+  def generatePackages(cachedData: CacheMap) =
```
MaciejRewera: There are some very 'suspicious' return types in these methods (like `Option[Seq[_]]`).
I think this is because of nested `map` methods, while there should be `flatMap`.

app/services/ItemsCachingService.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  private def createGoodsItem(seq: Int, cachedData: CacheMap) = {
+    val itemTypeData = goodsItemFromItemTypes(cachedData)
+    val commodity = itemTypeData.fold(Commodity())(_.commodity.getOrElse(Commodity()))
+    val updatedCommodity = commodity.copy(goodsMeasure = commodityGoodsMeasure(cachedData).flatMap(_.goodsMeasure))
+
+    GovernmentAgencyGoodsItem(
+      sequenceNumeric = seq,
+      statisticalValueAmount = itemTypeData.flatMap(_.statisticalValueAmount),
+      packagings = generatePackages(cachedData).getOrElse(Seq.empty),
```
MaciejRewera: You can avoid `getOrElse(Seq.empty)` if you solve return types for methods.
It makes unnecessary noise in the code.

app/services/ItemsCachingService.scala
```diff
@@ -0,0 +1,178 @@
...
+import models.DeclarationFormats._
+import models.declaration.supplementary.{AdditionalInformationData, DocumentsProducedData, ProcedureCodesData}
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.http.cache.client.CacheMap
+import uk.gov.hmrc.wco.dec._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@Singleton
+class ItemsCachingService @Inject()(cacheService: CustomsCacheService)(appConfig: AppConfig) {
```
MaciejRewera: I would advise to clean the code in this entire file. It is hard to read and to reason what is going on here, for e.g. splitting methods into smaller ones could be the first step.

app/models/DeclarationFormats.scala
MaciejRewera: IMHO we should raise a problem on data model once again. We continue to be highly dependent on wco-dec repository, while we should be loosely coupled to anything from outside. We could be more independent with our logic as well.

app/controllers/supplementary/ItemsSummaryController.scala
PatrykRudnicki: Yeah, I agree, but `Seq[...]` type is not a good choice, I don't know. Just comment

app/models/DeclarationFormats.scala
PatrykRudnicki: I think that we should get rid of those classes, but I understand that for know probably only this `Jsonx` will solve a problem. Just we need to start preparing model with representation of our declaration, not `wco-dec`

app/controllers/util/CacheIdGenerator.scala
```diff
@@ -19,11 +19,15 @@ import models.requests.AuthenticatedRequest
 
 object CacheIdGenerator {
 
+  val itemsId = "exportItems"
```
PatrykRudnicki: Please move it out of `CacheIdGenerator`. It's generator, not a place for all `IDs` I think better to put it inside the controller/model.

app/services/ItemsCachingService.scala
```diff
@@ -0,0 +1,179 @@
...
+    val updatedCommodity = commodity.copy(goodsMeasure = commodityGoodsMeasure(cachedData).flatMap(_.goodsMeasure))
+
+    GovernmentAgencyGoodsItem(
+      sequenceNumeric = seq + 1,
+      statisticalValueAmount = itemTypeData.flatMap(_.statisticalValueAmount),
+      packagings = generatePackages(cachedData).getOrElse(Seq.empty),
+      governmentProcedures = procedureCodes(cachedData).getOrElse(Seq.empty),
+      commodity = Some(updatedCommodity),
+      additionalInformations = additionalInfo(cachedData).getOrElse(Seq.empty),
+      additionalDocuments = documents(cachedData).getOrElse(Seq.empty)
```
PatrykRudnicki: I think that better place for `getOrElse` is to use it inside the method and return `Seq` from every method instead of `Option[Seq[...]]`

app/services/ItemsCachingService.scala
```diff
@@ -0,0 +1,179 @@
...
+          .fetchAndGetEntry[Seq[GovernmentAgencyGoodsItem]](supplementaryCacheId, itemsId)
+          .flatMap(
+            items =>
+              cacheService
+                .cache[Seq[GovernmentAgencyGoodsItem]](
+                  supplementaryCacheId,
+                  itemsId,
+                  items.getOrElse(Seq.empty) :+ createGoodsItem(items.fold(0)(_.size), cachedData)
+                )
+                .flatMap(_ => cacheService.remove(goodsItemCacheId).map(_.status == 204))
```
PatrykRudnicki: Please use `NO_CONTENT` instead of 204

app/services/ItemsCachingService.scala
```diff
@@ -0,0 +1,179 @@
...
+}
+
+object ExportsItemsCacheIds {
+  val itemsCachePages = Map(
+    PackageInformation.formId -> "package",
+    ProcedureCodesData.formId -> "procedure codes",
+    CommodityMeasure.commodityFormId -> "goods measure",
+    DocumentsProducedData.formId -> "additional documents",
+    ItemType.id -> "Item type",
+    AdditionalInformationData.formId -> "additional information"
```
PatrykRudnicki: First, here is a good place to move ID from `CacheIdGenerator`.
Second, please use one approach start with capital letter or lower case

test/base/TestHelper.scala
```diff
@@ -16,10 +16,31 @@
...
+import uk.gov.hmrc.http.cache.client.CacheMap
+
+import scala.annotation.tailrec
 import scala.util.Random
 
 object TestHelper {
 
   def createRandomString(length: Int): String = Random.alphanumeric.take(length).mkString
 
+  def getDataSeq[A](n: Int, elem: A) = {
```
PatrykRudnicki: Please add a return type to this method

test/base/TestHelper.scala
```diff
@@ -16,10 +16,31 @@
...
+      if (n == 0) {
+        seq
+      } else {
+        loop(n - 1, seq :+ elem)
+      }
+    loop(n, Seq.empty)
+  }
+
+
+  def getCacheMap[A](data: A, formId: String)(implicit writes: Writes[A]) =
```
PatrykRudnicki: return type

app/services/ItemsCachingService.scala
raghu1978: this is the easier version, the issue is with the way the screens are designed and how eco-dec model is and that is complicated I spent quite a time on this and handling this is not going to be easy and more over the designs are going to change in future and so as this class.

app/services/ItemsCachingService.scala
raghu1978: the getOrElse are there to understand the else scenario to later add validation in all those circumstances where it is None

app/services/ItemsCachingService.scala
raghu1978: We need to handle no data scenario and I have this so we can update these methods with a common error scenario

app/models/DeclarationFormats.scala
raghu1978: we will raise a tech debt story once we have a view on the most of declaration types until then ItemCachingService will serve a place for all complicated transformation.

app/controllers/supplementary/SummaryPageController.scala
```diff
@@ -118,4 +118,19 @@ class SummaryPageController @Inject()(
   private def prepareFlashScope(lrn: String) =
     Flash(Map("LRN" -> lrn))
 
+  //TODO : refactor to handle large data collection and move the logic of metadata creation to separate service
+  private def createMetaData(cacheMap: CacheMap, suppDecData: SupplementaryDeclarationData): MetaData = {
```
MaciejRewera: This method is not tested anywhere despite being crucial for the whole journey.
Moving it to a separate class/object is a good idea and it will be easier to test.

app/controllers/supplementary/SummaryPageController.scala
```diff
@@ -68,9 +69,8 @@ class SummaryPageController @Inject()(
       case Some(cacheMap) =>
         exportsMetrics.startTimer(submissionMetric)
         val suppDecData = SupplementaryDeclarationData(cacheMap)
```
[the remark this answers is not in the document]
raghu1978: we can't as suppDecData is used later             val ducr = suppDecData.consignmentReferences.flatMap(_.ducr)

test/controllers/supplementary/SummaryPageControllerSpec.scala
```diff
@@ -156,42 +154,9 @@ class SummaryPageControllerSpec extends CustomExportsBaseSpec {
...
-
-        resultAsString must include(messages("supplementary.summary.additionalInformation.header"))
-        resultAsString must include(messages("supplementary.summary.additionalDocumentation.header"))
-        resultAsString must include(messages("supplementary.summary.additionalDocumentation.documentTypeCode"))
-        resultAsString must include(messages("supplementary.summary.additionalDocumentation.documentId"))
-        resultAsString must include(messages("supplementary.summary.additionalDocumentation.documentPart"))
-        resultAsString must include(messages("supplementary.summary.additionalDocumentation.documentStatus"))
-        resultAsString must include(messages("supplementary.summary.additionalDocumentation.documentStatusReason"))
-      }
-
```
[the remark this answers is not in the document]
raghu1978: these are covered by  ItemsCachingService as the data has already been transformed into the wco-dec model in general we need a high level tests once we move away from the whole mapping

test/forms/supplementary/AdditionalInformationSpec.scala
```diff
@@ -1,69 +0,0 @@
...
-
-  "Additional Information object" should {
-    "contains correct limit value" in {
-      AdditionalInformationData.maxNumberOfItems must be(99)
-    }
-  }
-
-}
-
-object AdditionalInformationSpec {
```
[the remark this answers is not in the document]
raghu1978: I didn't see this used anywhere

app/controllers/supplementary/SummaryPageController.scala
raghu1978: agree

app/controllers/util/CacheIdGenerator.scala
```diff
@@ -19,11 +19,15 @@ import models.requests.AuthenticatedRequest
...
   def eoriCacheId()(implicit request: AuthenticatedRequest[_]): String = request.user.eori
 
   def supplementaryCacheId()(implicit request: AuthenticatedRequest[_]): String =
     s"supplementary-${request.user.eori}"
 
+  def goodsItemCacheId()(implicit request: AuthenticatedRequest[_]): String =
```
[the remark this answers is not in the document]
raghu1978: as discussed it should be specific to EORI as discussed.

app/controllers/supplementary/SummaryPageController.scala
raghu1978: I will have to refactor this file based on the changes that happen with CEDS-1009 I started working on it and will raise a PR

PR #163, opened 2019-03-25
CEDS-1063 Use JourneyAction in all of our declaration journeys

PauloPortugal:
I made the decision not to use it on the `Movements`, `Cancel` and `Notifications`
journeys, as it is not needed there.

### DIFF
--- a/app/controllers/declaration/DeclarantDetailsPageController.scala
+++ b/app/controllers/declaration/DeclarantDetailsPageController.scala
@@ -17,8 +17,8 @@
 package controllers.declaration
 
 import config.AppConfig
-import controllers.actions.AuthAction
-import controllers.util.CacheIdGenerator.supplementaryCacheId
+import controllers.actions.{AuthAction, JourneyAction}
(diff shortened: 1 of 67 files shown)


app/controllers/declaration/DeclarantDetailsPageController.scala
```diff
@@ -33,28 +33,28 @@ import scala.concurrent.{ExecutionContext, Future}
 class DeclarantDetailsPageController @Inject()(
   appConfig: AppConfig,
   override val messagesApi: MessagesApi,
-  authenticate: AuthAction,
+  authenticate: AuthAction, journeyType: JourneyAction,
```
PatrykRudnicki: Yeah, like Maciej said, please run formatter. This `journeyType` should be in new line

PauloPortugal: Yes... Well spotted. Done! :)

PauloPortugal:
Going in. Fire in the hole!

PR #165, opened 2019-03-26
[CEDS-919] changing Quantity to Bigdecimal and handling the same in I…

raghu1978:
…tems caching.


PauloPortugal:
I am OK with the implementation :+1: 

I am just a bit puzzled by this. So we can have 2.5 or 2.123124 produced documents?
I know that this is not related to your story, but where is the requirement to handle a decimal  quantity?

And why use `BigDecimal` for validation. Sounds a bit overcomplicated, as we are now making some assumptions, such as precision (16). Again what are the requirements for this?

Something like this would suffice
```
def isDecimal(str :String): Boolean = str.forall(c => Character.isDigit(c) || c == '.')
```

raghu1978:
@MaciejRewera we need to refactor to use BigDecimal or Int in case it is a number that are out of the box from play-framework forms so we dont have any custom code and additional tests around it, also with the approach we can handle different errors precisely on the failure, once we have the content for errors it will be evident.

raghu1978:
@PauloPortugal i believe we have discussed and agreed the reason with using BigDecimal.

PR #169, opened 2019-03-27
CEDS-1024 Add 3/31, 3/32 Carrier details to the full journey

PauloPortugal:
* Move `ConsigneeDetailsViewSpec` inside `'test/views/declaration/'`
 * Add 3/31, 3/32 Carrier details to the full journey
Implement the `toMetadataProperties` method
Update the view, form, controller and model

### DIFF
--- a/app/controllers/declaration/CarrierDetailsPageController.scala
+++ b/app/controllers/declaration/CarrierDetailsPageController.scala
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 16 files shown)


app/controllers/declaration/CarrierDetailsPageController.scala
```diff
@@ -0,0 +1,66 @@
...
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+import views.html.declaration.carrier_details
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * This controller is not used in supp dec journey
+  */
+class CarrierDetailsPageController @Inject()(
+  appConfig: AppConfig,
```
MaciejRewera: Could you, please, make the appConfig an implicit?

test/controllers/declaration/RepresentativeDetailsPageControllerSpec.scala
```diff
@@ -158,7 +162,17 @@ class RepresentativeDetailsPageControllerSpec extends CustomExportsBaseSpec with
       header.headers.get("Location") must be(Some("/customs-declare-exports/declaration/additional-actors"))
     }
 
-    "accept form with status and address only" in {
+    "accept form with status and EORI if on standard journey" in {
```
MaciejRewera: Here you can use the format:
`"accept form with status and EORI" when {
    "on standard journey" in {...}
    "on supplementary journey" in {...}
}`

app/controllers/declaration/CarrierDetailsPageController.scala
PauloPortugal: :man_facepalming: Forgot about that!!!!!

PR #172, opened 2019-03-28
[CEDS-919] Data mapping update to allow CUS code only when available

raghu1978:


app/services/ItemsCachingService.scala
```diff
@@ -108,9 +108,8 @@ class ItemsCachingService @Inject()(cacheService: CustomsCacheService)(appConfig
                   Classification(
                     Some(item.combinedNomenclatureCode),
                     identificationTypeCode = Some(CombinedNomenclatureCode)
-                  ),
-                  Classification(item.cusCode, identificationTypeCode = Some(CUSCode))
-                ) ++
+                  )
+                ) ++  item.cusCode.map(id => Classification(Some(id), identificationTypeCode = Some(CUSCode))) ++
```
PatrykRudnicki: Hmm, based on this change I understand that you need to build Classification model only when cusCode is defined?

Because I don't see sense for mapping Option and after that wrapping this value once again inside the `Some`.
The same is below, so probably this is approach that you want to achieve but in my opinion we should build models like this using method where you will use something like this

`if(item.cusCode.isDefined) Classification(item.cusCode, identificationTypeCode = Some(CUSCode)) else None` or something that will be more readable from developer perspective. For me it takes some time to understand what happen here and why it looks like this.

raghu1978: agree we can update these further as we are evolving with mapping further and further.

PR #175, opened 2019-04-02
CEDS-1055: fixed save of movement submission (saving dummy values) an…

mattclark-zerogravit:
…d changed config ready for switch to new backend

### DIFF
--- a/app/controllers/movement/MovementSummaryController.scala
+++ b/app/controllers/movement/MovementSummaryController.scala
@@ -16,14 +16,17 @@
 
 package controllers.movement
 
+import java.util.UUID
+
 import config.AppConfig
-import connectors.CustomsInventoryLinkingExportsConnector
+import connectors.{CustomsDeclareExportsConnector, CustomsDeclareExportsMovementsConnector, CustomsInventoryLinkingExportsConnector}
 import controllers.actions.AuthAction
 import controllers.util.CacheIdGenerator.movementCacheId
 import forms.inventorylinking.MovementRequestSummaryMappingProvider
 import handlers.ErrorHandler
 import javax.inject.Inject
 import metrics.{ExportsMetrics, MetricIdentifiers}
+import models.MovementSubmission
 import play.api.Logger
 import play.api.data.Form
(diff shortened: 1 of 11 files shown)


app/controllers/movement/MovementSummaryController.scala
```diff
@@ -57,26 +61,42 @@ class MovementSummaryController @Inject()(
...
         exportsMetrics.startTimer(metricIdentifier)
 
         customsInventoryLinkingExportsConnector
           .sendMovementRequest(movementCacheId, data.toXml)
-          .map {
-            case accepted if accepted.status == ACCEPTED =>
-              exportsMetrics.incrementCounter(metricIdentifier)
-              Redirect(controllers.movement.routes.MovementSummaryController.displayConfirmation())
+          .flatMap {
+            case accepted if accepted.status == ACCEPTED => {
```
PatrykRudnicki: My propose to change the name to `response`.

It's easier to understand `response.status == ACCEPTED` than `accepted.status ...`

app/controllers/movement/MovementSummaryController.scala
```diff
@@ -41,6 +44,7 @@ class MovementSummaryController @Inject()(
   authenticate: AuthAction,
   errorHandler: ErrorHandler,
   customsCacheService: CustomsCacheService,
+  customsDeclareExportsConnector: CustomsDeclareExportsConnector,
```
PatrykRudnicki: My propose is to create `CustomsDeclareExportsMovementConnector` and not use connector for backend, because we're not connecting through the backend to movements, but directly to movements.

app/controllers/movement/MovementSummaryController.scala
mattclark-zerogravit: This code was already there I will change as I agree it is easier to read

app/controllers/movement/MovementSummaryController.scala
mattclark-zerogravit: have changed this

PR #176, opened 2019-04-02
CEDS-1025 Add countries of routing to the standard declaration journey

PatrykRudnicki:


app/controllers/declaration/DestinationCountriesController.scala
```diff
@@ -51,25 +51,11 @@ class DestinationCountriesController @Inject()(
...
-  )(implicit request: JourneyRequest[AnyContent], hc: HeaderCarrier): Future[Html] =
-    decType match {
+    (request.choice.value match {
       case SupplementaryDec => handleSupplementary()
       case StandardDec      => handleStandard()
-    }
+    }).map(resultView => Ok(resultView))
+  }
 
   private def handleSupplementary()(implicit request: JourneyRequest[AnyContent], hc: HeaderCarrier): Future[Html] =
```
MaciejRewera: Maybe `displayFormSupplementary` would be better? More clear about the purpose of it.

app/forms/declaration/destinationCountries/DestinationCountries.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package forms.declaration.destinationCountries
+
+import play.api.data.Form
+
+object DestinationCountries {
+  val formId = "DestinationCountries"
+
+  val limit = 99
```
MaciejRewera: This `val` is related to standard journey only. Better move it to corresponding object, not the general one.
And the name is too general. Maybe `coutriesOfRoutingMaxAmount` ?

app/forms/declaration/destinationCountries/DestinationCountries.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def supplementaryForm(): Form[DestinationCountriesSupplementary] = Form(DestinationCountriesSupplementary.mapping)
```
MaciejRewera: I think that these methods should be placed in proper companion objects as well. The usage is not intuitive.

app/forms/declaration/destinationCountries/DestinationCountries.scala
PatrykRudnicki: I'll stay with this object because it will be difficult to use standardForm and supplementaryForm or formId easy inside the controller

app/forms/declaration/destinationCountries/DestinationCountries.scala
PatrykRudnicki: As above

PR #177, opened 2019-04-02
[CEDS-1090] Data mapping for previous documents

raghu1978:


app/views/declaration/transport_information.scala.html
```diff
@@ -37,6 +37,10 @@
 
         @components.page_title(Some("supplementary.transportInfo.title"), None)
 
+
+
+@helper.select(field = form("inlandModeOfTransportCode"), options = Seq("foo"->"Foo", "bar"->"Bar"))
```
PatrykRudnicki: this Foo and Bar should be here or you use it only for testing?

raghu1978: will remove, I was using it for testing.

PR #182, opened 2019-04-04
CEDS-1051 Supplementary declaration content update

PauloPortugal:


app/views/declaration/office_of_exit.scala.html
```diff
@@ -30,11 +30,13 @@
...
         @components.error_summary(form.errors)
 
+        @components.section_header(messages("supplementary.summary.locations.header"))
+
         @components.page_title(Some("supplementary.officeOfExit.title"))
 
         @components.input_text(
             field = form("officeId"),
-            label = messages("supplementary.officeOfExit"),
+            label = "",
```
PatrykRudnicki: Just to be sure. Do you want to remove label?

app/views/declaration/supervising_office.scala.html
```diff
@@ -30,11 +30,13 @@
...
         @components.error_summary(form.errors)
 
+        @components.section_header(messages("supplementary.summary.locations.header"))
+
         @components.page_title(Some("supplementary.supervisingCustomsOffice.title"))
 
         @components.input_text(
             field = form("supervisingCustomsOffice"),
-            label = messages("supplementary.supervisingCustomsOffice"),
+            label = "",
```
PatrykRudnicki: The same question here

app/views/declaration/office_of_exit.scala.html
PauloPortugal: Yes. We have a hint, but no longer a label

app/views/declaration/supervising_office.scala.html
PauloPortugal: Yes same thing here. No label, but we have a hint.

PR #187, opened 2019-04-08
 CEDS-1097 Show consignee details for Standard journeys

PauloPortugal:
* Show consignee details for Standard journeys
* Add consignee details to the summary page
* Supplementary declaration content change


node_modules/.staging/help-fd0a0cbe/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
PatrykRudnicki: Why we need this? What's this?

The same question about `.npmignore` above

PauloPortugal: Oh God!!!! this is a `Node.js` screw up!!! Thanks for noticing it!

PauloPortugal: I have fixed this!!! :)

PR #188, opened 2019-04-10
CEDS-1033: removed movement specific code from frontend

mattclark-zerogravit:


app/models/viewmodels/HtmlTableRow.scala
```diff
@@ -22,16 +22,8 @@ class HtmlTableRow(val label: String, val values: Seq[Option[String]])
 
 object HtmlTableRow {
 
-  def apply(label: String, value: String): HtmlTableRow = new HtmlTableRow(label, Seq(Some(value)))
```
PatrykRudnicki: Why you removed it?

app/models/viewmodels/HtmlTableRow.scala
```diff
@@ -22,16 +22,8 @@ class HtmlTableRow(val label: String, val values: Seq[Option[String]])
...
-
   def apply(label: String, value: Iterable[_]): HtmlTableRow = new HtmlTableRow(label, adjust(value))
 
-  private def adjust(option: Option[_]): Seq[Option[String]] = option match {
```
PatrykRudnicki: And I'm not sure about removing this

app/controllers/ChoiceController.scala
```diff
@@ -59,7 +59,6 @@ class ChoiceController @Inject()(
             validChoice.value match {
               case SupplementaryDec | StandardDec =>
                 Redirect(controllers.declaration.routes.DispatchLocationPageController.displayPage())
-              case Arrival | Departure => Redirect(controllers.movement.routes.MovementController.displayDucrPage())
```
PatrykRudnicki: Hmm, so now when you choose arrival or departure the page will refresh. Do we have ticket to redirect to another microservice?

Because view still contains those radio buttons

app/models/viewmodels/HtmlTableRow.scala
mattclark-zerogravit: Not being used... so why keep it?

app/controllers/ChoiceController.scala
mattclark-zerogravit: Ah need to change the view... to remove those options... will do this

PR #191, opened 2019-04-10
CEDS-1051 Get HMTL elements by ID when doing content tests

PauloPortugal:
* **CEDS-1051** - This replaces the `getElementByCSS` with `getElementById` ont the cases
we have deep nested elements in our pages. Does not address the deeply nested table elements
* **CEDS-1042** - Correct Not Eligible view page title


leszeklorek:
Look much better

app/views/declaration/not_eligible.scala.html
```diff
@@ -22,7 +22,7 @@
 @(appConfig: AppConfig)(implicit request: Request[_], messages: Messages)
 
 @main_template(
-title = messages("notEligible.title"),
+title = messages("notEligible.title") + " " + messages("notEligible.titleLineTwo"),
```
PatrykRudnicki: I'm thinking about new message for both of it, not add with space between

PauloPortugal: I have addressed your comment.

PR #193, opened 2019-04-12
CEDS-951 Add changeReason to Cancellation Model

PauloPortugal:
* This also included adding a Cancellation model spec class to test the
mapping was correct.
* I have also added a View spec class.

### DIFF
--- a/app/forms/cancellation/CancellationChangeReason.scala
+++ b/app/forms/cancellation/CancellationChangeReason.scala
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package forms.cancellation
+
+object CancellationChangeReason extends Enumeration {
+  type CancellationChangeReason = Value
+
+  val NoLongerRequired = Value("1")
+  val Duplication = Value("2")
+  val OtherReason = Value("3")
+}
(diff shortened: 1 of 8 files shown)


app/forms/cancellation/CancellationChangeReason.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package forms.cancellation
+
+object CancellationChangeReason extends Enumeration {
```
PatrykRudnicki: Why you use Enumeration here? What do you think about sealed trait?

test/forms/CancelDeclarationSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  val emptyRepresentativeDetails =
+    CancelDeclaration(
+      functionalReferenceId = "",
+      declarationId = "",
+      statementDescription = "",
+      changeReason = NoLongerRequired.toString
+    )
+
+  val correctCancelDeclarationJSON: JsValue = JsObject(
```
PatrykRudnicki: I'm just thinking that to prepare a JsValue is not enought to use `Json.toJson(correctCancelDeclaration)` but I don't know where you need this etc, so it's just a comment

app/forms/cancellation/CancellationChangeReason.scala
PauloPortugal: I wanted to use the Enum because I could then serialise and deserialise from the request much easiser. In the end I did not end up doing that.
https://stackoverflow.com/questions/1898932/case-objects-vs-enumerations-in-scala
These is this interesting page about it. 
If you think we should use the sealed trait then I am happy to change, if not I believe I am not introducing any disruption to the code base ;)

PatrykRudnicki: As you see I approved. I think sealed trait is better approach but more difficult or time consuming.

Enumeration is okay, just personally don't like it, I don't want to block it because of it :) So it's up to you

test/forms/CancelDeclarationSpec.scala
PauloPortugal: That is a very good point. This was a pattern we have in our code but it would make more sense to use that.

PauloPortugal: I just did that and it worked!!!! :) Well that was very useful

app/forms/cancellation/CancellationChangeReason.scala
PauloPortugal: Ok! Thank you! :)

PR #211, opened 2019-04-30
CEDS-1146 Add featureFlag to switch between WCO-DEC Mapping strategies

PauloPortugal:
* **Add featureFlag to switch between WCO-DEC Mapping strategies**

* **Refactor `submitExportDeclaration` signature**
This is to ensure ease the creation of the WCO-DEC payload a avoid
leaking the logic on how to get the XML String payload into the
`CustomsDeclareExportsConnector.submitExportDeclaration` function.

* **Update wco-dec version to allow serialisation of WCO-DEC Declaration**

* **Map WCO-DEC Metadata based on a Mapping Strategy**
. There is never an elegant way to maintain backwards compatibility in
code, so this is the best attempt at it:
 Create a Java and Scala WCO-DEC Mapping classes
. Classes renamed to represent which model (Scala or Java) they are
mapping too. This is better than using "Old" and "New" naming
conventions.
. Having two mappers, for a Scala(old) and Java(new) models, allows us
to make use of the Strategy design pattern when time comes to marshall
the WCO-DEC declaration.
. Again the reason why going with a WCO-DEC Java model is due to ease
of serialization/deserialization.

### DIFF
--- a/app/services/WcoMetadataJavaMappingStrategy.scala
+++ b/app/services/WcoMetadataJavaMappingStrategy.scala
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services
+
+import javax.xml.bind.JAXBElement
+import services.mapping.MetaDataBuilder
+import uk.gov.hmrc.http.cache.client.CacheMap
+import wco.datamodel.wco.dec_dms._2.Declaration
+import wco.datamodel.wco.documentmetadata_dms._2.MetaData
+
+trait WcoMetadataJavaMappingStrategy extends WcoMetadataMappingStrategy {
+
+  override def produceMetaData(cacheMap: CacheMap): MetaData =
+    MetaDataBuilder.build(cacheMap)
+
+  override def declarationUcr(metaData: Any): Option[String] =
+    Option(
+      metaData
+        .asInstanceOf[MetaData]
+        .getAny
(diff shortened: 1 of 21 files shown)


app/services/WcoMetadataJavaMappingStrategy.scala
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package services
+
+import javax.xml.bind.JAXBElement
+import services.mapping.MetaDataBuilder
+import uk.gov.hmrc.http.cache.client.CacheMap
+import wco.datamodel.wco.dec_dms._2.Declaration
+import wco.datamodel.wco.documentmetadata_dms._2.MetaData
+trait WcoMetadataJavaMappingStrategy extends WcoMetadataMappingStrategy {
```
PatrykRudnicki: There should be empty line between imports and trait/classes etc.

It'll be more readable

app/services/WcoMetadataMapper.scala
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services
+
+import uk.gov.hmrc.http.cache.client.CacheMap
+class WcoMetadataMapper {
```
PatrykRudnicki: The same here

app/services/WcoMetadataMappingStrategy.scala
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services
+
+import uk.gov.hmrc.http.cache.client.CacheMap
+trait WcoMetadataMappingStrategy {
```
PatrykRudnicki: And here

app/services/WcoMetadataJavaMappingStrategy.scala
PauloPortugal: Comment Addressed!

app/services/WcoMetadataMapper.scala
PauloPortugal: Comment Addressed!

app/services/WcoMetadataMappingStrategy.scala
PauloPortugal: Comment Addressed!

PR #219, opened 2019-05-01
[CEDS-1130] - ip address whitelisting

raghu1978:


app/config/AppConfig.scala
```diff
@@ -120,4 +123,28 @@ class AppConfig @Inject()(override val runModeConfiguration: Configuration, val
   private def feature2Key(feature: Feature): String = s"microservice.services.features.$feature"
 
   private def str2FeatureStatus(str: String): FeatureStatus = FeatureStatus.withName(str)
+
+  private def whitelistConfig(key: String): Seq[String] =
+    Some(
```
PatrykRudnicki: Just wondering why we have `Some` here?

app/config/filters/ExcludingCSRFFilter.scala
```diff
@@ -0,0 +1,38 @@
...
+class ExcludingCSRFFilter @Inject()(filter: CSRFFilter) extends EssentialFilter {
+
+  override def apply(nextFilter: EssentialAction): EssentialAction = new EssentialAction {
+
+    override def apply(rh: RequestHeader): Accumulator[ByteString, Result] = {
+      val chainedFilter = filter.apply(nextFilter)
+      if (rh.tags.getOrElse("ROUTE_COMMENTS", "").contains("NOCSRF")) {
+        nextFilter(rh)
+      } else {
+        chainedFilter(rh)
```
PatrykRudnicki: Here we can just use: `filter.apply(nextFilter)`

We don't need to evaluate it before if, just only when we need it

conf/application.conf
```diff
@@ -32,7 +32,9 @@ play.filters.csrf.contentType.whiteList = ["application/xml", "application/json"
 
 play.http.requestHandler = "uk.gov.hmrc.play.bootstrap.http.RequestHandler"
 play.http.errorHandler = "handlers.ErrorHandler"
-play.http.filters = "uk.gov.hmrc.play.bootstrap.filters.FrontendFilters"
+#play.http.filters = "uk.gov.hmrc.play.bootstrap.filters.FrontendFilters"
```
PatrykRudnicki: Can you remove this comment?

And base on this change I think that we can get rid of `FrontendFilters` too

app/config/AppConfig.scala
raghu1978: either Some or using Seq with similar situation with splitting the string

PR #222, opened 2019-05-03
CEDS-1146 Log WCO-DEC mapper being used

PauloPortugal:


PauloPortugal:
> My propose is to use specific logger for classes. It'll be easier to maintain and read it from the logs.
> 
> I implemented this approach in backend application, but not sure that it's the final approach.
> 
> You can achieve it with `private val logger = Logger(this.getClass())` and after that use `logger` not `Logger`. This will give a specific name for this logger in logs, so you can easy find that logs are from app config.
> 
> But LGTM! Approved, comment for the future.

I have addressed your comment.
I fully agree with you!
Thank you

PR #223, opened 2019-05-03
CEDS-1141 Missing logs error

PatrykRudnicki:


PatrykRudnicki:
@PauloPortugal It's old implementation, will check it carefully during testing

PR #229, opened 2019-05-07
[CEDS-1142] - Audit Service for frontend with single audit event

raghu1978:


app/services/SubmissionService.scala
```diff
@@ -0,0 +1,90 @@
...
+  appConfig: AppConfig,
+  cacheService: CustomsCacheService,
+  exportsConnector: CustomsDeclareExportsConnector,
+  auditService: AuditService,
+  exportsMetrics: ExportsMetrics
+) {
+
+  def submit(
+    cacheMap: CacheMap
+  )(implicit request: JourneyRequest[_], hc: HeaderCarrier, ec: ExecutionContext): Future[Option[String]] = {
```
PatrykRudnicki: Service shouldn't have any request, headerCarrier, executionContext. All of this should be handled in the controller.

app/services/SubmissionService.scala
```diff
@@ -0,0 +1,90 @@
...
+    exportsConnector.submitExportDeclaration(data.lrn, data.ducr, data.payload).flatMap {
+      case HttpResponse(ACCEPTED, _, _, _) =>
+        cacheService.remove(cacheId).map { _ =>
+          auditService.audit(AuditTypes.Submission, auditData(data.lrn, data.ducr, Success.toString))
+          exportsMetrics.incrementCounter(submissionMetric)
+          timerContext.stop()
+          data.lrn
+        }
+      case error =>
+        Logger.error(s"Error response from backend ${error.body}")
```
PatrykRudnicki: Please use specific logger for this service. Don't use general one from play. You can check other services how we implemented this.

app/services/audit/AuditService.scala
```diff
@@ -0,0 +1,70 @@
...
+import uk.gov.hmrc.play.audit.AuditExtensions
+import uk.gov.hmrc.play.audit.http.connector.AuditResult.{Disabled, Failure, Success}
+import uk.gov.hmrc.play.audit.http.connector.{AuditConnector, AuditResult}
+import uk.gov.hmrc.play.audit.model.DataEvent
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AuditService @Inject()(connector: AuditConnector, appConfig: AppConfig)(implicit ec: ExecutionContext) {
+
+  private def createAuditEvent(audit: Audit, auditData: Map[String, String])(implicit hc: HeaderCarrier) =
```
PatrykRudnicki: The same here about headerCarrier

test/controllers/declaration/SummaryPageControllerSpec.scala
```diff
@@ -37,7 +37,7 @@ import scala.concurrent.Future
 class SummaryPageControllerSpec extends CustomExportsBaseSpec {
 
   private trait Test {
-    implicit val headerCarrierMock = mock[HeaderCarrier]
+    //implicit val headerCarrierMock = mock[HeaderCarrier]
```
PatrykRudnicki: Please remove comment

test/base/CustomExportsBaseSpec.scala
```diff
@@ -63,6 +66,12 @@ trait CustomExportsBaseSpec
   val mockNrsService: NRSService = mock[NRSService]
   val mockItemsCachingService: ItemsCachingService = mock[ItemsCachingService]
 
+  implicit val hc: HeaderCarrier =
```
PatrykRudnicki: I'm not sure that it's the correct place for HeaderCarrier as implicit. Too big scope. I prefer like before, in specific test if you need it.

app/services/SubmissionService.scala
PauloPortugal: I agree with this approach. We should not leak http request middleware to the business logic service.
Advantages: 
 - Easier to maintain/read
 - Easier to unit test

app/services/SubmissionService.scala
PauloPortugal: Ahh the Logger Police is back in town :)
I agree with it though. I was advised the same thing last week :+1:

test/base/CustomExportsBaseSpec.scala
PauloPortugal: I actually side with @raghu1978 on this one. 
Happy either way.

PatrykRudnicki: Just my thoughts, sometimes you don't want to have authorisation inside. We can stay with this

PR #232, opened 2019-05-09
CEDS-1146 Fix Empty values on new mapping strategy

PauloPortugal:
If empty fields/strings reach the mapping strategy we should not
instantiate new instances as it will cause the WCO-DEC schema to fail

### DIFF
--- a/app/forms/common/Address.scala
+++ b/app/forms/common/Address.scala
@@ -28,7 +28,10 @@ case class Address(
   townOrCity: String, // alphanumeric length 1 - 35
   postCode: String, // alphanumeric length 1 - 9
   country: String // full country name, convert to 2 upper case alphabetic characters for backend
-)
+) {
+  def isDefined(): Boolean =
+    fullName.nonEmpty || addressLine.nonEmpty || townOrCity.nonEmpty || postCode.nonEmpty || country.nonEmpty
+}
 
 object Address {
   implicit val format = Json.format[Address]
--- a/app/services/mapping/goodsshipment/ConsignmentBuilder.scala
+++ b/app/services/mapping/goodsshipment/ConsignmentBuilder.scala
@@ -28,89 +28,117 @@ import wco.datamodel.wco.declaration_ds.dms._2._
 
 object ConsignmentBuilder {
 
-  def build(implicit cacheMap: CacheMap): GoodsShipment.Consignment =
-    cacheMap
-      .getEntry[CarrierDetails](CarrierDetails.id)
-      .map(carrierDetails => createConsignmentAddress(carrierDetails.details))
-      .orNull
-
-  private def createConsignmentAddress(
-    details: EntityDetails
-  )(implicit cacheMap: CacheMap): GoodsShipment.Consignment = {
-    val id = new GoodsLocationIdentificationIDType()
-    id.setValue(details.eori.orNull)
-
-    val name = new GoodsLocationNameTextType()
-
-    val goodsAddress = new Consignment.GoodsLocation.Address()
-    details.address.map(address => {
-      name.setValue(address.fullName)
-
-      val line = new AddressLineTextType()
-      line.setValue(address.addressLine)
-
-      val city = new AddressCityNameTextType
-      city.setValue(address.townOrCity)
-
-      val postcode = new AddressPostcodeIDType()
-      postcode.setValue(address.postCode)
-
-      val countryCode = new AddressCountryCodeType
-      countryCode.setValue(
-        allCountries.find(country => address.country.contains(country.countryName)).map(_.countryCode).getOrElse("")
-      )
-
-      goodsAddress.setLine(line)
-      goodsAddress.setCityName(city)
-      goodsAddress.setCountryCode(countryCode)
-      goodsAddress.setPostcodeID(postcode)
-    })
-
-    val goodsLocation = new GoodsShipment.Consignment.GoodsLocation()
-    goodsLocation.setID(id)
-    goodsLocation.setName(name)
-    goodsLocation.setAddress(goodsAddress)
-
+  def build(implicit cacheMap: CacheMap): GoodsShipment.Consignment = {
     val consignment = new GoodsShipment.Consignment()
-    consignment.setGoodsLocation(goodsLocation)
+
+    consignment.setGoodsLocation(buildGoodsLocation(cacheMap))
     consignment.setContainerCode(buildContainerCode)
     consignment.setArrivalTransportMeans(buildArrivalTransportMeans)
     consignment.setDepartureTransportMeans(buildDepartureTransportMeans())
     consignment
   }
 
+  private def buildGoodsLocation(cacheMap: CacheMap): Consignment.GoodsLocation =
+    cacheMap
+      .getEntry[CarrierDetails](CarrierDetails.id)
+      .filter(
+        carrierDetails =>
+          carrierDetails.details.eori.getOrElse("").nonEmpty ||
+            (carrierDetails.details.address.isDefined && carrierDetails.details.address.get.isDefined())
+      )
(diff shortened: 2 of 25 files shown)


app/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
```diff
@@ -33,28 +33,47 @@ object PreviousDocumentsBuilder {
...
+      .filter(isDefined)
       .map(_.documents.map(createPreviousDocuments))
       .getOrElse(Seq.empty)
       .toList
       .asJava
 
+  private def isDefined(previousDocumentsData: PreviousDocumentsData): Boolean =
+    previousDocumentsData.documents.nonEmpty && previousDocumentsData.documents.forall(
+      doc =>
+        !doc.goodsItemIdentifier.getOrElse("").isEmpty ||
```
PatrykRudnicki: Instead of using `!` and `isEmpty` probably it will be easier to use nonEmpty

Comment for lines 66 and 71 too.

app/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
```diff
@@ -33,28 +33,47 @@ object PreviousDocumentsBuilder {
...
+    )
+
   private def createPreviousDocuments(document: Document): GoodsShipment.PreviousDocument = {
-    val categoryCode = new PreviousDocumentCategoryCodeType()
-    categoryCode.setValue(document.documentCategory)
+    val previousDocument = new GoodsShipment.PreviousDocument()
 
-    val id = new PreviousDocumentIdentificationIDType()
-    id.setValue(document.documentReference)
+    if (!Option(document.documentCategory).getOrElse("").isEmpty) {
```
PatrykRudnicki: The same here about `!` and `isEmpty` and I don't understand why you wrapping something in Option and after that using GetOrElse

If `documentCategory` is String you can directly use isEmpty/nonEmpty on the string

Comment for lines 60 and 71 too.

app/services/mapping/goodsshipment/UCRBuilder.scala
```diff
@@ -26,22 +26,19 @@ object UCRBuilder {
   def build(implicit cacheMap: CacheMap): UCR =
     cacheMap
       .getEntry[ConsignmentReferences](ConsignmentReferences.id)
+      .filter(isDefined)
       .map(createUCR)
       .orNull
 
-  private def createUCR(data: ConsignmentReferences): UCR = {
+  private def isDefined(reference: ConsignmentReferences): Boolean = !reference.ducr.getOrElse(Ducr("")).ducr.isEmpty
```
PatrykRudnicki: The same about `!` and `isEmpty`

I'm thinking about preparing isEmpty method inside Ducr class. Using will be easier and we can do it with every class that you're using for dec

app/services/mapping/goodsshipment/WarehouseBuilder.scala
```diff
@@ -25,20 +25,28 @@ object WarehouseBuilder {
   def build(implicit cacheMap: CacheMap): Warehouse =
     cacheMap
       .getEntry[WarehouseIdentification](WarehouseIdentification.formId)
+      .filter(isDefined)
       .map(createWarehouse)
       .orNull
 
+  private def isDefined(warehouse: WarehouseIdentification): Boolean =
+    !warehouse.identificationNumber.getOrElse("").isEmpty
```
PatrykRudnicki: In whole file `!` and `isEmpty`

app/services/mapping/goodsshipment/WarehouseBuilder.scala
```diff
@@ -25,20 +25,28 @@ object WarehouseBuilder {
...
+
   private def createWarehouse(data: WarehouseIdentification): Warehouse = {
+    val warehouse = new Warehouse()
 
-    val id = new WarehouseIdentificationIDType()
-    id.setValue(data.identificationNumber.map(_.drop(1).toString).getOrElse(""))
+    if (!data.identificationNumber.map(_.drop(1).toString).getOrElse("").isEmpty) {
+      val id = new WarehouseIdentificationIDType()
+      id.setValue(data.identificationNumber.map(_.drop(1).toString).getOrElse(""))
```
PatrykRudnicki: `data.identificationNumber.map(_.drop(1).toString).getOrElse("")`

We can save it in variable and declare before if. Reusing in condition and here.

The same below

app/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
PauloPortugal: :wink: Unfortunately, the `String` class does not have a `isNotEmpty` method and we are not using `Ruby` where we could override the `String` class.

app/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
PauloPortugal: Good point!
Done! :)

app/services/mapping/goodsshipment/PreviousDocumentsBuilder.scala
PatrykRudnicki: `nonEmpty`

PauloPortugal: and as you so eloquently, demonstrated there is a `nonEmpty` method :smile: 
Awesome! 
Thank you so much! :)

app/services/mapping/goodsshipment/WarehouseBuilder.scala
PauloPortugal: I have removed that condition. I am using `filter`

PauloPortugal:
@raghu1978 has mentioned that we could use a generic function to test for empty Strings/Values:
```
```

app/services/mapping/goodsshipment/ConsignmentBuilder.scala
```diff
@@ -122,4 +150,15 @@ object ConsignmentBuilder {
...
     codeType
   }
+
+  private def isDefined(carrierDetails: CarrierDetails, cacheMap: CacheMap): Boolean =
+    carrierDetails.details.eori.getOrElse("").nonEmpty ||
+      (carrierDetails.details.address.isDefined && carrierDetails.details.address.get.isDefined()) ||
+      (cacheMap
+        .getEntry[TransportInformation](TransportInformation.id)
+        .isDefined && isTransportInformationDefined(
+        cacheMap
```
PatrykRudnicki: You can only once retrieve this data from cache, save it in variable and reuse in both places. You don't need to getEntry two times

raghu1978:
we might need to move the conditional logic to a generic function or to filter out in some kind of reads within form objects

app/services/mapping/goodsshipment/ConsignmentBuilder.scala
PauloPortugal: I do not see this as a problem. because `cacheMap` is just a map in memory.
We are losing some efficiency in transversing through a `Map` structure instead of just referring to the reference, but I prefer to have that performance hit and have a more readable and stable code pattern.
If you think this is something that we should be doing, then I will address that on my next commit/PR.

Please let me know ;)

app/forms/common/Address.scala
```diff
@@ -28,7 +28,10 @@ case class Address(
   townOrCity: String, // alphanumeric length 1 - 35
   postCode: String, // alphanumeric length 1 - 9
   country: String // full country name, convert to 2 upper case alphabetic characters for backend
-)
+) {
```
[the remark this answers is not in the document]
PauloPortugal: This is something we will address and experiment in the next commit/PR ;)

PR #233, opened 2019-05-10
CEDS-1121: added new autocomplete component and changed tests

mattclark-zerogravit:


test/base/ViewValidator.scala
```diff
@@ -59,6 +59,18 @@ trait ViewValidator extends MustMatchers {
     element
   }
 
+  def getSelectedValue(html: Html, name: String): String = {
+
+    var selectedVal = ""
```
PatrykRudnicki: ahh, I hate vars

I don't want to block it, because it's for tests, but my propose:

```
asDocument(html).getElementById(name).children().map { option =>
      if(option.hasAttr("selected")) option.`val`()
      else ""
    }.filter(_.nonEmpty).headOption.getOrElse("")
```

of course you can extract this `asDocument(html).getElementById(name).children()` to options and use map on options, I mean to avoid using var or some other approach :)

Just proposal

mattclark-zerogravit: Yep will change to this. Still looks odd but is more functional. Problem of using Jsoup with Scala

PR #235, opened 2019-05-13
CEDS-1143 Check the form validation and align the mapping

PauloPortugal:
After a conversation with @raghu1978 we have realised that a lot of the
mapping logic checks are overly defensive as the production code will
never allow data to be in a given state.

For example on a form where a EORI and an address is being requested,
only one or them will be requested and none of them can be
`None/Null/Empty`.

### DIFF
--- a/app/services/mapping/declaration/AgentBuilder.scala
+++ b/app/services/mapping/declaration/AgentBuilder.scala
@@ -28,47 +28,51 @@ object AgentBuilder {
   def build(implicit cacheMap: CacheMap): Declaration.Agent =
     cacheMap
       .getEntry[RepresentativeDetails](RepresentativeDetails.formId)
-      .map(data => createAgent(data.details.orNull))
+      .filter(isDefined)
+      .map(data => createAgent(data.details.get))
       .orNull
+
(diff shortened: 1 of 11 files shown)


app/services/mapping/declaration/AgentBuilder.scala
```diff
@@ -28,47 +28,51 @@ object AgentBuilder {
...
+      val agentId = new AgentIdentificationIDType()
+      agentId.setValue(details.eori.get)
+      agent.setID(agentId)
+    } else {
 
-    val agentAddress = new Agent.Address()
+      val agentAddress = new Agent.Address()
 
-    details.address.map(address => {
+      details.address.map(address => {
```
raghu1978: Address is used in different mappings for all the pages , it would be good if we have a common method of extracting and mapping header if the the types are same.

PauloPortugal: The return type is always of a different type.

PauloPortugal: We could convert to a Map or Json and then create an instance of another type.
Let me that onboard and unblock our pipeline

PR #238, opened 2019-05-14
WIP: CEDS-1046 Map standard journey fields

PauloPortugal:
The fulldec is not supported in the new or old mapping strategy,
simply because we did not do the mapping for destination `ExportCountry`
 and `DestinationCountry` (both of them have a supplementary/standard form)

We need to go to the form and check which fields we need to map to.


PauloPortugal:
I am not sure these will fix the standard journey, as we forgot to map one field here.
Apologies for making that clear.

PR #246, opened 2019-05-21
CEDS-1213 Record the journey choice in relevant DB record

PauloPortugal:
We have two records in the DB for each EORI:
 1. EORI only: which the JourneyAction looks at to understand which
    Jounrey the user is using
 2. JOURNEYType_EORI: which is based on the Jouney choice,

This change is to also record the Choice in option 2, to aid the
Mapping Strategy for both Supplementary and Standard journeys


app/controllers/ChoiceController.scala
```diff
@@ -54,16 +54,19 @@ class ChoiceController @Inject()(
...
-              case SupplementaryDec | StandardDec =>
-                Redirect(controllers.declaration.routes.DispatchLocationPageController.displayPage())
-              case CancelDec =>
-                Redirect(controllers.routes.CancelDeclarationController.displayForm())
-              case Submissions =>
-                Redirect(controllers.routes.NotificationsController.listOfSubmissions())
-              case _ =>
-                Redirect(controllers.routes.ChoiceController.displayChoiceForm())
+          customsCacheService.cache[Choice](eoriCacheId, choiceId, validChoice).flatMap { _ =>
+            customsCacheService.cache[Choice](s"${validChoice.value}-${request.user.eori}", choiceId, validChoice).map {
```
raghu1978: just want to check why is this required? why can't you get choice from journeyAction, for each request the choice exists for the method to pass to services to use jourenytype I believe

PatrykRudnicki: This is ChoiceController, so you don't know what is the user choice.

But I'm not sure that I understand idea to save choice two times in different caches.

Is it a part of migration? and when you change retrieving Choice from different place you'll remove one of the method. In other case I don't see any specific reason to do it in this way.

MaciejRewera: Maybe I am missing something, but what is the point in storing choice value in 2 places? What is more, will we not end up with 3 documents in mongo when the user is making any declaration?

PauloPortugal: I have removed this :) Please review again.

app/services/ItemsCachingService.scala
```diff
@@ -44,6 +34,52 @@ import scala.concurrent.{ExecutionContext, Future}
 @Singleton
 class ItemsCachingService @Inject()(cacheService: CustomsCacheService)(appConfig: AppConfig) {
 
+  /*
+    Fetch cache elements of all the pages using goodsItemCacheId for pages procedure codes, itemType, packaging, goods measure,
+    additional information , documents produced.
+    create related wco-dec elements
+    create goods item for the cache and append to items already added in the cache
+    save updated items to cache and remove the elements in goodsItemCacheId
+   */
```
PatrykRudnicki: It's better to use the JavaDoc to describe the method

app/services/mapping/declaration/ExitOfficeBuilder.scala
```diff
@@ -26,14 +26,11 @@ import wco.datamodel.wco.declaration_ds.dms._2._
...
-  def build(implicit cacheMap: CacheMap): Declaration.ExitOffice =
-    cacheMap
-      .getEntry[Choice](Choice.choiceId)
-      .map {
-        case Choice(AllowedChoiceValues.SupplementaryDec) => buildExitOfficeFromSupplementary(cacheMap)
-        case Choice(AllowedChoiceValues.StandardDec)      => buildExitOfficeFromStandard(cacheMap)
-      }
-      .orNull
+  def build(implicit cacheMap: CacheMap, choice: Choice): Declaration.ExitOffice =
+    Option(choice).map {
```
PatrykRudnicki: Hmm, why this? You can just use:

```
choice match {
  case ...
  case ...
}
```

app/services/mapping/declaration/PresentationOfficeBuilder.scala
```diff
@@ -26,9 +26,8 @@ import wco.datamodel.wco.declaration_ds.dms._2._
 
 object PresentationOfficeBuilder {
 
-  def build(implicit cacheMap: CacheMap): Declaration.PresentationOffice =
-    cacheMap
-      .getEntry[Choice](Choice.choiceId)
+  def build(implicit cacheMap: CacheMap, choice: Choice): Declaration.PresentationOffice =
+    Option(choice)
```
PatrykRudnicki: Similar situation here, why you're wrapping Choice into Option?

app/services/mapping/declaration/SpecificCircumstancesCodeBuilder.scala
```diff
@@ -25,9 +25,8 @@ import wco.datamodel.wco.declaration_ds.dms._2._
 
 object SpecificCircumstancesCodeBuilder {
 
-  def build(implicit cacheMap: CacheMap): DeclarationSpecificCircumstancesCodeCodeType =
-    cacheMap
-      .getEntry[Choice](Choice.choiceId)
+  def build(implicit cacheMap: CacheMap, choice: Choice): DeclarationSpecificCircumstancesCodeCodeType =
+   Option(choice)
```
PatrykRudnicki: The same here

app/services/mapping/goodsshipment/DestinationBuilder.scala
```diff
@@ -26,14 +30,11 @@ import wco.datamodel.wco.declaration_ds.dms._2.DestinationCountryCodeType
...
-  def build(implicit cacheMap: CacheMap): GoodsShipment.Destination =
-    cacheMap
-      .getEntry[Choice](Choice.choiceId)
-      .map {
-        case Choice(AllowedChoiceValues.SupplementaryDec) => buildSupplementary(cacheMap)
-        case Choice(AllowedChoiceValues.StandardDec)      => buildStandard(cacheMap)
-      }
-      .orNull
+  def build(implicit cacheMap: CacheMap, choice: Choice): GoodsShipment.Destination =
+    Option(choice).map {
```
PatrykRudnicki: Here too

test/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemData.scala
```diff
@@ -95,4 +98,17 @@ trait GovernmentAgencyGoodsItemData {
   val additionalInformation = AdditionalInformation(statementCode, descriptionValue)
   val additionalInformationData = AdditionalInformationData(Seq(additionalInformation))
 
+  val amount = Amount(Some("GBP"), Some(BigDecimal(123)))
+  val goodsMeasure = GoodsMeasure(Some(Measure(Some("kg"), Some(BigDecimal(10)))))
+  val commodity = Commodity(Some("someDescription"), Seq(), Seq(), Some(goodsMeasure))
+//
+//  val govermentAgencyGoodsItem = GovernmentAgencyGoodsItem(1, Some(amount)
```
PatrykRudnicki: Please remove comments :)

app/services/mapping/declaration/ExitOfficeBuilder.scala
PauloPortugal: Thanks! Please review again! :)

app/services/mapping/declaration/PresentationOfficeBuilder.scala
PauloPortugal: Thanks! Please review again! :)

app/services/mapping/declaration/SpecificCircumstancesCodeBuilder.scala
PauloPortugal: Thanks! Please review again! :)

app/services/mapping/declaration/ExitOfficeBuilder.scala
PauloPortugal: I have addressed this! :)

app/services/mapping/declaration/PresentationOfficeBuilder.scala
PauloPortugal: I have addressed this! :)

app/services/mapping/declaration/SpecificCircumstancesCodeBuilder.scala
PauloPortugal: I have addressed this! :)

app/services/mapping/goodsshipment/DestinationBuilder.scala
PauloPortugal: I have addressed this! :)

test/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemData.scala
PauloPortugal: Done! :)

PR #249, opened 2019-05-23
CEDS-1093 Retain other fields values when adding new value

MaciejRewera:
When a page contains both single and multi-value fields, there was a
problem that after adding new value the single ones would disappear
after page refresh.

### DIFF
--- a/app/controllers/declaration/DestinationCountriesController.scala
+++ b/app/controllers/declaration/DestinationCountriesController.scala
@@ -28,10 +28,9 @@ import javax.inject.Inject
 import models.requests.JourneyRequest
 import play.api.data.{Form, FormError}
 import play.api.i18n.{I18nSupport, MessagesApi}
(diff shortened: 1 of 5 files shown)


app/controllers/declaration/DestinationCountriesController.scala
```diff
@@ -54,52 +53,49 @@ class DestinationCountriesController @Inject()(
...
     }
 
   def saveCountries(): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
-    (request.choice.value match {
-      case SupplementaryDec => handleSavingSupplementary()
-      case StandardDec      => handleSavingStandard()
-    }).flatMap {
-      case Left(resultView) => Future.successful(BadRequest(resultView))
-      case Right(call)      => Future.successful(Redirect(call))
+    request.choice.value match {
```
PauloPortugal: :+1:

PR #250, opened 2019-05-24
CEDS-1216 Add missing removal tests for Item Type page

MaciejRewera:


app/controllers/declaration/DestinationCountriesController.scala
```diff
@@ -178,20 +180,20 @@ class DestinationCountriesController @Inject()(
   private def removeRoutingCountry(keys: Seq[String], cachedData: DestinationCountriesStandard)(
     implicit request: JourneyRequest[_]
   ): Future[Result] = {
-    val key = keys.headOption.getOrElse("")
+    val key = keys.headOption.getOrElse("").toInt
```
PatrykRudnicki: `getOrElse` will return you String. Empty string cannot be cast to Int
```
scala> "".toInt
java.lang.NumberFormatException: For input string: ""
  at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
  at java.lang.Integer.parseInt(Integer.java:592)
  at java.lang.Integer.parseInt(Integer.java:615)
  at scala.collection.immutable.StringLike.toInt(StringLike.scala:304)
  at scala.collection.immutable.StringLike.toInt$(StringLike.scala:304)
  at scala.collection.immutable.StringOps.toInt(StringOps.scala:33)
  ... 28 elided
```

app/controllers/util/FormAction.scala
```diff
@@ -34,9 +36,19 @@ object FormAction {
     }.headOption.getOrElse(Unknown)
 }
 
-case object Unknown extends FormAction
-case object Add extends FormAction
-case object SaveAndContinue extends FormAction
-case object Continue extends FormAction
+case object Unknown extends FormAction {
+  override val label: String = "Unknown"
```
PatrykRudnicki: Why you do this?

Result from val label is the same like `toString` method.

toString method from case objects return their name. So I'm not sure about approach and intension here.

test/controllers/declaration/ItemTypePageControllerSpec.scala
```diff
@@ -569,15 +775,25 @@ object ItemTypePageControllerSpec {
...
+        case (code, idx) => nationalAdditionalCodesKey + "[" + idx + "]" -> code
+      }
+    )
+
+  def addActionTypeToFormData(formAction: FormAction, formData: Map[String, String]): Map[String, String] =
+    Map(formAction.label -> getKey(formAction)) ++ formData
+
+  private def getKey(action: FormAction): String = action match {
+    case Remove(keys) => keys.head
+    case _            => ""
```
PatrykRudnicki: For `_` you want to have empty string? are you sure about it? I'm thinking that for `Add` you should have `Add`, but I can understand this  method wrong.

If it's only affects remove key, you can change it to `getRemovalKey`.

I'm worried about this file, because I know that we decided to keep controller tests short and small, but this one is huge. It was big before but now... Added 940 lines. I'm thinking to add some annotation, comment, TODO to this test to split it and extract a huge part of it.

app/controllers/declaration/DestinationCountriesController.scala
MaciejRewera: Thank you for your piece od advice. I've changed it to throw an exception explicitly. It is more clear now, that this is expected behaviour.

app/controllers/util/FormAction.scala
MaciejRewera: It is not the same for the Remove case class, because `toString()` adds all the fields from case class. Here, we want to have the class name only.

test/controllers/declaration/ItemTypePageControllerSpec.scala
MaciejRewera: In this case, the key is the index of the element to remove. It is misleading so I changed slightly implementation to be more explanatory.

PR #251, opened 2019-05-28
CEDS-1156 Upgrade for libraries and migration to play 2.6

PatrykRudnicki:


project/AppDependencies.scala
```diff
@@ -6,24 +6,24 @@ object AppDependencies {
...
-    "uk.gov.hmrc" %% "play-reactivemongo" % "6.7.0",
+    "uk.gov.hmrc" %% "play-conditional-form-mapping" % "1.0.0-play-26",
+    "uk.gov.hmrc" %% "simple-reactivemongo" % "7.19.0-play-26",
     "uk.gov.hmrc" %% "logback-json-logger" % "4.6.0",
-    "uk.gov.hmrc" %% "govuk-template" % "5.35.0-play-25",
-    "uk.gov.hmrc" %% "play-health" % "3.14.0-play-25",
-    "uk.gov.hmrc" %% "play-ui" % "7.39.0-play-25",
-    "uk.gov.hmrc" %% "http-caching-client" % "8.3.0",
-    "uk.gov.hmrc" %% "bootstrap-play-25" % "4.11.0",
-    "uk.gov.hmrc" %% "play-language" % "3.4.0",
```
PatrykRudnicki: Just comment here, this library is our blocker. We can't use it with actual version of play.

I decided to remove it, we're not using this and moreover language implementation is easy. I'll implement it in our service or prepare PR with upgrade for this library. But now we don't need it

test/config/filters/WhitelistFilterSpec.scala
```diff
@@ -20,15 +20,20 @@ import config.AppConfig
...
 import play.api.mvc.Results.Ok
 import play.api.mvc.{Action, Call}
 import play.api.test.FakeRequest
 import play.api.test.Helpers._
+import uk.gov.hmrc.play.bootstrap.config.{RunMode, ServicesConfig}
 
 class WhitelistFilterSpec extends PlaySpec with GuiceOneServerPerSuite {
 
-  lazy val mockConfig = new MockAppConfig(app.configuration)
+  //val servicesConfig = app.injector.instanceOf[ServicesConfig]
```
MaciejRewera: If there is no extra reason to keep this line commented, then I would remove it.

test/controllers/actions/FakeAuthAction.scala
```diff
@@ -57,6 +63,9 @@ object FakeAuthAction extends AuthAction {
 }
 
 class TestAuthAction extends AuthAction {
+  override val executionContext: ExecutionContext = ???
+  override val parser: BodyParser[AnyContent] = ???
```
MaciejRewera: Why are these not implemented?

test/views/ChoiceViewSpec.scala
```diff
@@ -87,7 +89,11 @@ class ChoiceViewSpec extends ViewSpec with ChoiceMessages with CommonMessages {
 
       """.stripMargin)
 
-      val supplementaryAppConfig = new AppConfig(Configuration(supplementaryConfig), Environment.simple())
+
+      val conf: Configuration = Configuration(supplementaryConfig)
+      val runMode: RunMode = new RunMode(conf, Test)
+      val servicesConfig = new ServicesConfig(conf, runMode)
+      val supplementaryAppConfig = new AppConfig(conf, Environment.simple(), servicesConfig, "AppName")
```
MaciejRewera: Just wondering... Can we use something like FakeAppConfig? And have a method that would build and return this config?

test/config/filters/WhitelistFilterSpec.scala
PatrykRudnicki: Yeah, good spot. I had a problem with this tests. I tried few different approach, just forgot about cleaning code after everything.

test/controllers/actions/FakeAuthAction.scala
PatrykRudnicki: It's just need to be overridden when you extends AuthAction. FakeAuthAction is for test purposes and both of those vals are not used. Implementation can be empty.

PR #254, opened 2019-05-30
CEDS-1213 Correctly map WCO DEC using the new mapping strategy

PauloPortugal:
* Map Declaration.Agent.FunctionCode correctly
* Correctly map WCO BorderMeansTransport 
* Fix WCO Consignment mapping 
* Remove TransportInformation form 
* Fix AEOMutualRecognitionParty mapping


app/services/mapping/declaration/AgentBuilder.scala
```diff
@@ -29,50 +29,66 @@ object AgentBuilder {
...
+            agentAddress.setPostcodeID(postcode)
+          })
+          agent.setAddress(agentAddress)
+      }
+    )
     agent
   }
+  private def setStatusCode(data: RepresentativeDetails) =
+    data.statusCode
+      .map(value => {
```
PatrykRudnicki: You don't need to put `{` after `value =>`

Just use

```
.map { value =>
   val ....
}
```

It's in a lot of places, it's not wrong, just to inform you :)

app/services/mapping/goodsshipment/consignment/ArrivalTransportMeansBuilder.scala
```diff
@@ -0,0 +1,38 @@
...
+import wco.datamodel.wco.dec_dms._2.Declaration.GoodsShipment.Consignment.ArrivalTransportMeans
+import wco.datamodel.wco.declaration_ds.dms._2.ArrivalTransportMeansModeCodeType
+
+object ArrivalTransportMeansBuilder {
+
+  def build(implicit cacheMap: CacheMap): GoodsShipment.Consignment.ArrivalTransportMeans =
+    cacheMap
+      .getEntry[WarehouseIdentification](WarehouseIdentification.formId)
+      .filter(_.inlandModeOfTransportCode.getOrElse("").nonEmpty)
+      .map(transportInformation => {
```
PatrykRudnicki: Same here about parenthesis

PauloPortugal: I have addressed those

app/services/mapping/declaration/AgentBuilder.scala
PauloPortugal: I have addressed those

PR #259, opened 2019-06-04
Audit

raghu1978:
update audit event to audit full payload of user submitted data.


app/services/SubmissionService.scala
```diff
@@ -91,4 +104,62 @@ class SubmissionService @Inject()(
 
   protected case class FormattedData(lrn: Option[String], ducr: Option[String], payload: String)
 
+  def getCachedData(cacheMap: CacheMap)(implicit request: JourneyRequest[_]) = {
```
PatrykRudnicki: Please add return type to this method

app/services/SubmissionService.scala
```diff
@@ -91,4 +104,62 @@ class SubmissionService @Inject()(
...
+      Seal.formId -> Json.toJson(cacheMap.getEntry[Seq[Seal]](Seal.formId)),
+      Document.formId -> Json.toJson(
+        cacheMap.getEntry[PreviousDocumentsData](Document.formId).map(_.documents).getOrElse(Seq.empty)
+      ),
+      TotalNumberOfItems.formId -> Json.toJson(cacheMap.getEntry[TotalNumberOfItems](TotalNumberOfItems.formId)),
+      WarehouseIdentification.formId -> Json.toJson(
+        cacheMap.getEntry[WarehouseIdentification](WarehouseIdentification.formId)
+      )
+    )
+    Json.toJson(userInput).as[JsObject]
```
PatrykRudnicki: Instead of doing this (I'm not sure about this JsObject). What do you think about 
`Json.obj(Map(...))`
instead
`val userInput = Map(...)` and `Json.toJson(userInput).as[JsObject]`

test/services/audit/AuditServiceSpec.scala
```diff
@@ -46,15 +46,15 @@ class AuditServiceSpec extends AuditTestSupport with OptionValues {
     }
 
     "audit full payload" in {
-      auditService.auditAllPagesUserInput(cacheMapAllRecords)
+      auditService.auditAllPagesUserInput(Json.toJson(cacheMapAllRecords.data).as[JsObject])
```
PatrykRudnicki: Similar here, can you use `Json.obj(cacheMapAllRecords.data)`.

The same comment in line 57

app/services/SubmissionService.scala
raghu1978: I will try

app/services/SubmissionService.scala
raghu1978: updated

test/services/audit/AuditServiceSpec.scala
raghu1978: its taking a Map and then converting into varargs tuple

app/services/SubmissionService.scala
raghu1978: Json.obj expecting obj(fields: (String, JsValueWrapper)*)

it would be changing from Map to tuple and the method does the something similar as below = JsObject(fields.map(f => (f._1, f._2.asInstanceOf[JsValueWrapperImpl].field)))

PR #266, opened 2019-06-11
CEDS-1044: Add Fiscal Information Page

NC-1234:
* Add Fiscal Information Page
* Change view layout for fiscal information page
Introduced in-line radio button for yes and no answer and
introduced break line page title to follow design


app/controllers/declaration/BorderTransportController.scala
```diff
@@ -40,24 +41,34 @@ class BorderTransportController @Inject()(
...
+        .fold(
+          (formWithErrors: Form[BorderTransport]) => badRequest(formWithErrors),
+          borderTransport =>
+            customsCacheService
+              .cache[BorderTransport](cacheId, BorderTransport.formId, borderTransport)
+              .map(_ => Redirect(TransportDetailsController.displayForm()))
+        )
+    }
+
+    def badRequest(formWithErrors: Form[BorderTransport])(implicit request: JourneyRequest[_], appConfig: AppConfig) =
```
PatrykRudnicki: I think that instead of extracting `badRequest` method (it used in only one place) you can extract method to generate view based on the form. So you will be able to reuse it in `displayForm` method too.

hs_err_pid71087.log
```diff
@@ -0,0 +1,639 @@
...
+# Problematic frame:
+# J 9872 C2 java.io.File.<init>(Ljava/lang/String;)V (47 bytes) @ 0x0000000114b5879f [0x0000000114b58760+0x3f]
+#
+# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
+#
+# If you would like to submit a bug report, please visit:
+#   http://bugreport.java.com/bugreport/crash.jsp
+#
+
+---------------  T H R E A D  ---------------
```
PatrykRudnicki: Please remove this file. JVM errors, we shouldn't have something like this in codebase

PauloPortugal: :man_facepalming:

PauloPortugal: Done! :)

app/controllers/declaration/BorderTransportController.scala
PauloPortugal: Addressed! :)

PauloPortugal:
Closed because it required more Business Analysis

PR #269, opened 2019-06-12
Test cleanup to simplify assertions and reduce parens

benjaminparker:


PauloPortugal:
How did you replace all `must be(...)` with `mustBe` ?

What are the advantages? Simply a preferred syntax?

PauloPortugal:
@leszeklorek @MaciejRewera @PatrykRudnicki @bethany-kish @NC-1234 What is your opinion on the test sugar syntax?

PatrykRudnicki:
Using `must be()` has of course unnecessary parenthesis but it's closer to normal language. `mustBe` just shorter. That's all. Both approaches make the same, so it's just preferences. For me it doesn't matter what we will be using, but I'm not a fan of unnecessary changes that can be changed in two weeks, because someone will prefer `should be`. And after another two `shouldBe`...

PatrykRudnicki:
I'll close it without removing branch, because probably we stopped discussion for now

PR #271, opened 2019-06-17
CEDS-1246 remove to metadata properties

PauloPortugal:
Also ran ScalaFmt on another commit


app/config/AppConfig.scala
```diff
@@ -104,9 +104,9 @@ class AppConfig @Inject()(
     FeatureStatus.withName(loadConfig(feature2Key(Feature.default)))
 
   def wcoMetadataMapper(): WcoMetadataMapper with WcoMetadataMappingStrategy = {
-      logger.warn("Using WcoMetadataJavaMappingStrategy as the WCO-DEC mapper")
-      new WcoMetadataMapper with WcoMetadataJavaMappingStrategy
-    }
+    logger.warn("Using WcoMetadataJavaMappingStrategy as the WCO-DEC mapper")
```
PatrykRudnicki: I think it should be debug, it's information for us, not warning. Old implementation, just commented.

app/config/AppConfig.scala
```diff
@@ -103,10 +104,7 @@ class AppConfig @Inject()(
   lazy val defaultFeatureStatus: features.FeatureStatus.Value =
     FeatureStatus.withName(loadConfig(feature2Key(Feature.default)))
 
-  def wcoMetadataMapper(): WcoMetadataMapper with WcoMetadataMappingStrategy = {
-    logger.warn("Using WcoMetadataJavaMappingStrategy as the WCO-DEC mapper")
-    new WcoMetadataMapper with WcoMetadataJavaMappingStrategy
-  }
+  def wcoMetadataMapper(): WcoMetadataMapper = new WcoMetadataMapper
```
PatrykRudnicki: Do you use this method somewhere? Is this method important? It just creates `WcoMetadataMapper`, so you can inline it.

app/config/AppConfig.scala
PauloPortugal: I have removed it ;)

PauloPortugal: I have removed it

PR #279, opened 2019-06-25
CEDS-1132 Countries dropdown throughout the journey.

benjaminparker:
Replaces all existing country AutoComplete inputs with accessible component. Also reworks how the country list is supplied to each component, pulling it out of the Controllers and into the views.


PauloPortugal:
Commits should include the `JIRA id`. `git rebase -i HEAD~8` if you want to change.

app/controllers/declaration/DestinationCountriesController.scala
```diff
@@ -69,14 +61,14 @@ class DestinationCountriesController @Inject()(
     hc: HeaderCarrier
   ): Future[Result] =
     customsCacheService.fetchAndGetEntry[DestinationCountriesSupplementary](cacheId, formId).map {
-      case Some(data) => Ok(destination_countries_supplementary(supplementaryForm.fill(data)))
-      case _          => Ok(destination_countries_supplementary(supplementaryForm))
+      case Some(data) => Ok(destination_countries_supplementary(supplementaryForm.fill(data), Countries.allCountries))
```
PatrykRudnicki: What do you think to move import to view directly?

conf/location-autocomplete-canonical-list.json
```diff
@@ -1,280 +1,236 @@
...
-  ],
-  [
-    "Andorra",
-    "country:AD"
+    "Armenia",
+    "country:AM"
   ],
   [
-    "Angola",
+    "Angola including Cabinda",
```
PauloPortugal: Wow!
That seems a bit harsh.
This is a massive political statement.
A bit of political context here:
Cabinda is an exclave and province of Angola, surrounded by Democratic Republic of the Congo, they are the richest province of Angola and are responsibles for the majority of oil production.
They are effectively the motor of the Angolan economy and have suggested in the past the will to become independent, although that would probably cause a civil war followed by a war with DRC.

It is the first time I see mentioned `Angola including Cabinda`

app/controllers/declaration/DestinationCountriesController.scala
ed0906: Done

test/views/declaration/spec/ViewSpec.scala
```diff
@@ -41,6 +42,8 @@ trait ViewSpec extends PlaySpec with GuiceOneAppPerSuite with ViewValidator {
   implicit lazy val fakeRequest: FakeRequest[AnyContentAsEmpty.type] = FakeRequest("", "")
   implicit lazy val messages: Messages = messagesApi.preferred(FakeRequest())
   implicit lazy val flash: Flash = new Flash()
+  
+  
   implicit lazy val countries: List[Country] = Countries.allCountries
```
PatrykRudnicki: Probably this line is incorrect. We don't need to pass countries as implicit.

ed0906: Removed

PR #280, opened 2019-06-25
CEDS-1044 Fixes for wrong mapping

PatrykRudnicki:


PauloPortugal:
As long as the `metadata xml` payload is updated I am fine with it :)

PR #281, opened 2019-06-25
CEDS-942 New model and notification sorting

MaciejRewera:


app/models/declaration/submissions/CancellationStatus.scala
```diff
@@ -0,0 +1,63 @@
...
+  implicit object CancellationStatusReads extends Reads[CancellationStatus] {
+    def reads(jsValue: JsValue): JsResult[CancellationStatus] = jsValue match {
+      case JsString("CancellationRequestExists") => JsSuccess(CancellationRequestExists)
+      case JsString("CancellationRequested")     => JsSuccess(CancellationRequested)
+      case JsString("MissingDeclaration")        => JsSuccess(MissingDeclaration)
+      case _                                     => JsError("Incorrect cancellation status")
+    }
+  }
+
+  implicit object CancellationStatusWrites extends Writes[CancellationStatus] {
```
ed0906: I think we could pull out some generic “enum format” code here which can be reused

app/models/declaration/submissions/CancellationStatus.scala
```diff
@@ -0,0 +1,63 @@
...
+import play.api.libs.json._
+
+sealed trait CancellationStatus
+
+object CancellationStatus {
+
+  def unapply(status: CancellationStatus): Option[(String, JsValue)] = {
+    val (prod: Product, sub) = status match {
+      case CancellationRequestExists =>
+        (CancellationRequestExists, Json.toJson(CancellationRequestExists.toString))
```
ed0906: Maybe case x => (x, JsString(x.toString)) so we don’t need a case for each option

app/models/declaration/submissions/SubmissionStatus.scala
```diff
@@ -85,56 +85,77 @@ object Status {
       case "18"                                                                  => AwaitingExitResults
       case _                                                                     => UnknownStatus
     }
+
+  def retrieve(functionCode: String, nameCode: Option[String]): SubmissionStatus =
```
ed0906: If SubmissionStatus had a value method (or something) that had to be overridden we could set the “value” to the below

app/models/declaration/submissions/CancellationStatus.scala
MaciejRewera: Could you explain in more detail what do you mean? In case of case objects it is tricky to create reads and writes.

app/models/declaration/submissions/CancellationStatus.scala
MaciejRewera: This is the same as previous one: case objects are a bit tricky and you can't avoid multiple cases. At least I am not aware of any way to do it differently.

app/models/declaration/submissions/SubmissionStatus.scala
MaciejRewera: I agree and I tried it in the morning. Complication appears around 2 statuses which require nameCode field. This would require additional thought. And if the solution is found, the backend should be updated as well.

MaciejRewera: Found working solution for this now. The commit follows.

PR #282, opened 2019-06-25
Ceds 1132

benjaminparker:


test/views/declaration/CarrierDetailsViewSpec.scala
```diff
@@ -101,8 +101,8 @@ class CarrierDetailsViewSpec extends ViewSpec with CarrierDetailsMessages with C
 
       val view = createView()
 
-      getElementById(view, "details_address_country-label").text() must be(messages(country))
-      getElementById(view, "details.address.country").attr("value") must be("")
+      getElementById(view, "details.address.country-label").text() mustBe "Country"
```
PatrykRudnicki: We decided to stay with messages here, testing specific message is in the different case. It will help fixing tests when something is broken

PR #286, opened 2019-06-26
CEDS-1132 Fix Country AutoComplete

ed0906:


app/services/model/AutoCompleteItem.scala
```diff
@@ -20,5 +20,5 @@ case class AutoCompleteItem(label: String, value: String)
 
 object AutoCompleteItem {
   def from(countries: List[Country], value: Country => String = _.countryName): List[AutoCompleteItem] =
-    countries map (c => AutoCompleteItem(c.countryName, value(c)))
+    countries map (c => AutoCompleteItem(s"${c.countryName} - ${c.countryCode}", value(c)))
```
PauloPortugal: This Scala syntax sugar is not what we tend to use here.
Fine with me now but let's discuss this

app/views/components/autocomplete_field.scala.html
```diff
@@ -74,7 +74,7 @@
     <select class="form-control" id="@(fieldNameWithoutBrackets)" name="@field.name" value="@field.value">
         <option value="">@emptySelectValue</option>
         @items.map { i =>
-            <option value="@i.value" @if(selectedValue == i.value) {selected="selected"}>@i.label - @i.value</option>
+            <option value="@i.value" @if(selectedValue == i.value) {selected="selected"}>@i.label</option>
```
PauloPortugal: :+1:

PR #287, opened 2019-06-26
CEDS-1303: add reactivemongo cache model

mattclark-zerogravit:
Added test for cache service and reverted auth code changes
Added cache to submit on initial 5 controllers

### DIFF
--- a/app/controllers/ChoiceController.scala
+++ b/app/controllers/ChoiceController.scala
@@ -28,6 +28,7 @@ import forms.Choice.AllowedChoiceValues._
 import forms.Choice._
 import handlers.ErrorHandler
 import javax.inject.Inject
+import play.api.Logger
 import play.api.data.Form
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
@@ -39,20 +40,22 @@ import views.html.choice_page
 import scala.concurrent.{ExecutionContext, Future}
 
 class ChoiceController @Inject()(
-                                  authenticate: AuthAction,
-                                  customsCacheService: CustomsCacheService,
-                                  exportsCacheService: ExportsCacheService,
-                                  errorHandler: ErrorHandler,
-                                  mcc: MessagesControllerComponents
-                                )(implicit ec: ExecutionContext, appConfig: AppConfig)
-  extends {
-    val cacheService = exportsCacheService
-  } with FrontendController(mcc) with I18nSupport with ModelCacheable with SessionIdAware {
+  authenticate: AuthAction,
+  customsCacheService: CustomsCacheService,
+  exportsCacheService: ExportsCacheService,
+  errorHandler: ErrorHandler,
+  mcc: MessagesControllerComponents
+)(implicit ec: ExecutionContext, appConfig: AppConfig)
+    extends {
+  val cacheService = exportsCacheService
+} with FrontendController(mcc) with I18nSupport with ModelCacheable with SessionIdAware {
+
+  val logger = Logger.apply(this.getClass)
 
   def displayChoiceForm(): Action[AnyContent] = authenticate.async { implicit request =>
     customsCacheService.fetchAndGetEntry[Choice](eoriCacheId, choiceId).map {
       case Some(data) => Ok(choice_page(Choice.form().fill(data)))
-      case _ => Ok(choice_page(Choice.form()))
+      case _          => Ok(choice_page(Choice.form()))
     }
   }
 
@@ -61,26 +64,35 @@ class ChoiceController @Inject()(
       .bindFromRequest()
       .fold(
         (formWithErrors: Form[Choice]) => Future.successful(BadRequest(choice_page(formWithErrors))),
-        validChoice =>
-          customsCacheService.cache[Choice](eoriCacheId, choiceId, validChoice).map { _ =>
-            validChoice.value match {
-              case SupplementaryDec | StandardDec =>
+        validChoice => {
 
-                val createdDateTime = LocalDateTime.now
-                val model = ExportsCacheModel(authenticatedSessionId, UUID.randomUUID().toString, createdDateTime, createdDateTime, validChoice.value)
-                exportsCacheService.update(authenticatedSessionId, model)
-
-                Redirect(controllers.declaration.routes.DispatchLocationPageController.displayPage())
-              case CancelDec =>
-                Redirect(controllers.routes.CancelDeclarationController.displayForm())
-              case Submissions =>
-                Redirect(controllers.routes.SubmissionsController.displayListOfSubmissions())
-              case _ =>
-                Redirect(controllers.routes.ChoiceController.displayChoiceForm())
-            }
+          (for {
+            _ <- customsCacheService.cache[Choice](eoriCacheId, choiceId, validChoice)
+            _ <- exportsCacheService.update(
+              authenticatedSessionId,
+              ExportsCacheModel(
+                authenticatedSessionId,
+                UUID.randomUUID().toString,
+                createdDateTime = LocalDateTime.now,
+                updatedDateTime = LocalDateTime.now,
+                validChoice.value
+              )
+            )
+          } yield ()).map {
+            _ =>
+              validChoice.value match {
+                case SupplementaryDec | StandardDec =>
+                  Redirect(controllers.declaration.routes.DispatchLocationPageController.displayPage())
+                case CancelDec =>
+                  Redirect(controllers.routes.CancelDeclarationController.displayForm())
+                case Submissions =>
+                  Redirect(controllers.routes.SubmissionsController.displayListOfSubmissions())
+                case _ =>
+                  Redirect(controllers.routes.ChoiceController.displayChoiceForm())
+              }
           }
+        }
       )
   }
-
 
 }
--- a/app/controllers/ChoiceController.scala
+++ b/app/controllers/ChoiceController.scala
@@ -16,8 +16,12 @@
 
 package controllers
 
+import java.time.LocalDateTime
+import java.util.UUID
+
 import config.AppConfig
 import controllers.actions.AuthAction
+import controllers.declaration.{ModelCacheable, SessionIdAware}
 import controllers.util.CacheIdGenerator.eoriCacheId
 import forms.Choice
 import forms.Choice.AllowedChoiceValues._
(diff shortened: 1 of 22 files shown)


app/controllers/declaration/AdditionalDeclarationTypePageController.scala
```diff
@@ -61,13 +59,25 @@ class AdditionalDeclarationTypePageController @Inject()(
...
+            .map(updateCache(_, validAdditionalDeclarationType))
+
           customsCacheService
             .cache[AdditionalDeclarationType](cacheId, decType.formId, validAdditionalDeclarationType)
             .map(_ => Redirect(controllers.declaration.routes.ConsignmentReferencesController.displayPage()))
+        }
       )
   }
 
+  private def updateCache(sessionId: String, formData: AdditionalDeclarationType): Future[Option[Future[Either[String, JourneyCacheModel]]]] = {
```
PauloPortugal: @PatrykRudnicki @ed0906 @benjaminparker is there a way of making this method `generic`, so that we do not need to  repeat ourselves. 

This is the line that would be nice to generalise 
```
model.copy(additionalDeclarationType = Some(formData))
```

mattclark-zerogravit: We have made this more generic by adding ModelCacheable Trait

app/controllers/ChoiceController.scala
```diff
@@ -36,10 +41,13 @@ import scala.concurrent.{ExecutionContext, Future}
 class ChoiceController @Inject()(
   authenticate: AuthAction,
   customsCacheService: CustomsCacheService,
+  exportsCacheService: ExportsCacheService,
   errorHandler: ErrorHandler,
   mcc: MessagesControllerComponents
 )(implicit ec: ExecutionContext, appConfig: AppConfig)
-    extends FrontendController(mcc) with I18nSupport {
+    extends {
+  val cacheService = exportsCacheService
```
PatrykRudnicki: Why you're assisgning this to cache service here? below you're using `exportsCacheService`

app/controllers/declaration/AdditionalDeclarationTypePageController.scala
```diff
@@ -41,9 +36,12 @@ class AdditionalDeclarationTypePageController @Inject()(
   authenticate: AuthAction,
   journeyType: JourneyAction,
   customsCacheService: CustomsCacheService,
+  exportsCacheService: ExportsCacheService,
   mcc: MessagesControllerComponents
 )(implicit appConfig: AppConfig, ec: ExecutionContext)
-    extends FrontendController(mcc) with I18nSupport {
+    extends {
+  val cacheService = exportsCacheService
```
PatrykRudnicki: The same here

app/controllers/ChoiceController.scala
PatrykRudnicki: Discussed, this val is necessary for trait

test/services/cache/ExportsCacheModelRepositorySpec.scala
```diff
@@ -0,0 +1,81 @@
...
+        val model = createModel(sessionId)
+        val result = repo.upsert(sessionId, model).futureValue
+        result must be(Some(model))
+      }
+
+    }
+
+    "on get" should {
+      "return Right with model when exists" in {
+        val sessionId = "12345"
```
PatrykRudnicki: I think we can extract this outside every test and define in one place before cases

mattclark-zerogravit: Have made this change

app/controllers/ChoiceController.scala
```diff
@@ -57,6 +65,16 @@ class ChoiceController @Inject()(
...
               case SupplementaryDec | StandardDec =>
+                val createdDateTime = LocalDateTime.now
+                val model = ExportsCacheModel(
+                  authenticatedSessionId,
+                  UUID.randomUUID().toString,
+                  createdDateTime,
+                  createdDateTime,
+                  validChoice.value
+                )
+                exportsCacheService.update(authenticatedSessionId, model)
```
ed0906: Does this return a future? + if it does we aren't using it

app/controllers/declaration/AdditionalDeclarationTypePageController.scala
```diff
@@ -59,12 +57,13 @@ class AdditionalDeclarationTypePageController @Inject()(
       .form()
       .bindFromRequest()
       .fold(
-        (formWithErrors: Form[AdditionalDeclarationType]) =>
-          Future.successful(BadRequest(declaration_type(formWithErrors))),
-        validAdditionalDeclarationType =>
+        formWithErrors => Future.successful(BadRequest(declaration_type(formWithErrors))),
+        validAdditionalDeclarationType => {
+          updateCache(journeySessionId, validAdditionalDeclarationType)
```
ed0906: Doesn't look like the response (future) of `updateCache` is ever used. I appreciate this is intentional, we could maybe make it more obvious though for when we change over e.g.
```
for {
_ <- updateCache(...)
_ <- customsCacheService.cache(...).map(
} yield Redirect(...)
```

ed0906: This would mean we wait until *both*  futures have finished before sending the Redirect. At the moment we don't wait until the new `updateCache` finishes, so it could fail and we'd never know

app/controllers/declaration/ItemsSummaryController.scala
```diff
@@ -60,4 +62,17 @@ class ItemsSummaryController @Inject()(
       case Some(data) => data.onwardSupplyRelief == FiscalInformation.AllowedFiscalInformationAnswers.yes
       case None       => false
     }
+
+  def addItem(): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    sessionId.map(updateCache(_, ExportItem()))
```
ed0906: Similar to other comments, the `Option[Future]`  is never used
```
sessionId.match {
  case Some(id) => updateCache(id, ExportItem()))
  case None  => Future.successful()
} map (_ =>  Redirect(...))
```

app/services/cache/ExportsCacheService.scala
```diff
@@ -0,0 +1,45 @@
...
+@Singleton
+class ExportsCacheService @Inject()(journeyCacheModelRepo: ExportsCacheModelRepository)(implicit ec: ExecutionContext) {
+  private val logger = Logger(this.getClass)
+
+  def get(sessionId: String): Future[Either[String, ExportsCacheModel]] = journeyCacheModelRepo.get(sessionId)
+
+  def update(sessionId: String, model: ExportsCacheModel): Future[Either[String, ExportsCacheModel]] = {
+    journeyCacheModelRepo.upsert(sessionId, model.copy(updatedDateTime = now())).map {
+      case Some(retrievedModel) => Right(retrievedModel)
+      case None                 => Left(s"Unable to retrieve a model for session id $sessionId")
```
ed0906: As discussed, seems a little strange to use Either when we have:
- `Future.failed` if  something shouldn't happen
- `Option` if something realistically might not be there

I'm not 100% sure on the context this gets used in, but if its fine for the model not to exist (i.e means a timeout or something) then I guess `Either` is ok

app/services/cache/ExportsCacheService.scala
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }.recover {
+    case exception: Throwable =>
```
ed0906: Any reason why we don't let this bubble up to the top?

ed0906:
I've added a few comments (sorry!), I **really** think we should be using the Future responses, it'll be really easy to miss them when we fully migrate over to the new cache, if we use the responses then we are forced to fix things

app/controllers/declaration/ItemsSummaryController.scala
ed0906: Also with this one, technically the redirect and subsequent page load could happen before the cache update future completes. So I could fill in the form, click add, and the page could load without the updated content

app/services/cache/ExportsCacheService.scala
mattclark-zerogravit: None and I agree we don't really need to handle this unless we have a requirement to handle exceptional circumstances in a nicer way. i.e. if mongo is down or if we are not able to upsert a document for some reason what would we actually do? I'll remove.

app/controllers/ChoiceController.scala
mattclark-zerogravit: This is probably not fully implemented i.e. we should probably handle the future and check it was ok / RIght then redirect. I'll make that change now as I agree it is confusing looking at it as it stands.

app/controllers/declaration/AdditionalDeclarationTypePageController.scala
mattclark-zerogravit: yep agreed, will make this change to the other controllers as well

app/controllers/declaration/ExporterDetailsPageController.scala
```diff
@@ -52,10 +56,18 @@ class ExporterDetailsPageController @Inject()(
       .fold(
         (formWithErrors: Form[ExporterDetails]) =>
           Future.successful(BadRequest(exporter_details(appConfig, formWithErrors))),
-        form =>
+        form => {
+          updateCache(journeySessionId, form)
```
ed0906: Similar

app/controllers/declaration/ProcedureCodesPageController.scala
```diff
@@ -71,16 +74,28 @@ class ProcedureCodesPageController @Inject()(
             Future.successful(BadRequest(procedure_codes(appConfig, formWithErrors, cache.additionalProcedureCodes))),
           validForm => {
             actionTypeOpt match {
-              case Some(Add)             => addAnotherCodeHandler(validForm, cache)
-              case Some(SaveAndContinue) => saveAndContinueHandler(validForm, cache)
-              case Some(Remove(values))  => removeCodeHandler(retrieveProcedureCode(values), cache)
-              case _                     => errorHandler.displayErrorPage()
+              case Some(Add) => addAnotherCodeHandler(validForm, cache)
+              case Some(SaveAndContinue) => {
+                updateCache(journeySessionId, validForm)
```
ed0906: similar

app/controllers/declaration/ItemsSummaryController.scala
mattclark-zerogravit: we have addressed this

app/controllers/declaration/ExporterDetailsPageController.scala
mattclark-zerogravit: Have addressed

app/controllers/declaration/ProcedureCodesPageController.scala
mattclark-zerogravit: Have addressed

ed0906:
Smash it in

PR #290, opened 2019-06-28
Fix and cleanup of startup script and rename to make more intuititive

benjaminparker:


run.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sm --start CUSTOMS_DECLARE_EXPORTS_FRONTEND_DEPS -f && sbt run
```
ed0906: Don't you need the assets part?

PR #292, opened 2019-06-28
CEDS-1340 Fix "Unknown status" issue + add missing Submissions View Spec

ed0906:


app/views/submissions.scala.html
```diff
@@ -27,6 +27,8 @@
     appConfig = appConfig,
     bodyClasses = None) {
 
+    @components.back_link("choice")
```
PatrykRudnicki: To be consistent with other pages my propose is to use `/customs-declare-exports/choice` as a back link

Edit: or like below... with `routes.ChoiceController.displayChoiceForm().url` if it's possible. This option looks good

test/helpers/views/declaration/SubmissionsMessages.scala
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package helpers.views.declaration
+
+trait SubmissionsMessages {
+
+  private val prefix: String = "submissions"
```
PatrykRudnicki: I think that we can start inlining prefix in every message as someone mentioned on some meeting. It's not a huge change and probably we don't discuss that small parts with bigger audience.

app/views/submissions.scala.html
ed0906: I didnt realise the argument was a href, I'll fix

PR #293, opened 2019-07-01
CEDS-1044 Fix for adding fiscal references when user choose No

PatrykRudnicki:


app/services/mapping/governmentagencygoodsitem/GovernmentAgencyGoodsItemBuilder.scala
```diff
@@ -37,6 +37,7 @@ object GovernmentAgencyGoodsItemBuilder {
       .toList
       .asJava
 
+  //scalastyle:off method.length
```
PauloPortugal: :(

PatrykRudnicki: Yeah, I know, but it's better to read file that it's not underlined :( we need to solve and avoid really long methods

PR #294, opened 2019-07-01
CEDS 1187

ed0906:


app/services/model/AutoCompleteItem.scala
```diff
@@ -29,4 +29,8 @@ object AutoCompleteItem {
 
   def fromDocumentType(documents: List[DocumentType]): List[AutoCompleteItem] =
     documents map (d => AutoCompleteItem(s"${d.description} - ${d.code}", d.code))
+
+  def fromNationalAdditionalCode(codes: List[NationalAdditionalCode]): List[AutoCompleteItem] = {
+    codes map (c => AutoCompleteItem(c.value, c.value))
+  }
```
PatrykRudnicki: Only style comment, but unnecessary parenthesis. It'll be good to keep convention like above for similar methods.

app/views/components/fields/field_multiple_values.scala.html
```diff
@@ -0,0 +1,65 @@
...
+    @if(elements.nonEmpty) {
+        <div class="field-group">
+            <table >
+            @elements.zipWithIndex.map { case (elem, index) =>
+                <tr>
+                    <th>@elem</th>
+                    <th>
+                        <button
+                            class="button--secondary"
+                            name="@Remove.toString"
```
PatrykRudnicki: I'm not sure about it but it's better to ask than skip. I think we should wrap `Remove.toString` in parenthesis. Like "@{Remove.toString}"

ed0906:
For anyone doing autocomplete stories, I've changed the name of the component to `field_autocomplete` and moved it into the `components.fields` package

app/utils/validators/forms/supplementary/ItemTypeValidator.scala
```diff
@@ -67,14 +68,13 @@ object ItemTypeValidator extends Validator[ItemType] {
...
+        "declaration.itemType.nationalAdditionalCode.error.invalid",
+        isContainedIn(NationalAdditionalCode.all.map(_.value))
       )
-      .verifying("declaration.itemType.nationalAdditionalCode.error.specialCharacters", isAlphanumeric)
   ).verifying(
-      "declaration.itemType.nationalAdditionalCode.error.maxAmount",
-      codes => codes.size <= nationalAdditionalCodesMaxAmount
-    )
+    "declaration.itemType.nationalAdditionalCode.error.maxAmount",
+    codes => codes.size <= nationalAdditionalCodesMaxAmount
```
ed0906: Not sure this is technically needed as there isnt even 99 codes to use but I've left it in incase the csv grows in the future

app/views/components/fields/field_multiple_values.scala.html
ed0906: Fixed

app/services/model/AutoCompleteItem.scala
ed0906: Fixed

PR #299, opened 2019-07-02
CEDS 1283 - Holder of Authorisation AutoComplete

ed0906:
As part of this I've moved all code-lists into a code list directory as the conf directory was getting a little crowded


test/config/AppConfigSpec.scala
```diff
@@ -150,11 +150,11 @@ class AppConfigSpec extends CustomExportsBaseSpec {
     }
 
     "have countryCodesJsonFilename" in {
-      validConfigService.countryCodesJsonFilename must be("location-autocomplete-canonical-list.json")
+      validConfigService.countryCodesJsonFilename must be("code-lists/location-autocomplete-canonical-list.json")
```
benjaminparker: why are we having json source and a csv source for countries -

app/services/HolderOfAuthenticationCode.scala
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services
+
+import utils.FileUtil
+
+case class HolderOfAuthenticationCode(value: String)
```
benjaminparker: Is it a holder of AUTHENTICATION or holder of AUTHORISATION - a bit confused?

test/services/HolderOfAuthenticationCodeSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+package services
+
+import uk.gov.hmrc.play.test.UnitSpec
+
+class HolderOfAuthenticationCodeSpec extends UnitSpec {
+
+  "Holder of Authentication Code" should {
+    "read from file" in {
+      val codes = HolderOfAuthenticationCode.all
+      codes should contain(HolderOfAuthenticationCode("ACE")) // First in file
```
benjaminparker: Remind me to show you the scala test asserts for contains inOrder.... it's pretty cool and a bit stronger than this

test/config/AppConfigSpec.scala
ed0906: Beyond the scope of this story, we can raise a tech debt ticket if you like

app/services/HolderOfAuthenticationCode.scala
ed0906: Fixed

test/config/AppConfigSpec.scala
PatrykRudnicki: I agree with @ed0906 but moreove our countries are in json file so we should stay with json, we have support for csv and now when we add new stuff and we receive new stuff in csv we can use our new approach, but IMO we should support both things, read from Json and from csv, I'm not sure that we should convert all countries to csv, we're using this correctly.

app/services/HolderOfAuthenticationCode.scala
PatrykRudnicki: I think it's Holder of Authentication but inside contains Authorisation code. Just to clarify

test/services/HolderOfAuthorisationCodeSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package services
+
+import uk.gov.hmrc.play.test.UnitSpec
+
+class HolderOfAuthorisationCodeSpec extends UnitSpec {
+
+  "Holder of Authentication Code" should {
```
PatrykRudnicki: So here we can change too, from `Holder of Authentication` to `Holder of Authorisation`

ed0906: Fixed

PR #300, opened 2019-07-02
CEDS - Rename all Inputs to Fields and organise into separate package

ed0906:
We have some components that are *just* inputs (e.g.  `input_autocomplete`), some components that wrap these into a field with a label, errors etc (e.g. `field_autocomplete`) and some components that are actually fields but are called inputs (e.g. `input_text`). This PR renames all the `input_x` which should be `field_x` and moves them into the appropriate package


ed0906:
Not the most exiting PR I appreciate...

ed0906:
The build passes  even though  the check failed..

PR #301, opened 2019-07-02
CEDS-1287 Office Of Exit Autocomplete

benjaminparker:
Note that supplementary views are not changed as discussed with business.  We only do this when its "for free" with shared code.


app/services/model/AutoCompleteItem.scala
```diff
@@ -16,23 +16,33 @@
...
+  def fromNationalAdditionalCode(codes: List[NationalAdditionalCode]): List[AutoCompleteItem] = {
     codes map (c => AutoCompleteItem(c.value, c.value))
+  }
 
-  def fromHolderOfAuthorisationCode(codes: List[HolderOfAuthorisationCode]): List[AutoCompleteItem] =
+  def fromHolderOfAuthorisationCode(codes: List[HolderOfAuthorisationCode]): List[AutoCompleteItem] = {
     codes map (c => AutoCompleteItem(c.value, c.value))
+  }
+
+  def fromOfficeOfExit(offices: List[OfficeOfExit]): List[AutoCompleteItem] = {
```
PatrykRudnicki: Is there any specific reason to add brackets to whole methods? I'm not a fan of unnecessary parenthesis.

I prefer to stay with previous version

test/services/PackageTypesSpec.scala
```diff
@@ -16,21 +16,24 @@
...
 
-class PackageTypeSpec extends UnitSpec {
+class PackageTypesSpec extends UnitSpec {
 
   "Package type list" should {
+    
     "return package types containing commas and quotes" in {
-      val somePackageTypes: List[PackageType] =
-        PackageType.all.filter((packageType: PackageType) => packageType.code == "43")
+      val somePackageTypes = PackageTypes.all.filter(_.code == "43")
```
PatrykRudnicki: 👍

test/services/OfficeOfExitsSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+class OfficeOfExitsSpec extends WordSpec with Matchers {
+
+  "OfficeOfExits" should {
+
+    "read values from CSV and order by description, alphabetically ascending" in {
+      OfficeOfExits.all should contain inOrder(
+        OfficeOfExit("GB000411", "Aberdeen Airport"),
+        OfficeOfExit("GB000060", "Dover/Folkestone Eurotunnel Freight"),
+        OfficeOfExit("GB003280", "Workington")
+      )
```
ed0906: Maybe a test to force the `.tail` that you use? (filtering out the header line)

test/views/declaration/OfficeOfExitStandardViewSpec.scala
```diff
@@ -54,114 +54,106 @@ class OfficeOfExitStandardViewSpec extends ViewSpec with OfficeOfExitMessages wi
...
       "display page title" in {
 
-        getElementByCss(createView(), "title").text() must be(messages(title))
+        getElementByCss(createView(), "title").text() mustBe messages(title)
       }
 
       "display section header" in {
 
-        getElementById(createView(), "section-header").text() must be("Locations")
+        getElementById(createView(), "section-header").text() mustBe "Locations"
```
ed0906: Just my opinion but I think we should avoid doing changes like this as it costs time without really any added value. No need to revert as what's done is done

app/services/model/AutoCompleteItem.scala
ed0906: I guess for consistency with the other methods in the class, whats done is done now I guess,  no point reverting in my view

test/services/PackageTypesSpec.scala
ed0906: My bad, I did this for Nicks benefit, forgot to remove

ed0906:
I've highlighted one missing test but otherwise looks good

app/services/model/AutoCompleteItem.scala
benjaminparker: Will leave - I did it for consistency

test/views/declaration/OfficeOfExitStandardViewSpec.scala
benjaminparker: I guess this is one to discuss - removing brackets makes tests much easier to read imho.

app/services/model/AutoCompleteItem.scala
```diff
@@ -16,23 +16,33 @@
...
 
-  def fromDocumentType(documents: List[DocumentType]): List[AutoCompleteItem] =
+  def fromDocumentType(documents: List[DocumentType]): List[AutoCompleteItem] = {
     documents map (d => AutoCompleteItem(s"${d.description} - ${d.code}", d.code))
+  }
 
-  def fromNationalAdditionalCode(codes: List[NationalAdditionalCode]): List[AutoCompleteItem] =
+  def fromNationalAdditionalCode(codes: List[NationalAdditionalCode]): List[AutoCompleteItem] = {
     codes map (c => AutoCompleteItem(c.value, c.value))
+  }
```
PatrykRudnicki: I'll still vote for removing this. It's not for consistency. All methods before didn't have bracekts, so I don't see any reason to stay with brackets if everyone agree that those methods should live without it.

Some Intellij configuration will pointed those lines as warnings (IntelliJ not sbt) and there will be proposal to remove brackets. I don't block this because of the brackets, but I'm sure that this will be changed by someone in the nearest future to the old implementation.

We're trying to remove brackets from tests, to make it more readable, shorter etc and adding in places that not need it, I don't see any pros of this.

PR looks good, so from me LGTM but without approving

PR #302, opened 2019-07-02
CEDS-1344 Create new cache model for Parties section

PauloPortugal:
* Create 'Parties' section inside the Cache Model  …
* Add remaining 'Parties' pages to the cache model  …
* Add new cache model to WarehouseIdentification  …
* Apply new Cache model on BorderTranport Controller
* Add new cache model to TransportDetails page  …
* Create new cache model for containers data  …

### DIFF
--- a/app/controllers/declaration/BorderTransportController.scala
+++ b/app/controllers/declaration/BorderTransportController.scala
@@ -28,6 +28,7 @@ import play.api.data.Form
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import services.CustomsCacheService
+import services.cache.{ExportsCacheModel, ExportsCacheService}
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import views.html.declaration.border_transport
 
@@ -38,9 +39,12 @@ class BorderTransportController @Inject()(
   journeyType: JourneyAction,
   errorHandler: ErrorHandler,
   customsCacheService: CustomsCacheService,
+  exportsCacheService: ExportsCacheService,
   mcc: MessagesControllerComponents
 )(implicit ec: ExecutionContext, appConfig: AppConfig)
-    extends FrontendController(mcc) with I18nSupport {
+    extends {
+  val cacheService = exportsCacheService
+} with FrontendController(mcc) with I18nSupport with ModelCacheable with SessionIdAware {
 
   def displayForm(): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
     customsCacheService
@@ -54,10 +58,15 @@ class BorderTransportController @Inject()(
       .fold(
         (formWithErrors: Form[BorderTransport]) => Future.successful(BadRequest(border_transport(formWithErrors))),
         borderTransport =>
-          customsCacheService
-            .cache[BorderTransport](cacheId, BorderTransport.formId, borderTransport)
-            .map(_ => Redirect(TransportDetailsController.displayForm()))
+          for {
+            _ <- updateCache(journeySessionId, borderTransport)
+            _ <- customsCacheService.cache[BorderTransport](cacheId, BorderTransport.formId, borderTransport)
+          } yield Redirect(TransportDetailsController.displayForm())
       )
   }
 
+  private def updateCache(sessionId: String, formData: BorderTransport): Future[Either[String, ExportsCacheModel]] =
+    updateHeaderLevelCache(sessionId, model => {
+      exportsCacheService.update(sessionId, model.copy(borderTransport = Some(formData)))
+    })
 }
--- a/app/controllers/declaration/BorderTransportController.scala
+++ b/app/controllers/declaration/BorderTransportController.scala
@@ -66,7 +66,8 @@ class BorderTransportController @Inject()(
   }
 
   private def updateCache(sessionId: String, formData: BorderTransport): Future[Either[String, ExportsCacheModel]] =
-    updateHeaderLevelCache(sessionId, model => {
-      exportsCacheService.update(sessionId, model.copy(borderTransport = Some(formData)))
-    })
+    updateHeaderLevelCache(
+      sessionId,
+      model => exportsCacheService.update(sessionId, model.copy(borderTransport = Some(formData)))
+    )
 }
--- a/app/controllers/declaration/RepresentativeDetailsPageController.scala
+++ b/app/controllers/declaration/RepresentativeDetailsPageController.scala
@@ -87,7 +87,7 @@ class RepresentativeDetailsPageController @Inject()(
     formData: RepresentativeDetails
   ): Future[Either[String, ExportsCacheModel]] =
     updateHeaderLevelCache(sessionId, model => {
-      val updatedParties = model.parties.map(_.copy(representativeDetails = Some(formData)))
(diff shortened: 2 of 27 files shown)


app/controllers/declaration/BorderTransportController.scala
```diff
@@ -54,10 +58,15 @@ class BorderTransportController @Inject()(
...
-            .map(_ => Redirect(TransportDetailsController.displayForm()))
+          for {
+            _ <- updateCache(journeySessionId, borderTransport)
+            _ <- customsCacheService.cache[BorderTransport](cacheId, BorderTransport.formId, borderTransport)
+          } yield Redirect(TransportDetailsController.displayForm())
       )
   }
 
+  private def updateCache(sessionId: String, formData: BorderTransport): Future[Either[String, ExportsCacheModel]] =
+    updateHeaderLevelCache(sessionId, model => {
```
ed0906: I see this method or similar a lot, is there any value in pulling it out somehow?

test/controllers/declaration/RepresentativeDetailsPageControllerSpec.scala
```diff
@@ -195,6 +197,8 @@ class RepresentativeDetailsPageControllerSpec
 
       route(app, postRequest(uri, correctRepresentativeDetailsJSON)).get.futureValue
 
+      verify(mockExportsCacheService).update(any(), any())
```
ed0906: Are we happy verifying on any? Maybe some argument captors to verify what we update? If that’s tested elsewhere in this class though it’s fine

ed0906: FYI a pattern I've seen in the past which I like is:

```
def theThingThatWasUpdated: Something = {
    val captor = ArgumentCaptor.forClass(classOf[Something])
    verify(service).methodName(captor.capture)
    captor.getValue
}
```

Then in tests you can do 
```
given(service.methodName(any())).willReturn(SomeResult)

theThingThatWasUpdated shouldBe Something(...)
```

As apposed to verifying on expected values e.g. `verify(service).methodName(Something(...))` which gets a little messy with bigger objects

app/controllers/declaration/TransportDetailsController.scala
```diff
@@ -56,14 +60,20 @@ class TransportDetailsController @Inject()(
...
       )
   }
 
   private def redirect(transportDetails: TransportDetails)(implicit request: JourneyRequest[_]): Result =
     if (transportDetails.container) Redirect(TransportContainerController.displayPage())
     else if (request.choice.value == AllowedChoiceValues.StandardDec) Redirect(SealController.displayForm())
     else Redirect(SummaryPageController.displayPage())
+
+  private def updateCache(sessionId: String, formData: TransportDetails): Future[Either[String, ExportsCacheModel]] =
+    updateHeaderLevelCache(sessionId, model => {
```
PatrykRudnicki: You don't need curly brackets here: `sessionId, model => {` it's comment for all similar methods.

app/controllers/declaration/BorderTransportController.scala
PauloPortugal: > I see this style of method or similar a lot, is there any value in pulling it out somehow?
I did not understand what do you mean? Are you talking about the `updateCache` method?

app/controllers/declaration/TransportDetailsController.scala
PauloPortugal: This has been addressed!

PauloPortugal: This has been addressed

app/controllers/declaration/BorderTransportController.scala
ed0906: Discussed with Paulo, we can leave it for now

app/controllers/declaration/BorderTransportController.scala
```diff
@@ -38,9 +39,12 @@ class BorderTransportController @Inject()(
   journeyType: JourneyAction,
   errorHandler: ErrorHandler,
   customsCacheService: CustomsCacheService,
+  exportsCacheService: ExportsCacheService,
   mcc: MessagesControllerComponents
 )(implicit ec: ExecutionContext, appConfig: AppConfig)
-    extends FrontendController(mcc) with I18nSupport {
+    extends {
+  val cacheService = exportsCacheService
```
ed0906: Why have we done this in the extends block?  you can just have `override val cacheService: ExportsCacheService` as a constructor arg

app/controllers/declaration/RepresentativeDetailsPageController.scala
```diff
@@ -75,4 +81,13 @@ class RepresentativeDetailsPageController @Inject()(
       case StandardDec =>
         controllers.declaration.routes.CarrierDetailsPageController.displayForm()
     }
+
+  private def updateCache(
+    sessionId: String,
+    formData: RepresentativeDetails
+  ): Future[Either[String, ExportsCacheModel]] =
+    updateHeaderLevelCache(sessionId, model => {
+      val updatedParties = model.parties.map(_.copy(representativeDetails = Some(formData)))
```
ed0906: Can we be sure there will be `Some(parties)`? Otherwise we lose this form data

app/services/cache/ExportsCacheModelRepository.scala
```diff
@@ -77,7 +78,11 @@ case class ExportsCacheModel(
   dispatchLocation: Option[DispatchLocation] = None,
   additionalDeclarationType: Option[AdditionalDeclarationType] = None,
   consignmentReferences: Option[ConsignmentReferences] = None,
-  exporterDetails: Option[ExporterDetails] = None,
+  warehouseIdentification: Option[WarehouseIdentification] = None,
+  borderTransport: Option[BorderTransport] = None,
+  transportDetails: Option[TransportDetails] = None,
+  containerData: Option[TransportInformationContainerData] = None,
+  parties: Option[Parties] = None,
```
ed0906: It looks to me like `Parties` is just  a container class for some Optional fields. 
Why don't we change this to `parties: Parties = Parties()`

PauloPortugal: Super Cool!!! :) Thanks 
Addressed!

app/controllers/declaration/RepresentativeDetailsPageController.scala
ed0906: fixed

PauloPortugal: Addressed! :)

app/controllers/declaration/BorderTransportController.scala
PauloPortugal: This was just a style that we decided to go with on our Mob session.

test/controllers/declaration/RepresentativeDetailsPageControllerSpec.scala
PauloPortugal: This as been addressed!

app/controllers/declaration/BorderTransportController.scala
ed0906: Fair enough, it's a little more verbose but does the same thing I guess.
I think a side effect will be that in your `BorderTransportController` you can access the `ExportsCacheService` using the variable name `cacheService` or `exportsCacheService`  which might be a little confusing

PR #304, opened 2019-07-03
CEDS 1346: added save to new caching on items pages in journey

mattclark-zerogravit:


app/controllers/declaration/AdditionalFiscalReferencesController.scala
```diff
@@ -40,83 +41,104 @@ class AdditionalFiscalReferencesController @Inject()(
   journeyType: JourneyAction,
   errorHandler: ErrorHandler,
   customsCacheService: CustomsCacheService,
+  exportsCacheService: ExportsCacheService,
   mcc: MessagesControllerComponents
 )(implicit appConfig: AppConfig, ec: ExecutionContext)
-    extends FrontendController(mcc) with I18nSupport {
+    extends {
+  val cacheService = exportsCacheService
```
ed0906: FYI You can just have `override val  cacheService: ExportsCacheService` as a constructor param. I think a side effect of using the extends block here is you you can reference your `ExportsCacheService` using the variable name `exportsCacheService` *and* `cacheService` which might be a little confusing

app/controllers/declaration/AdditionalFiscalReferencesController.scala
```diff
@@ -40,83 +41,104 @@ class AdditionalFiscalReferencesController @Inject()(
...
-    implicit request: JourneyRequest[_]
-  ): Result = BadRequest(additional_fiscal_references(formWithErrors, references))
+  private def badRequest(
+    itemId: String,
+    formWithErrors: Form[AdditionalFiscalReference],
+    references: Seq[AdditionalFiscalReference]
+  )(implicit request: JourneyRequest[_]): Result =
+    BadRequest(additional_fiscal_references(itemId, formWithErrors, references))
+
+  private def updateCacheModels(itemId: String, updatedCache: Seq[AdditionalFiscalReference], redirect: Call)(
```
ed0906: Minor but if you called a method called "updateCacheModels" would you expect it to return a `Redirect`?

I'd generally expect it to return an updated model or at the very least `Future[Unit]`

app/controllers/declaration/AdditionalInformationController.scala
```diff
@@ -39,64 +42,117 @@ class AdditionalInformationController @Inject()(
...
-              Future.successful(BadRequest(additional_information(appConfig, formWithErrors, cache.items))),
-            updatedCache =>
-              if (updatedCache != cache.items)
-                customsCacheService
-                  .cache[AdditionalInformationData](goodsItemCacheId, formId, AdditionalInformationData(updatedCache))
-                  .map(_ => Redirect(controllers.declaration.routes.DocumentsProducedController.displayForm()))
-              else
-                Future.successful(Redirect(controllers.declaration.routes.DocumentsProducedController.displayForm()))
-          )
+  private def updateCacheModelAndRedirect(itemId: String, updatedCache: Seq[AdditionalInformation], redirectCall: Call)(
```
ed0906: Just my personal view but generally having `and` in a method name suggests it's doing a little too much.

I think it would be nice if we could break this up into an `updateCache` method which just does the `for` comprehension and yields `Future[Unit]` which we then just map to a `Redirect`

app/controllers/declaration/CommodityMeasureController.scala
```diff
@@ -35,33 +37,61 @@ import scala.concurrent.{ExecutionContext, Future}
...
         validForm =>
-          cacheService.cache[CommodityMeasure](goodsItemCacheId, commodityFormId, validForm).map { _ =>
-            Redirect(controllers.declaration.routes.AdditionalInformationController.displayForm())
+          updateCacheModels(itemId, validForm).map { _ =>
+            Redirect(controllers.declaration.routes.AdditionalInformationController.displayPage(itemId))
         }
       )
   }
+
+  private def updateCacheModels(itemId: String, updatedCache: CommodityMeasure)(
```
ed0906: :+1

app/controllers/declaration/DocumentsProducedController.scala
```diff
@@ -42,119 +43,154 @@ class DocumentsProducedController @Inject()(
...
     hc: HeaderCarrier
   ): Future[Result] = keys.headOption.fold(errorHandler.displayErrorPage()) { index =>
     val updatedCache = cachedData.copy(documents = cachedData.documents.patch(index.toInt, Nil, 1))
-    customsCacheService.cache[DocumentsProducedData](goodsItemCacheId, formId, updatedCache).map { _ =>
-      Redirect(routes.DocumentsProducedController.displayForm())
-    }
+    updateCacheAndRedirect(itemId, updatedCache, routes.DocumentsProducedController.displayPage(itemId))
   }
 
+  private def updateCacheAndRedirect(itemId: String, documentsToUpdate: DocumentsProducedData, redirectCall: Call)(
```
ed0906: Similar to the above

app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -35,32 +38,54 @@ class FiscalInformationController @Inject()(
...
+        .bindFromRequest()
+        .fold(
+          (formWithErrors: Form[FiscalInformation]) =>
+            Future.successful(BadRequest(fiscal_information(itemId, formWithErrors))),
+          validFiscalInformation => updateCacheModels(itemId, validFiscalInformation)
+        )
   }
 
-  private def specifyNextPage(answer: FiscalInformation): Result =
+  private def updateCacheModels(itemId: String, validFiscalInformation: FiscalInformation)(
```
ed0906: Similar to comments above, I think it would be more intuitive if this just returns `Future[Unit]` rather than a `Future[Result]` given it is a method  called `updateCacheModels`

Minor though

app/controllers/declaration/ModelCacheable.scala
```diff
@@ -25,7 +25,12 @@ import scala.concurrent.{ExecutionContext, Future}
 trait ModelCacheable {
   val cacheService: ExportsCacheService
 
-  protected def updateHeaderLevelCache(
+  /*
+  Compilation warning
+    match may not be exhaustive.
+    [warn] It would fail on the following input: Left(_)
+   */
+  protected def getAndUpdateExportCacheModel(
```
ed0906: Whats this?

app/controllers/declaration/ModelCacheable.scala
```diff
@@ -25,7 +25,12 @@ import scala.concurrent.{ExecutionContext, Future}
 trait ModelCacheable {
   val cacheService: ExportsCacheService
 
-  protected def updateHeaderLevelCache(
+  /*
+  Compilation warning
+    match may not be exhaustive.
+    [warn] It would fail on the following input: Left(_)
+   */
```
ed0906: Whats this for?

app/views/declaration/additional_information.scala.html
```diff
@@ -20,14 +20,15 @@
...
+@import forms.declaration.CommodityMeasure
+@(itemId: String, appConfig: AppConfig, form: Form[AdditionalInformation], items: Seq[AdditionalInformation])(implicit request: Request[_], messages: Messages)
 
 @main_template(
     title = messages("supplementary.additionalInformation"),
     appConfig = appConfig
 ) {
 
-    @components.back_link("/customs-declare-exports/declaration/commodity-measure")
+    @components.back_link(CommodityMeasureController.displayPage(itemId).url)
```
ed0906: 👍

ed0906: I think IntelliJ struggles to compile it unless you stick `routes.` as a prefix e.g. `routes.CommodityMeasureController.displayPage(...)`

I think it gets confused wether you want the `controllers.CommodityMeasureController` or the `routes.CommodityMeasureController`.  

All works fine though as it is, ^^ just gets rid of all the red in IntelliJ

test/controllers/declaration/AdditionalInformationControllerSpec.scala
```diff
@@ -91,6 +103,8 @@ class AdditionalInformationControllerSpec
         val result = route(app, postRequestFormUrlEncoded(uri, body: _*)).get
 
         status(result) must be(SEE_OTHER)
+
+        verify(mockExportsCacheService, times(1)).update(any(), any())
```
ed0906: I'm pretty sure you  don't need the `times(1)` as *I think* that is the default for verify.

Also do  you think it's worth doing a more exhaustive verify given theres some mapping code that currently isn't tested?

test/controllers/declaration/AdditionalInformationControllerSpec.scala
```diff
@@ -442,10 +462,13 @@ class AdditionalInformationControllerSpec
         val body = Seq(("code", "M1l3s"), ("description", "Davis"), saveAndContinueActionURLEncoded)
 
         val result = route(app, postRequestFormUrlEncoded(uri, body: _*)).get
-        val header = result.futureValue.header
 
         status(result) must be(SEE_OTHER)
-        header.headers.get("Location") must be(Some("/customs-declare-exports/declaration/add-document"))
+        redirectLocation(result) must be(
+          Some(s"/customs-declare-exports/declaration/items/${cacheModel.items.head.id}/add-document")
```
ed0906: You can use the  play generated routes here too if you want

test/controllers/declaration/CommodityMeasureControllerSpec.scala
```diff
@@ -169,7 +177,7 @@ class CommodityMeasureControllerSpec
             val result = route(app, postRequestFormUrlEncoded(uri, payload: _*)).value
             status(result) must be(SEE_OTHER)
             result.futureValue.header.headers.get("Location") must be(
-              Some("/customs-declare-exports/declaration/additional-information")
+              Some(s"/customs-declare-exports/declaration/items/${cacheModel.items.head.id}/additional-information")
```
ed0906: You can use the play generated routes here as well if you want

app/controllers/declaration/ModelCacheable.scala
mattclark-zerogravit: I think Patryk added it as we are not covering the Left(_) case that could happen.
I'll add a case for the left scenario when we might not get a model returned from the cache.

app/controllers/declaration/FiscalInformationController.scala
mattclark-zerogravit: Swings and roundabouts... if we take the redirect out we end up with extra lines in the code. Might be better to rename the function but then you'd complain it's doing too much ;)

app/views/declaration/additional_information.scala.html
mattclark-zerogravit: This is not just my changes though, seen this elsewhere.. another ticket for the dev board?

app/controllers/declaration/AdditionalInformationController.scala
mattclark-zerogravit: I don't mind this as it removes a few more lines of code and the actual code in the function is not that long / complex.

app/controllers/declaration/AdditionalFiscalReferencesController.scala
mattclark-zerogravit: commented elsewhere about this.

app/controllers/declaration/AdditionalFiscalReferencesController.scala
mattclark-zerogravit: this is the approach we have taken across our caching work. Any changes to this we should cover in an overall discussion and change all controllers if we decide there is a better way of doing this. I.e. get rid of the modelcacheable trait and add the functions it has directly into the exportsCacheService as one suggestion.

app/controllers/declaration/FiscalInformationController.scala
mattclark-zerogravit: I'll rename the function so it reflects what it is doing.

PR #305, opened 2019-07-04
CEDS-1345 New Cache: Destination Countries

ed0906:
I ended up having to combine DestinationCountriesStandard and DestinationCountriesSupplementary as when we were retrieving it from the cache the  JSON deserialiser couldn't work out which type it should be (as they both have the same formId). Also seems a little strange to have two different classes when theres only 1 field thats different.

Figured it would be best to do this as a separate PR


app/forms/declaration/destinationCountries/DestinationCountries.scala
```diff
@@ -16,12 +16,63 @@
 
 package forms.declaration.destinationCountries
 
-import play.api.data.Form
+import play.api.data.Forms.{default, seq, text}
+import play.api.data.{Form, Forms, Mapping}
+import play.api.libs.json.{Json, OFormat}
+import services.Countries.allCountries
+
+case class DestinationCountries(
```
ed0906: Just to highlight I've combined the Supplementary and Standard form. 
Reasons for doing this are:
- Countries Standard wasnt shown on the summary page
- Fixing the above meant pulling CountriesStandard out of the cache which failed with Deserializer errors because that and CountriesSupplementary have the same  formID
- Theres only one field extra in standard so why not use the same class with different forms

app/controllers/declaration/DestinationCountriesController.scala
```diff
@@ -82,23 +84,24 @@ class DestinationCountriesController @Inject()(
...
+          } yield Redirect(controllers.declaration.routes.LocationController.displayForm())
       )
 
   private def handleSubmitStandard()(implicit request: JourneyRequest[AnyContent]): Future[Result] = {
     val actionTypeOpt = request.body.asFormUrlEncoded.map(FormAction.fromUrlEncoded)
 
     val cachedData = customsCacheService
-      .fetchAndGetEntry[DestinationCountriesStandard](cacheId, formId)
-      .map(_.getOrElse(DestinationCountriesStandard.empty))
+      .fetchAndGetEntry[DestinationCountries](cacheId, formId)
```
mattclark-zerogravit: This is changing the behaviour of the old caching which is risky and not the purpose of the caching approach. I'm not a fan of the changes made here in the fact that although the models/forms differ by one field we may in future decide to split the journey out to its own service. By coupling these we make it harder to see the journeys differ.

app/forms/declaration/destinationCountries/DestinationCountries.scala
mattclark-zerogravit: I'd be more tempted to fix the existing implementation as part of another ticket. Sounds like it needs fixing but I dont think combining these is the correct approach.

app/controllers/declaration/DestinationCountriesController.scala
ed0906: We cant convert this controller to the new cache until this or another approach to fixing this is done @mattclark-zerogravit. It only works currently because the "pulling the data out of the cache" bit wasn't implemented fully for the Standard Journey

ed0906: I'd like to understand better why you think this changes the behaviour @mattclark-zerogravit. All the original tests passed, so as far as i'm aware existing behaviour has been preserved

app/services/mapping/goodsshipment/DestinationBuilder.scala
```diff
@@ -38,14 +34,14 @@ object DestinationBuilder {
 
   def buildSupplementary(implicit cacheMap: CacheMap): GoodsShipment.Destination =
```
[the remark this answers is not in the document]
ed0906: Good spot i'll fix

app/services/cache/ExportsCacheModelRepository.scala
```diff
@@ -78,7 +78,11 @@ case class ExportsCacheModel(
   dispatchLocation: Option[DispatchLocation] = None,
   additionalDeclarationType: Option[AdditionalDeclarationType] = None,
   consignmentReferences: Option[ConsignmentReferences] = None,
-  exporterDetails: Option[ExporterDetails] = None,
+  warehouseIdentification: Option[WarehouseIdentification] = None,
```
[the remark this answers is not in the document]
ed0906: I always squash and merge anyway don't worry there wont be any unwanted commits :)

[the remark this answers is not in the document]
ed0906: @PauloPortugal I presume it was right for exportedDetails to be removed? The conflict suggested so but it's best I clarify!

PR #306, opened 2019-07-04
CEDS-1162: Supervising customs office & office of presentation (5/26 & 5/27)

benjaminparker:


app/services/model/CustomsOffice.scala
```diff
@@ -14,14 +14,6 @@
...
-trait SupervisingCustomsOfficeMessages {
-
-  val supervisingCustomOffice: String = "supplementary.supervisingCustomsOffice"
-
-  val title: String = supervisingCustomOffice + ".title"
-  val hint: String = supervisingCustomOffice + ".hint"
-  val scoError: String = supervisingCustomOffice + ".error"
-  val scoEmpty: String = supervisingCustomOffice + ".empty"
-}
+case class CustomsOffice(code: String, description: String)
```
ed0906: Is this meant to be in test code?

ed0906: Ignore me, didnt noticed the package move too

app/services/view/AutoCompleteItem.scala
```diff
@@ -14,14 +14,15 @@
  * limitations under the License.
  */
 
-package services.model
+package services.view
```
ed0906: If we're moving this, i'd suggest it goes into the `views.components.inputs` package as its only relevant to the `input_radio` and stuff that uses that

test/controllers/declaration/WarehouseIdentificationControllerSpec.scala
```diff
@@ -46,34 +46,12 @@ class WarehouseIdentificationControllerSpec extends CustomExportsBaseSpec with W
       status(result) must be(OK)
     }
 
-    "display correct hints and questions" in {
```
ed0906: Have we agreed not to have these style tests?

PatrykRudnicki: It's inside controller. This is covered in view spec

app/services/view/AutoCompleteItem.scala
benjaminparker: services.views => views is usually only for .html.scala files
services.model is needed for the backend CSV parsing

Was discussed with Matt / Pat and agreed services.view ....

test/controllers/declaration/WarehouseIdentificationControllerSpec.scala
benjaminparker: Perhaps - I wasn't in the test meeting.  Can move to the view spec if required or remove?

PauloPortugal: These should be in the `WarehouseIdentificationViewSpec.scala`, which I think we already have, so this code is redundant.
Looks good.

PauloPortugal:
Please resolve conflicts

test/controllers/declaration/WarehouseIdentificationControllerSpec.scala
benjaminparker: It was testing code that was already in a ViewSpec

app/services/view/AutoCompleteItem.scala
ed0906: "views is usually only for .html.scala files" Just my view but anything view related should go in the `views` package, that includes AutoComplete, RadioOption and whatever other view helpers we have

ed0906: At the moment we're using the `services` package rather like a `utils` package

PR #317, opened 2019-07-10
CEDS-1370: dispatch controller new implementation of exports cache

PauloPortugal:
First approach

### DIFF
--- a/app/controllers/declaration/DispatchLocationPageController.scala
+++ b/app/controllers/declaration/DispatchLocationPageController.scala
@@ -45,9 +45,9 @@ class DispatchLocationPageController @Inject()(
 } with FrontendController(mcc) with I18nSupport with ModelCacheable with SessionIdAware {
 
   def displayPage(): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
-    exportsCacheService.get(journeySessionId).map {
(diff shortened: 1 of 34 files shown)


app/controllers/declaration/DispatchLocationPageController.scala
```diff
@@ -45,12 +45,10 @@ class DispatchLocationPageController @Inject()(
 } with FrontendController(mcc) with I18nSupport with ModelCacheable with SessionIdAware {
 
   def displayPage(): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
-    customsCacheService
-      .fetchAndGetEntry[DispatchLocation](cacheId, DispatchLocation.formId)
-      .map {
-        case Some(data) => Ok(dispatchLocationPage(DispatchLocation.form().fill(data)))
-        case _          => Ok(dispatchLocationPage(DispatchLocation.form()))
-      }
+    exportsCacheService.get(journeySessionId).map {
```
ed0906: Maybe  `.map(_.dispatchLocation)` before this map then `case  Right(Some(dispatchLocation))` below so we  don't need the `get`

ed0906: But it would  mean  you would end up on the empty form if the retrieve failed **or** returned nothing..... Might as well just use Option ;)

PR #320, opened 2019-07-11
CEDS-1372 Convert remaining cache retrieves

ed0906:


ed0906:
Precheck passes locally, I suspect the pr builder is being flakey again

PR #324, opened 2019-07-16
CEDS-1371: retrieve data from the new cache for the items pages

mattclark-zerogravit:


app/controllers/declaration/PackageInformationController.scala
```diff
@@ -77,42 +75,41 @@ class PackageInformationController @Inject()(
...
     if (!isFormEmpty(payload)) badRequest(itemId, packages, payload, USE_ADD)
-    else if (packages.size == 0) badRequest(itemId, packages, payload, ADD_ONE)
+    else if (packages.isEmpty) badRequest(itemId, packages, payload, ADD_ONE)
     else
       updateCacheModels(itemId, packages, controllers.declaration.routes.CommodityMeasureController.displayPage(itemId))
   }
 
   private def isFormEmpty[A](form: Form[A]): Boolean =
-    retrieveData(form).filter { case (_, value) => value.nonEmpty }.isEmpty
+    !retrieveData(form).exists { case (_, value) => value.nonEmpty }
```
PatrykRudnicki: 👍

test/controllers/declaration/AdditionalFiscalReferencesControllerSpec.scala
```diff
@@ -47,11 +49,15 @@ class AdditionalFiscalReferencesControllerSpec extends CustomExportsBaseSpec {
       val Some(result) = route(app, getRequest(uri))
 
       status(result) must be(OK)
+      verify(mockExportsCacheService).getItemByIdAndSession(any[String], any[String])
```
PatrykRudnicki: In `org.mockito.ArgumentMatchers` you can find method `anyString` it's probably better to use instead of `any[String]`

ed0906: Tbh I don't think theres much value in verifying stuff like that anyway, if we didn't call it the test wouldn't work

ed0906: I'll remove it

PR #325, opened 2019-07-16
CEDS-1370 Apply new Cache to Parties and Summary pages

PauloPortugal:
We still have a few old cache mocks there, that are need because we still
need to map it to the DMS schema. A subsequent story will take care of
this.

### DIFF
--- a/test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
+++ b/test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
@@ -39,6 +39,7 @@ class ConsignmentReferencesControllerSpec
   import ConsignmentReferencesControllerSpec._
 
   override def beforeEach() {
+    super.beforeEach()
     authorizedUser()
     withNewCaching(createModelWithNoItems())
     withCaching[ConsignmentReferences](None, ConsignmentReferences.id)
@@ -46,8 +47,8 @@ class ConsignmentReferencesControllerSpec
   }
 
   override def afterEach() {
-    reset(mockCustomsCacheService)
-    reset(mockExportsCacheService)
+    super.afterEach()
+    reset(mockCustomsCacheService, mockExportsCacheService)
   }
 
   "Consignment References Controller on GET" should {
--- a/test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
+++ b/test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
@@ -16,17 +16,20 @@
 
 package controllers.declaration
 
+import java.time.LocalDateTime
+
 import base.CustomExportsBaseSpec
-import forms.Choice
 import forms.Choice.choiceId
 import forms.declaration.ConsignmentReferences
 import forms.declaration.ConsignmentReferencesSpec._
+import forms.{Choice, Ducr}
 import helpers.views.declaration.{CommonMessages, ConsignmentReferencesMessages}
 import org.mockito.ArgumentMatchers
 import org.mockito.ArgumentMatchers.any
 import org.mockito.Mockito.{reset, verify}
 import play.api.libs.json.{JsObject, JsString, JsValue}
 import play.api.test.Helpers._
+import services.cache.ExportsCacheModel
 
 class ConsignmentReferencesControllerSpec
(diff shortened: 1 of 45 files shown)


test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
```diff
@@ -36,15 +39,13 @@ class ConsignmentReferencesControllerSpec
   import ConsignmentReferencesControllerSpec._
 
   override def beforeEach() {
-    super.beforeEach()
```
ed0906: Por que?

test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
```diff
@@ -36,15 +39,13 @@ class ConsignmentReferencesControllerSpec
...
     authorizedUser()
     withNewCaching(createModelWithNoItems())
     withCaching[ConsignmentReferences](None, ConsignmentReferences.id)
     withCaching[Choice](Some(Choice(Choice.AllowedChoiceValues.SupplementaryDec)), choiceId)
   }
 
   override def afterEach() {
-    super.afterEach()
```
ed0906: Think this should be there

test/controllers/declaration/DeclarantDetailsPageControllerSpec.scala
```diff
@@ -35,20 +40,37 @@ class DeclarantDetailsPageControllerSpec extends CustomExportsBaseSpec {
     withCaching[Choice](Some(Choice(Choice.AllowedChoiceValues.SupplementaryDec)), choiceId)
   }
 
+  override def afterEach() {
+    reset(mockExportsCacheService)
```
ed0906: `super.afterEach()`

otherwise if there is something defined it never gets called and you'll get all sorts of magical errors

test/controllers/declaration/DeclarationHolderControllerSpec.scala
```diff
@@ -43,18 +47,37 @@ class DeclarationHolderControllerSpec
     withCaching[Choice](Some(Choice(Choice.AllowedChoiceValues.SupplementaryDec)), choiceId)
   }
 
+  override def afterEach() {
+    reset(mockCustomsCacheService)
```
ed0906: Minor but you can do `reset(a, b, c, d....)` rather  than one on each line

test/controllers/declaration/SummaryPageControllerSpec.scala
```diff
@@ -35,67 +35,79 @@ import scala.concurrent.Future
...
     when(mockCustomsCacheService.fetch(anyString())(any(), any()))
       .thenReturn(Future.successful(Some(minimumValidCacheData)))
     when(mockCustomsCacheService.remove(anyString())(any(), any()))
       .thenReturn(Future.successful(HttpResponse(OK)))
-    successfulCustomsDeclareExportsResponse()
     withCaching[Choice](Some(Choice(Choice.AllowedChoiceValues.SupplementaryDec)), choiceId)
   }
 
+  override def afterEach() {
+    reset(mockCustomsCacheService)
```
ed0906: Similar, you can do `reset(a,b,c)`
Also theres a missing `super.afterEach()`

test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
PauloPortugal: _The Assassination of `super.beforeEach()` by the Coward @PauloPortugal_ 

`super.beforeEach()` is promoting inheritance, which with the advent of the functional paradigm has lost some of it's reputation, some languages are today even designed not allow inheritance (such as `Go` or `Clojure`, although they still allow polymorphism)

And that is why Scala loses the appeal. It allows you to write OOP and Functional all in the same place, a bit like Java 8-12.

This is also saying, we should not allow generic behaviour to be available in traits/parent classes.

PauloPortugal:
@ed0906 I have addressed all of your comments.

ed0906:
> @ed0906 I have addressed all of your comments.

Smash it in

PR #326, opened 2019-07-16
CEDS-1249 Approach without app with Stubs helper object

PatrykRudnicki:


PauloPortugal:
It looks OK. 
From a unit test POV, this is probably the better approach than Guice.

test/unit/controllers/StartControllerSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  "Start Controller" should {
+
+    "return 200" when {
+      
+      "display page method is invoked" in {
+
+        val result = controller.displayStartPage()(FakeRequest("", ""))
+
+        status(result) must be(OK)
```
mattclark-zerogravit: This is good. I like that the controller spec is now only testing what it should be testing i.e. the status return. No content as string, check what html is returned.

test/base/MockCustomsCacheService.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.mockito.Mockito.when
+import org.mockito.stubbing.OngoingStubbing
+import org.scalatest.mockito.MockitoSugar
+import play.api.data.Form
+import services.CustomsCacheService
+import uk.gov.hmrc.http.cache.client.CacheMap
+
+import scala.concurrent.Future
+
+trait MockCustomsCacheService extends MockitoSugar {
```
ed0906: Nice

test/unit/base/ControllerSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+import play.api.libs.json.JsValue
+import play.api.mvc.{AnyContentAsEmpty, AnyContentAsJson, Request}
+import play.api.test.FakeRequest
+import play.twirl.api.Html
+import unit.tools.Stubs
+import utils.FakeRequestCSRFSupport._
+
+import scala.concurrent.ExecutionContext
+
+trait ControllerSpec extends UnitSpec with Stubs with MockAuthAction with MockConnectors with MockCustomsCacheService {
```
ed0906: 👍

PR #329, opened 2019-07-17
CEDS-1382 Add Remove Item Button

ed0906:


app/controllers/declaration/ItemsSummaryController.scala
```diff
@@ -63,11 +63,22 @@ class ItemsSummaryController @Inject()(
       .map(_ => Redirect(controllers.declaration.routes.ProcedureCodesPageController.displayPage(newItem.id)))
   }
 
-  /*
-  Compilation warning
-    match may not be exhaustive.
-    [warn] It would fail on the following input: Left(_)
-   */
+  def removeItem(itemId: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    exportsCacheService.get(journeySessionId) flatMap {
```
mattclark-zerogravit: could we use the get item by id and sessionid here so we can do everything in one go?

ed0906: Afraid not, that method just  returns the Export Item when I  need the whole cache object to be able to do the update

ed0906: This kind of logic could potentially be pulled down into the service layer though I guess

mattclark-zerogravit: yeah maybe. I suppose if you get the item.... then get the cachemodel to update its and extra db call. Your approach avoids that.

conf/declaration.routes
```diff
@@ -117,7 +117,9 @@ POST        /nature-of-transaction               controllers.declaration.NatureO
 
 GET         /export-items                        controllers.declaration.ItemsSummaryController.displayPage()
 
-GET         /export-items/add                                  controllers.declaration.ItemsSummaryController.addItem()
+GET         /export-items/add                    controllers.declaration.ItemsSummaryController.addItem()
+
+GET         /export-items/:itemId/remove         controllers.declaration.ItemsSummaryController.removeItem(itemId)
```
MaciejRewera: IMHO it would be 'cleaner' to use DELETE in this place.

ed0906: Browser can only send POST and GET

PR #330, opened 2019-07-18
CEDS-1383 Add TTL Index based on last updated time

ed0906:


app/services/cache/ExportsCacheModelRepository.scala
```diff
@@ -88,5 +98,21 @@ case class ExportsCacheModel(
 )
 
 object ExportsCacheModel {
-  implicit val format = Json.format[ExportsCacheModel]
+  implicit val formatInstant: OFormat[LocalDateTime] = new OFormat[LocalDateTime] {
```
ed0906: This will mean we have to clear the DB down

ed0906: As  discussed in other PRs we shouldn't really be using LocalDateTime as it doesnt contain time zone info. Instant would be preferred (which is what Mongo uses)

ed0906:
We probably want to think about how we manage changes to Indexes e.g. TTL length changes.

At the moment the only way we can do it is by manually deleting the index (via jenkins or in code).
Given the indexes are created at start up some other options are:
- On Start Up Always delete all indexes and recreate them
- Add Delete lines in when we know we need to redeploy them
- Write some diff logic to look at the current indexes and recreate only those that have changed

app/services/cache/ExportsCacheModelRepository.scala
```diff
@@ -42,6 +44,14 @@ class ExportsCacheModelRepository @Inject()(mc: ReactiveMongoComponent)(implicit
 
   implicit val journeyFormats = ExportsCacheModel.format
 
+  override def indexes: Seq[Index] = super.indexes ++ Seq(
```
PauloPortugal: Sweet!

PauloPortugal:
@ed0906 In the past I have used MongoBee (Java) to migrate/keep track of MongoDB changes.
https://github.com/synhaptein/migmongo is probably what we need or should use for Scala projects

PR #332, opened 2019-07-19
CEDS-1249 Tests improvements

PatrykRudnicki:


build.sbt
```diff
@@ -64,7 +64,8 @@ lazy val scoverageSettings: Seq[Setting[_]] = Seq(
     "metrics\\..*",
     "features\\..*",
     "test\\..*",
-    ".*(BuildInfo|Routes|Options|TestingUtilitiesController).*"
+    ".*(BuildInfo|Routes|Options|TestingUtilitiesController).*",
+    "logger.*\\(.*\\)"
```
PatrykRudnicki: I tried to exclude logs from coverage, this line exclude just `logger.error(` If someone know how to make it to exclude message too please ping me

app/controllers/ChoiceController.scala
```diff
@@ -88,8 +88,6 @@ class ChoiceController @Inject()(
                   Redirect(controllers.routes.CancelDeclarationController.displayForm())
                 case Submissions =>
                   Redirect(controllers.routes.SubmissionsController.displayListOfSubmissions())
-                case _ =>
-                  Redirect(controllers.routes.ChoiceController.displayPage())
```
PatrykRudnicki: There is no possibility to achieve this case. Form validation will block this. Should we stay with this or removing is okay?

build.sbt
mattclark-zerogravit: nice! 👍

app/controllers/ChoiceController.scala
mattclark-zerogravit: Will we get a warning if it is removed? if so can we suppress it?

PatrykRudnicki: No, there is no warning

PR #333, opened 2019-07-19
CEDS-1410  Update an Item

ed0906:


app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -63,33 +63,54 @@ class FiscalInformationController @Inject()(
         .fold(
           (formWithErrors: Form[FiscalInformation]) =>
             Future.successful(BadRequest(fiscalInformationPage(itemId, formWithErrors))),
-          validFiscalInformation => updateCacheModelsAndRedirect(itemId, validFiscalInformation)
+          formData =>
+            formData.onwardSupplyRelief match {
+              case FiscalInformation.AllowedFiscalInformationAnswers.yes =>
```
PauloPortugal: Why do we need `updateCacheForYES` and `updateCacheForNo`?
Could you also reformat `YES` to `Yes`, so that it is consistent?

app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -63,33 +63,54 @@ class FiscalInformationController @Inject()(
...
+  ): Future[Unit] =
+    for {
+      _ <- legacyCacheService.cache[FiscalInformation](goodsItemCacheId, formId, updatedFiscalInformation)
+      _ <- getAndUpdateExportCacheModel(
+        sessionId,
+        model => {
+          val item: Option[ExportItem] =
+            model.items
+              .find(item => item.id.equals(itemId))
+              .map(_.copy(fiscalInformation = Some(updatedFiscalInformation), additionalFiscalReferencesData = None))
```
PauloPortugal: Is this the only change?
So if the user goes back we assign `None` to `additionalFiscalReferencesData`?

app/controllers/declaration/FiscalInformationController.scala
ed0906: The model gets updated differently

ed0906: Done

app/controllers/declaration/FiscalInformationController.scala
ed0906: If the user choses no, we remove the answers from the "yes" path

PR #335, opened 2019-07-19
CEDS-1414 Sort Items Table by Sequence Number

ed0906:


test/views/declaration/spec/ViewMatchers.scala
```diff
@@ -0,0 +1,226 @@
...
+ * limitations under the License.
+ */
+
+import org.jsoup.nodes.Element
+import org.jsoup.select.Elements
+import org.scalatest.matchers._
+import play.api.mvc.Call
+
+//noinspection ScalaStyle
+trait ViewMatchers {
```
ed0906: I've pulled this across from Movements. It enables us to do JSOUP matchers e.g.

`document must containElementWithId("id") and containText("text") and haveHref("url")` etc

test/views/declaration/spec/ViewSpec.scala
```diff
@@ -42,6 +49,9 @@ trait ViewSpec extends PlaySpec with GuiceOneAppPerSuite with ViewValidator {
   implicit lazy val fakeRequest: Request[AnyContentAsEmpty.type] = FakeRequest("", "").withCSRFToken
   implicit lazy val messages: Messages = messagesApi.preferred(FakeRequest())
   implicit lazy val flash: Flash = new Flash()
+  implicit protected def htmlBodyOf(html: Html): Document = Jsoup.parse(html.toString())
+  implicit protected def htmlBodyOf(page: String): Document = Jsoup.parse(page)
+  implicit protected def htmlBodyOf(result: Future[Result]): Document = htmlBodyOf(contentAsString(result))
```
ed0906: Allows us to create a view as a `HTML/String/Future[Result]` type and use JSOUP `Document` matchers without extra boilerplate

PR #336, opened 2019-07-22
CEDS-1411: Started mapping to wco-dec from new exportsCacheModel

mattclark-zerogravit:
This is for the 1st part of the new caching. Old cache mapping is intact.

### DIFF
--- a/app/forms/declaration/RepresentativeDetails.scala
+++ b/app/forms/declaration/RepresentativeDetails.scala
@@ -70,4 +70,8 @@ object RepresentativeDetails {
     val DirectRepresentative = "2"
     val IndirectRepresentative = "3"
   }
+
+  def isDefined(representativeDetails: RepresentativeDetails): Boolean =
+    representativeDetails.details.isDefined && (representativeDetails.details.get.eori.isDefined || representativeDetails.details.get.address.isDefined)
+
 }
--- a/app/services/cache/mapping/declaration/DeclarationBuilder.scala
+++ b/app/services/cache/mapping/declaration/DeclarationBuilder.scala
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services.cache.mapping.declaration
+
+import forms.declaration.RepresentativeDetails
+import services.cache.ExportsCacheModel
+import services.mapping.declaration._
+import wco.datamodel.wco.dec_dms._2.Declaration
+import wco.datamodel.wco.declaration_ds.dms._2.{DeclarationFunctionalReferenceIDType, DeclarationTypeCodeType}
+
+object DeclarationBuilder {
+
+  val defaultFunctionCode = "9"
+
+  def build(exportsCacheModel: ExportsCacheModel): Declaration = {
+    val declaration = new Declaration()
+
+    declaration.setFunctionCode(FunctionCodeBuilder.build(defaultFunctionCode))
+    declaration.setFunctionalReferenceID(createDeclarationFunctionalReferenceIDType(exportsCacheModel))
+
+    val typeCode: Option[DeclarationTypeCodeType] = exportsCacheModel.additionalDeclarationType.map(
+      additionalDeclarationType =>
+        TypeCodeBuilder.createTypeCode(additionalDeclarationType, exportsCacheModel.dispatchLocation)
+    )
+
+    declaration.setTypeCode(typeCode.getOrElse(new DeclarationTypeCodeType()))
+    declaration.setGoodsItemQuantity(GoodsItemQuantityBuilder.createGoodsItemQuantity(exportsCacheModel.items.toSeq))
+
+    generateAgent(exportsCacheModel).foreach { data =>
+      declaration.setAgent(data)
+    }
+
+    //    declaration.setGoodsShipment(GoodsShipmentBuilder.build)
+    //    declaration.setExitOffice(ExitOfficeBuilder.build)
+    //    declaration.setBorderTransportMeans(BorderTransportMeansBuilder.build)
+    //    declaration.setExporter(ExporterBuilder.build)
+    //    declaration.setDeclarant(DeclarantBuilder.build)
+    //    declaration.setInvoiceAmount(InvoiceAmountBuilder.build)
+    //    declaration.setPresentationOffice(PresentationOfficeBuilder.build)
+    //    declaration.setSpecificCircumstancesCodeCode(SpecificCircumstancesCodeBuilder.build)
+    //    declaration.setSupervisingOffice(SupervisingOfficeBuilder.build)
+    //    declaration.setTotalPackageQuantity(TotalPackageQuantityBuilder.build)
+    //    declaration.setConsignment(DeclarationConsignmentBuilder.build)
+    //    declaration.setTypeCode(TypeCodeBuilder.build)
+    //
+    //    val authorisationHolders = AuthorisationHoldersBuilder.build
+    //    if (authorisationHolders != null && !authorisationHolders.isEmpty) {
+    //      declaration.getAuthorisationHolder.addAll(authorisationHolders)
+    //    }
+    //
+    //    val currencyExchangeList = CurrencyExchangeBuilder.build
+    //    if (currencyExchangeList != null && !currencyExchangeList.isEmpty) {
+    //      declaration.getCurrencyExchange.addAll(currencyExchangeList)
+    //    }
+
+    declaration
+  }
+
+  private def createDeclarationFunctionalReferenceIDType(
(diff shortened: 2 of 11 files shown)


app/services/cache/mapping/MetaDataBuilder.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  def build(cacheModel: ExportsCacheModel): MetaData = {
+    createMetaDataWithConstants
+
+//    val element: JAXBElement[Declaration] = new JAXBElement[Declaration](
+//      new QName("urn:wco:datamodel:WCO:DEC-DMS:2", "Declaration"),
+//      classOf[Declaration],
+//      DeclarationBuilder.build(cacheMap, choice)
+//    )
+//    createMetaDataWithConstants.setAny(element)
```
PatrykRudnicki: Do you need this comment for some future references? If not please remove

app/services/cache/mapping/MetaDataBuilder.scala
```diff
@@ -0,0 +1,75 @@
...
+    createMetaDataWithConstants
```
PatrykRudnicki: You're invoking this method two times. In lines 39 and 47

app/services/cache/mapping/declaration/DeclarationBuilder.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  val defaultFunctionCode = "9"
+
+  def build(exportsCacheModel: ExportsCacheModel): Declaration = {
+    val declaration = new Declaration()
+
+    declaration.setFunctionCode(FunctionCodeBuilder.build(defaultFunctionCode))
+    declaration.setFunctionalReferenceID(createDeclarationFunctionalReferenceIDType(exportsCacheModel))
+
+    val typeCode: Option[DeclarationTypeCodeType] = for {
```
PatrykRudnicki: Just style/preference comment. Can be ignored :)

I'm thinking about using map here. for yield for one thing is probably more complicated

app/services/cache/mapping/declaration/DeclarationBuilder.scala
```diff
@@ -0,0 +1,91 @@
...
+    //
+    //    val authorisationHolders = AuthorisationHoldersBuilder.build
+    //    if (authorisationHolders != null && !authorisationHolders.isEmpty) {
+    //      declaration.getAuthorisationHolder.addAll(authorisationHolders)
+    //    }
+    //
+    //    val currencyExchangeList = CurrencyExchangeBuilder.build
+    //    if (currencyExchangeList != null && !currencyExchangeList.isEmpty) {
+    //      declaration.getCurrencyExchange.addAll(currencyExchangeList)
+    //    }
```
PatrykRudnicki: Is those comments necessary? is it for the future commits or just can be removed?

app/forms/declaration/RepresentativeDetails.scala
```diff
@@ -70,4 +70,8 @@ object RepresentativeDetails {
     val DirectRepresentative = "2"
     val IndirectRepresentative = "3"
   }
+
+  def isDefined(representativeDetails: RepresentativeDetails): Boolean =
```
PatrykRudnicki: hmm, I think you can move this method to the class. First you can avoid to have argument so you'll have

`def isDefined(): Boolean = details.isDefined && (details.get.eori.isDefined || details.get.address.isDefined)`

So you can make whole method shorter.

Or something like this, but it's up to you:

`def isDefined(): Boolean = details.map(_.eori.isDefined).getOrElse(false) || details.map(_.address.isDefined).getOrElse(false)`

app/services/cache/mapping/MetaDataBuilder.scala
mattclark-zerogravit: Yes this is the 2nd level down in the mapping the declaration level. I could have uncommented this but left it for now as the declaration level is incomplete.

app/services/cache/mapping/MetaDataBuilder.scala
mattclark-zerogravit: I'll fix this.

app/services/cache/mapping/declaration/DeclarationBuilder.scala
mattclark-zerogravit: I've copied the existing old cache builder and I'm working my way down the mapping so I need these in so I dont miss anything. They will eventually be replaced with new implementations as I get to them.

app/forms/declaration/RepresentativeDetails.scala
mattclark-zerogravit: Swings and roundabouts.. not sure either way is better. Problem of a form being made up of optional elements but the fact we need ALL of them for it to be useable / send to the backend.

app/services/cache/mapping/MetaDataBuilder.scala
mattclark-zerogravit: I'll uncomment and make this function clearer. Declaration Builder has some functionality now so can be called.

app/services/cache/mapping/declaration/DeclarationBuilder.scala
mattclark-zerogravit: I'll haver a look and try and refactor this based on your comments.

app/forms/declaration/RepresentativeDetails.scala
PatrykRudnicki: I just mean that to invoke this method you must have `RepresentativeDetails`. So this method it's related with class, you must have object to invoke this, so it can be just easier without passing arguments and usage will be easier to.

Instead of using companion object `RepresentativeDetails.isDefined(someObjectToCheck)` you will have `someObjectToCheck.isDefined`

app/services/cache/mapping/declaration/DeclarationBuilder.scala
```diff
@@ -0,0 +1,91 @@
...
+
+    declaration.setFunctionCode(FunctionCodeBuilder.build(defaultFunctionCode))
+    declaration.setFunctionalReferenceID(createDeclarationFunctionalReferenceIDType(exportsCacheModel))
+
+    val typeCode: Option[DeclarationTypeCodeType] = exportsCacheModel.additionalDeclarationType.map(
+      additionalDeclarationType =>
+        TypeCodeBuilder.createTypeCode(additionalDeclarationType, exportsCacheModel.dispatchLocation)
+    )
+
+    declaration.setTypeCode(typeCode.getOrElse(new DeclarationTypeCodeType()))
```
PauloPortugal: We cannot and should not instantiate empty types on the declaration.

app/services/cache/mapping/declaration/DeclarationBuilder.scala
```diff
@@ -0,0 +1,91 @@
...
+    declaration.setGoodsItemQuantity(GoodsItemQuantityBuilder.createGoodsItemQuantity(exportsCacheModel.items.toSeq))
+
+    generateAgent(exportsCacheModel).foreach { data =>
+      declaration.setAgent(data)
+    }
+
+    //    declaration.setGoodsShipment(GoodsShipmentBuilder.build)
```
PauloPortugal: We should not commit commented code to master, even if we are going to replace them.
Will this cause the smoke tests to fail as well?

app/services/cache/mapping/declaration/DeclarationBuilder.scala
```diff
@@ -0,0 +1,91 @@
...
+    //
+    //    val currencyExchangeList = CurrencyExchangeBuilder.build
+    //    if (currencyExchangeList != null && !currencyExchangeList.isEmpty) {
+    //      declaration.getCurrencyExchange.addAll(currencyExchangeList)
+    //    }
+
+    declaration
+  }
+
+  private def createDeclarationFunctionalReferenceIDType(
```
PauloPortugal: This is still early days, but are you thinking of declaring all the functions to map a DMS declaration?
This will turn this file into a quite big file.

app/services/cache/mapping/declaration/DeclarationBuilder.scala
```diff
@@ -0,0 +1,91 @@
...
+    exportsCacheModel: ExportsCacheModel
+  ): DeclarationFunctionalReferenceIDType = {
+    val returnVal = new DeclarationFunctionalReferenceIDType()
```
PauloPortugal: You should not instantiate a type unless the user has entered that it requires us to do so, otherwise, we will fail the submission

app/services/cache/mapping/declaration/DeclarationBuilder.scala
```diff
@@ -0,0 +1,91 @@
...
+    )
+    returnVal
+  }
+
+  private def generateAgent(exportsCacheModel: ExportsCacheModel): Option[Declaration.Agent] =
+    exportsCacheModel.parties.representativeDetails.map { representativeDetails =>
+      if (RepresentativeDetails.isDefined(representativeDetails)) {
+        AgentBuilder.createAgent(representativeDetails)
+      } else {
+        new Declaration.Agent()
```
PauloPortugal: This should not happen. You need to remove so.
Also, I still think it is better to delegate the creation of DMS declaration subtypes to its own classes, otherwise the declaration builder will be a very long class and quite bloated

app/services/mapping/declaration/AgentBuilder.scala
```diff
@@ -28,14 +28,11 @@ object AgentBuilder {
   def build(implicit cacheMap: CacheMap): Declaration.Agent =
     cacheMap
       .getEntry[RepresentativeDetails](RepresentativeDetails.formId)
-      .filter(isDefined)
+      .filter(RepresentativeDetails.isDefined)
```
PauloPortugal: `.filter(RepresentativeDetails.isDefined)`
I would have done it
`filter(_.isDefined)` instead, as it is more functional, but that is ok with me.

PR #337, opened 2019-07-22
CEDS-1409 Change JourneyAction to use the new cache model

PauloPortugal:


test/controllers/declaration/OfficeOfExitControllerSpec.scala
```diff
@@ -75,20 +66,21 @@ class OfficeOfExitControllerSpec extends CustomExportsBaseSpec with OfficeOfExit
 
   "Office Of Exit Controller during standard declaration on GET" should {
 
-    "return 200 with a success" in new StandardSetUp {
-
+    "return 200 with a success" in {
```
PatrykRudnicki: With removing traits you lost spec for supplementary and standard. The tests are the same, because working of choices we have tested in different place, but you left duplicated test.

Line 45 and 69

```
"return 200 with a success" in {

      val Some(result) = route(app, getRequest(uri))

      status(result) must be(OK)
    }
```

if you compare with test with this comment it's exactly the same..

So we have all tests duplicated, but with different name (standard and supplementary), so compilation and everything passed

test/unit/base/ControllerSpec.scala
```diff
@@ -37,18 +37,16 @@ trait ControllerSpec
...
   when(mockErrorHandler.standardErrorTemplate(anyString, anyString, anyString)(any()))
     .thenReturn(Html.apply(""))
 
-  val mockExportsMetrics = new ExportsMetrics(new MetricsImpl(new DefaultApplicationLifecycle(), minimalConfiguration))
-
-  val fakeRequest = FakeRequest("", "").withCSRFToken
+  private val fakeRequest = FakeRequest("", "").withSession(("sessionId", "sessionId"))
 
-  def getRequest(): Request[AnyContentAsEmpty.type] = fakeRequest
+  def getRequest(): Request[AnyContentAsEmpty.type] = fakeRequest.withCSRFToken
```
PatrykRudnicki: 👍

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -18,43 +18,48 @@ package unit.controllers
...
 
-    val controller = new ChoiceController(
-      mockAuthAction,
-      mockCustomsCacheService,
-      mockExportsCacheService,
-      mockErrorHandler,
-      stubMessagesControllerComponents(),
-      choicePage
-    )(ec, minimalAppConfig)
+  private val controller = new ChoiceController(
```
PatrykRudnicki: I just created those tests few days ago... way with trait was to have new controller for every test case.

If we want to use this BeforeAndAfterEach, it must be moved to the ControllerSpec.

There is no reason to extends every test with this trait.

PatrykRudnicki: I want to have a decision about it as soon as possible, because I'm working around those tests

PauloPortugal: Sure :) I am going to redo your work and then we can have a chat. 
I am taking my Wrestling spandex kit, so that we can work this one out like civilized folks.

PR #342, opened 2019-07-23
CEDS-1410 Fix ProcedureCode form validation

PauloPortugal:
If we had a list of additional procedure codes, but the form was empty
we would get a pattern matching error.
We should ensure the current form has the `procedureCode` sent on the
reuqest payload.

### DIFF
--- a/app/controllers/declaration/ProcedureCodesPageController.scala
+++ b/app/controllers/declaration/ProcedureCodesPageController.scala
@@ -209,7 +209,7 @@ class ProcedureCodesPageController @Inject()(
 
       case (procedureCode, seq) =>
         procedureCode match {
-          case ProcedureCodes(None, _) if cachedData.procedureCode.isEmpty =>
+          case ProcedureCodes(None, _) =>
             handleErrorPage(
               itemId,
               Seq(("procedureCode", "supplementary.procedureCodes.procedureCode.error.empty")),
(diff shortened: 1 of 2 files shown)


app/controllers/declaration/ProcedureCodesPageController.scala
```diff
@@ -209,7 +209,7 @@ class ProcedureCodesPageController @Inject()(
 
       case (procedureCode, seq) =>
         procedureCode match {
-          case ProcedureCodes(None, _) if cachedData.procedureCode.isEmpty =>
+          case ProcedureCodes(None, _) if procedureCode.procedureCode.isEmpty =>
```
PatrykRudnicki: This is wrong. You're matching procedure code so `ProcedureCodes(None, _)` is exactly the same like `if procedureCode.procedureCode.isEmpty`

Previous approach checked form and cached data

PatrykRudnicki: If this logic is correct, you don't need `if`. case is enough, but I think it's worth to double check this.

Whole this controller should be refactored, it's huge with really old logic. One of the first pages with multiply item, so approach is not the best

PatrykRudnicki: Moreover with this approach now, in the first case you will catch everything that don't have `procedureCode`, doesn't matter that you have adiitional procedure code and the same with cached data etc

PauloPortugal: Thanks for spotting this :)

PR #343, opened 2019-07-24
CEDS-1114 Make Container Radio Button Mandatory

ed0906:


app/forms/declaration/TransportDetails.scala
```diff
@@ -38,15 +38,17 @@ object TransportDetails {
...
-  val formMapping = mapping(
+  val formMapping: Mapping[TransportDetails] = mapping(
     "meansOfTransportCrossingTheBorderNationality" -> optional(
       text()
         .verifying(
           "supplementary.transportInfo.meansOfTransport.crossingTheBorder.nationality.error.incorrect",
           isContainedIn(allCountries.map(_.countryName))
         )
     ),
-    "container" -> boolean,
```
ed0906: The `boolean` mapping defaults to false if it is absent. The workaround below ensures it is populated by the user

PR #346, opened 2019-07-25
CEDS-1411 Consignment Mapping

ed0906:


app/services/mapping/goodsshipment/consignment/ConsignmentCarrierBuilder.scala
```diff
@@ -79,34 +90,26 @@ object ConsignmentCarrierBuilder {
 
     val carrierAddress = new Carrier.Address()
 
-    if (address.addressLine.nonEmpty) {
```
ed0906: No need for these as they are filtered above based on this

PR #348, opened 2019-07-26
CEDS-1411 More Mapping Builders

ed0906:


test/services/cache/ExportsCacheModelBuilder.scala
```diff
@@ -111,7 +128,41 @@ trait ExportsCacheModelBuilder {
 
   def withoutCarrierDetails(): CacheModifier = m => m.copy(parties = m.parties.copy(carrierDetails = None))
 
-  def withCarrierDetails(eori: Option[String] = None,
-                         address: Option[Address] =  None): CacheModifier =
+  def withCarrierDetails(eori: Option[String] = None, address: Option[Address] = None): CacheModifier =
     m => m.copy(parties = m.parties.copy(carrierDetails = Some(CarrierDetails(EntityDetails(eori, address)))))
+
+  def withoutWarehouseIdentification(): CacheModifier =
+    m => m.copy(locations = m.locations.copy(warehouseIdentification = None))
```
PatrykRudnicki: I see that in old code it's the same, but I think we should change this `m` to something better. It's unreadable. I can expect that this `m` is from `modifier` so from `CacheModifier` but it will be more readable if you will use e.g. `cache`

ed0906: Agreed, I've renamed them all cache

PR #354, opened 2019-07-30
Remove early initialization from controller

bambuchaAdm:
In scala road map [early initializers are going to be deprecated][1].
We are able to make our code more robust by not using deprecated language feature.

This PR makes controllers code more consistent and increase readability.

[1]: https://github.com/scala/scala-dev/issues/513

### DIFF
--- a/app/controllers/declaration/CarrierDetailsController.scala
+++ b/app/controllers/declaration/CarrierDetailsController.scala
@@ -39,14 +39,14 @@ class CarrierDetailsController @Inject()(
   authenticate: AuthAction,
   journeyType: JourneyAction,
   customsCacheService: CustomsCacheService,
-  override val cacheService: ExportsCacheService,
+  override val exportsCacheService: ExportsCacheService,
   mcc: MessagesControllerComponents,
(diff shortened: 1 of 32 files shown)


app/controllers/declaration/CarrierDetailsController.scala
```diff
@@ -39,14 +39,14 @@ class CarrierDetailsController @Inject()(
   authenticate: AuthAction,
   journeyType: JourneyAction,
   customsCacheService: CustomsCacheService,
-  override val cacheService: ExportsCacheService,
+  val exportsCacheService: ExportsCacheService,
```
ed0906: I think `override val` is technically correct given it is an override

bambuchaAdm:
@PauloPortugal because in most of controller we have two types of cache just `cacheService` would create ambiguity which one is `cacheService`.

@ed0906 It is just change semantic from abstract `def` to concrete `val` without changing of semantic. I am OK with this change but i personally don't like to put multiple modifiers in constructor arguments, `val` is enough intrusive.

PatrykRudnicki:
Rebased and merging...

PR #355, opened 2019-07-30
CEDS-1441 Convert GovernmentAgencyGoodsItemBuilder

ed0906:
+ remove CacheTestData as it is supersceeded by ExportsCacheItemBuilder


test/integration/services/mapping/goodsshipment/GoodsShipmentBuilderIntegrationSpec.scala
```diff
@@ -25,16 +25,17 @@ import forms.declaration.GoodsLocationTestData.correctGoodsLocation
...
+import services.cache.ExportsCacheModelBuilder
+import services.mapping.goodsshipment.GoodsShipmentBuilder
 import wco.datamodel.wco.dec_dms._2.Declaration
 
 class GoodsShipmentBuilderIntegrationSpec
-    extends WordSpec with Matchers with CacheTestData with ExportsCacheModelBuilder {
+    extends WordSpec with Matchers with ExportsCacheModelBuilder with GuiceOneAppPerSuite {
 
-  private def builder = new GoodsShipmentBuilder(new GoodsShipmentNatureOfTransactionBuilder(), new ConsigneeBuilder())
+  private def builder = app.injector.instanceOf[GoodsShipmentBuilder]
```
ed0906: Just to highlight instead of creating the builder and all its dependencies, I'm letting Guice do it for us.

Given there's quite a few more builders that will be added to this constructor it might become a big messy if we have to create them all manually

mattclark-zerogravit: 👍 I though this was slower as it had to set up an application but having just tested locally it seems it doesn't add any time to the tests. And it's better.

ed0906: It is a tiny bit slower, but providing it's just a single test it doesn't have much impact.

I think generally all IntegrationTests should be using Guice, otherwise how to we know everything integrates properly :D

PR #358, opened 2019-07-30
CEDS-1369 Make Warehouse ID mandatory if Type is selected

ed0906:


app/forms/declaration/WarehouseIdentification.scala
```diff
@@ -56,6 +56,10 @@ object WarehouseIdentification {
         )
     )
   )(WarehouseIdentification.apply)(WarehouseIdentification.unapply)
+    .verifying("supplementary.warehouse.identificationNumber.error", _ match {
+      case WarehouseIdentification(_, Some(_), None, _) | WarehouseIdentification(_, Some(_), Some(""), _) => false
```
MaciejRewera: :+1: really like the idea, but in my opinion it is not readable. As I did not know the context it took me a while to figure out what is going on.
I'd suggest maybe to encapsulate this into function with a descriptive name.

ed0906: I've pushed a change to pull it out into a method

app/forms/declaration/WarehouseIdentification.scala
```diff
@@ -56,6 +56,13 @@ object WarehouseIdentification {
         )
     )
   )(WarehouseIdentification.apply)(WarehouseIdentification.unapply)
+    .verifying("supplementary.warehouse.identificationNumber.error", idNumberIsPopulatedWhenIDTypeIsSelected)
+
+  private def idNumberIsPopulatedWhenIDTypeIsSelected: WarehouseIdentification => Boolean =
```
PauloPortugal: :+1:

PR #361, opened 2019-07-31
CEDS Fix Local Auth redirect

ed0906:


conf/application.conf
```diff
@@ -171,7 +171,7 @@ mongo-async-driver {
 
 urls {
   login = "http://localhost:9949/auth-login-stub/gg-sign-in"
-  loginContinue = "http://localhost:9000/customs-declare-exports-frontend"
+  loginContinue = "http://localhost:6791/customs-declare-exports-frontend/start"
```
PatrykRudnicki: hmm, I think it should be: `http://localhost:6791/customs-declare-exports/start` without frontend. We're not using frontend in the name from few months

test/handlers/ErrorHandlerSpec.scala
```diff
@@ -39,11 +41,13 @@ class ErrorHandlerSpec extends CustomExportsBaseSpec {
...
   "resolve error" should {
 
+    def urlEncode(value: String): String = URLEncoder.encode(value, "UTF-8")
+
     "handle no active session authorisation exception" in {
       val res = errorHandler.resolveError(req, new NoActiveSession("A user is not logged in") {})
       res.header.status must be(Status.SEE_OTHER)
       res.header.headers.get(HeaderNames.LOCATION) must be(
-        Some("/gg/sign-in?continue=%2Ffoo&origin=customs-declare-exports-frontend")
+        Some(s"http://localhost:9949/auth-login-stub/gg-sign-in?continue=${urlEncode("http://localhost:6791/customs-declare-exports-frontend/start")}")
```
PatrykRudnicki: The same here about `frontend` word

conf/application.conf
ed0906: Good point. I think it only worked for me locally because auto-submit overrode it...

PR #362, opened 2019-07-31
CEDS-1456 Index Manager Spike

ed0906:
On Start up, instead of using the standard ReactiveMongo `ensureIndexes` method, this changes it to use a custom method that does a diff on the Indexes in the DB and configured in code, and updates them accordingly.

### DIFF
--- a/app/services/cache/IndexManager.scala
+++ b/app/services/cache/IndexManager.scala
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services.cache
+
+import play.api.Logger
+import reactivemongo.api.CollectionMetaCommands
+import reactivemongo.api.indexes.Index
+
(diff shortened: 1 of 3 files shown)


PauloPortugal:
I would try https://github.com/synhaptein/migmongo first to see if this helps us.
This is a MongoDB data migration tool, backed by a community, with enough documentation and that people might know about or at least relate to, if you mention `Flyway`, `LiquidBase` or `MongoBee`. Open source frameworks over bespoke solutions is my preference, otherwise we have another `servicemanager` tool that is quite opinionated and now we need to maintain it.

The advantage you have with a tool like `migmongo` is that allows you to do and undo changes in a a critical environment (QA, Staging, Live).

Other Gov Uk project already use similar libraries: https://github.com/alphagov/pay-connector/blob/0fe3cf327f7a99cbee673a491d06b0059480c54d/src/main/resources/migrations.xml
Gov.uk Pay for their payments DB where the change set is in xml (to programming language/db agnostic porpuses)

app/services/cache/IndexManager.scala
```diff
@@ -0,0 +1,98 @@
...
+            } yield (): Unit
+          } else {
+            Future.successful((): Unit)
+          }
+
+        case _ => Future.successful((): Unit)
+      }).map(_ => (): Unit)
+  }
+
+  def ensureIndexes(): Future[Unit] = for {
```
PauloPortugal: This is how one change set would look with `migmongo`
```
changeSet("Ensure Index", "edwilson") { db =>
        List(
          db[BSONCollection]("exportsCacheModel").indexesManager.ensure(Index(Seq("field1" -> Ascending)))
        )
}
```

ed0906: Seems quite nice. As long as the set up and ongoing maintenance isn't huge. Although My preference will always be for a "make the DB look like this" approach rather than maintaining a large change log

ed0906:
I'm not overly passionate if this gets merged or not, it was a _fun_ experiment whatever approach we go with

PatrykRudnicki:
test this please

PatrykRudnicki:
So command above will trigger tests if we need to rerun :)

PR #365, opened 2019-07-31
CEDS-1249 Cleared app config in application

PatrykRudnicki:


app/views/declaration/additional_information.scala.html
```diff
@@ -24,7 +24,7 @@
 
 @this(main_template: views.html.main_template)
```
ed0906: Minor but could we inject it here rather than as an implicit? Then potentially take all `AppConfig` code out of the controllers? (maybe)

PatrykRudnicki: Yes, you're right. Removed!

app/views/main_template.scala.html
```diff
@@ -14,13 +14,11 @@
  * limitations under the License.
  *@
 
-@import config.AppConfig
 @import uk.gov.hmrc.play.views.html.layouts.{Article, Sidebar}
 
 @this(govuk_wrapper: views.html.govuk_wrapper, sidebar_di: Sidebar, article_di: Article)
 
 @(title: String,
-  appConfig: AppConfig,
```
ed0906: 🥇

app/views/govuk_wrapper.scala.html
```diff
@@ -26,11 +27,11 @@
         main_content_header_di: MainContentHeader,
         main_content_di: MainContent,
         footer_links_di: FooterLinks,
-        hmrc_gov_uk_template: GovUkTemplate
+        hmrc_gov_uk_template: GovUkTemplate,
+        appConfig: AppConfig
 )
 
-@(appConfig: config.AppConfig,
```
ed0906: 🥇

app/handlers/ErrorHandler.scala
```diff
@@ -32,16 +32,17 @@ import views.html.error_template
 import scala.concurrent.Future
 
 @Singleton
-class ErrorHandler @Inject()(appConfig: AppConfig, override val messagesApi: MessagesApi, errorPage: error_template)
-    extends FrontendErrorHandler with AuthRedirects {
+class ErrorHandler @Inject()(override val messagesApi: MessagesApi, errorPage: error_template)(
+  implicit appConfig: AppConfig
```
bambuchaAdm: Inject here `config` and `env`

app/views/choice_page.scala.html
```diff
@@ -20,13 +20,12 @@
 @import uk.gov.hmrc.play.views.html._
 @import views.components.inputs.RadioOption
 
-@this(main_template: views.html.main_template)
+@this(main_template: views.html.main_template, appConfig: AppConfig)
```
bambuchaAdm: why here ?

PatrykRudnicki: We need to have list of available journeys, I stay with appConfig in this file to didn't break app. We can think how we can do it differently

PR #368, opened 2019-08-01
CEDS-1464: switch submission to use new mapping.

mattclark-zerogravit:
And fixed related tests.


app/services/SubmissionService.scala
```diff
@@ -98,58 +89,8 @@ class SubmissionService @Inject()(
       SubmissionResult.toString -> result
     )
 
-  def getCachedData(cacheMap: CacheMap)(implicit request: JourneyRequest[_]): JsObject = {
```
ed0906: 🥇

PR #373, opened 2019-08-02
CEDS-1466 CEDS-1249 Removed scala check from test and rewrite to be unit

PatrykRudnicki:


test/generators/Generators.scala
```diff
@@ -27,61 +27,12 @@ trait Generators {
 
   implicit val dontShrink: Shrink[String] = Shrink.shrinkAny
 
-  def genIntersperseString(gen: Gen[String], value: String, frequencyV: Int = 1, frequencyN: Int = 10): Gen[String] = {
```
ed0906: 🥇

test/unit/controllers/declaration/PackageInformationControllerSpec.scala
```diff
@@ -0,0 +1,166 @@
...
+
+        val result = controller.displayPage(itemId)(getRequest())
+
+        status(result) must be(OK)
+      }
+
+      "display page method is invoked and cache contain some data" in new SetUp {
+
+        val itemWithPackageInformation =
+          ExportItem(itemId, packageInformation = List(PackageInformation(Some("12"), Some(10), Some("123"))))
```
ed0906: There's an `ExportsCacheItemBuilder` similar to the `ExportsCacheModelBuilder` if you think it helps aid readability

ed0906: You can do `aCacheItem(withPackageInformation(Some("12"), Some(10), Some("123")))`

PR #378, opened 2019-08-05
CEDS-1468: fix for document prodcued page and clean out old itemscach…

mattclark-zerogravit:
…ing service.


app/services/mapping/declaration/GoodsItemQuantityBuilder.scala
```diff
@@ -41,7 +40,7 @@ object GoodsItemQuantityBuilder {
 
   def build(implicit cacheMap: CacheMap): DeclarationGoodsItemQuantityType =
     cacheMap
-      .getEntry[Seq[GovernmentAgencyGoodsItem]](ExportsItemsCacheIds.itemsId)
+      .getEntry[Seq[GovernmentAgencyGoodsItem]]("exportItems")
```
PatrykRudnicki: Why this change? I think it, it's better to have value in val, to change this if necessary only in one place

app/services/mapping/governmentagencygoodsitem/AdditionalDocumentsBuilder.scala
```diff
@@ -19,14 +19,23 @@ package services.mapping.governmentagencygoodsitem
...
+  WriteOff => WCOWriteOff
+}
+import wco.datamodel.wco.declaration_ds.dms._2.AdditionalDocumentEffectiveDateTimeType.{
+  DateTimeString => WCODateTimeString
+}
+import wco.datamodel.wco.declaration_ds.dms._2.{
+  AdditionalDocumentEffectiveDateTimeType,
+  SubmitterNameTextType,
+  WriteOffQuantityQuantityType,
+  _
```
PatrykRudnicki: If we're importing at the end everything, do we need to have three types specified before?

mattclark-zerogravit: This will be intelliJ doing its thing as I cleaned up the imports as I had unused ones.

app/services/mapping/declaration/GoodsItemQuantityBuilder.scala
mattclark-zerogravit: ExportsItemsCacheIds was a class inside itemscachingService. In reality this reference is going to be deleted as the cachemap / exportItems reference will cease to exist once the old caching is full removed. Decided to do this as a quick fix so the code compiles.

PR #379, opened 2019-08-05
Ceds 1249 warehouse identification

bambuchaAdm:
Include setup to reset mocks after each test case.

### DIFF
--- a/app/controllers/actions/JourneyAction.scala
+++ b/app/controllers/actions/JourneyAction.scala
@@ -30,10 +30,9 @@ import uk.gov.hmrc.play.HeaderCarrierConverter
 
 import scala.concurrent.{ExecutionContext, Future}
 
-case class JourneyAction @Inject()(cacheService: ExportsCacheService, mcc: MessagesControllerComponents)
+case class JourneyAction @Inject()(cacheService: ExportsCacheService)
+                                  (implicit override val executionContext: ExecutionContext)
     extends ActionRefiner[AuthenticatedRequest, JourneyRequest] with SessionIdAware {
-
-  implicit override val executionContext: ExecutionContext = mcc.executionContext
 
   private val logger = Logger(this.getClass())
 
--- a/app/controllers/actions/JourneyAction.scala
+++ b/app/controllers/actions/JourneyAction.scala
@@ -30,9 +30,9 @@ import uk.gov.hmrc.play.HeaderCarrierConverter
 
 import scala.concurrent.{ExecutionContext, Future}
 
-case class JourneyAction @Inject()(cacheService: ExportsCacheService)
-                                  (implicit override val executionContext: ExecutionContext)
-    extends ActionRefiner[AuthenticatedRequest, JourneyRequest] with SessionIdAware {
+class JourneyAction @Inject()(cacheService: ExportsCacheService)(
+  implicit override val executionContext: ExecutionContext
+) extends ActionRefiner[AuthenticatedRequest, JourneyRequest] with SessionIdAware {
 
   private val logger = Logger(this.getClass())
 
--- a/app/controllers/declaration/SealController.scala
+++ b/app/controllers/declaration/SealController.scala
@@ -39,7 +39,6 @@ class SealController @Inject()(
   authenticate: AuthAction,
   journeyType: JourneyAction,
   errorHandler: ErrorHandler,
-  customsCacheService: CustomsCacheService,
   override val exportsCacheService: ExportsCacheService,
   mcc: MessagesControllerComponents,
   sealPage: seal
@@ -108,15 +107,9 @@ class SealController @Inject()(
     }
 
   private def updateCache(sessionId: String, formData: Seq[Seal])(implicit req: JourneyRequest[_]): Future[Unit] =
-    for {
-      _ <- getAndUpdateExportCacheModel(
-        sessionId,
-        model =>
-          exportsCacheService
-            .update(sessionId, model.copy(seals = formData))
-      )
-      _ <- customsCacheService.cache[Seq[Seal]](cacheId, formId, formData)
-    } yield Unit
+    getAndUpdateExportCacheModel(sessionId, model => {
+      exportsCacheService.update(sessionId, model.copy(seals = formData))
+    }).map(_ => ())
 
   private def addSeal(boundForm: Form[Seal], elementLimit: Int, seals: Seq[Seal])(
     implicit request: JourneyRequest[_]
--- a/test/base/MockCustomsCacheService.scala
+++ b/test/base/MockCustomsCacheService.scala
@@ -17,10 +17,11 @@
 package base
 
 import forms.Choice
-import org.mockito.ArgumentMatchers
+import org.mockito.{ArgumentMatchers, Mockito}
 import org.mockito.ArgumentMatchers.any
 import org.mockito.Mockito.when
 import org.mockito.stubbing.OngoingStubbing
+import org.scalatest.{BeforeAndAfterEach, Suite}
 import org.scalatest.mockito.MockitoSugar
 import play.api.data.Form
 import services.CustomsCacheService
@@ -28,7 +29,7 @@ import uk.gov.hmrc.http.cache.client.CacheMap
 
 import scala.concurrent.Future
 
-trait MockCustomsCacheService extends MockitoSugar {
+trait MockCustomsCacheService extends MockitoSugar with BeforeAndAfterEach { self: Suite =>
 
   val mockCustomsCacheService = mock[CustomsCacheService]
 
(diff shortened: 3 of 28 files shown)


PatrykRudnicki:
test this please

bambuchaAdm:
test this please

test/base/MockExportsCacheService.scala
```diff
@@ -16,15 +16,17 @@
...
 import org.mockito.Mockito.{never, verify, when}
+import org.scalatest.{BeforeAndAfterEach, Suite}
 import org.scalatest.mockito.MockitoSugar
 import services.cache.{ExportsCacheModel, ExportsCacheModelBuilder, ExportsCacheService}
 
 import scala.concurrent.Future
 
-trait MockExportsCacheService extends MockitoSugar with ExportsCacheModelBuilder {
+trait MockExportsCacheService extends MockitoSugar with ExportsCacheModelBuilder with BeforeAndAfterEach {
+  self: Suite =>
```
PatrykRudnicki: Why we need this Suite?

test/services/cache/ExportItemBuilder.scala
```diff
@@ -0,0 +1,47 @@
...
+  def anItem(modifiers: ExportItemModifier*): ExportItem =
+    modifiers.foldLeft(ExportItem(id = generateId())) {
+      case (entry, modifier) => modifier(entry)
+    }
+
+  def withPackageInformations(newInformations: List[PackageInformation]): ExportItemModifier = { item =>
+    item.copy(packageInformation = newInformations)
+  }
+
+  def withPackage(
```
PatrykRudnicki: We have similar: `ExportsCacheItemBuilder` please check this and use it if it's necessary. If not, tell me why.
As I See withPackage method is similar. The same with whole trait

test/unit/controllers/declaration/PackageInformationControllerSpec.scala
```diff
@@ -110,7 +124,7 @@ class PackageInformationControllerSpec extends ControllerSpec {
       "user reached limit of items" in new SetUp {
 
         val packageInformation = PackageInformation(None, None, None)
-        val maxItems = aCachedItem(withPackageInformation(packageInformation, Seq.fill(98)(packageInformation): _*))
+        val maxItems = anItem(withPackageInformations(packageInformation :: List.fill(98)(packageInformation)))
```
PatrykRudnicki: We have created `aCachedItem` method inside ExportsCacheItemBuilder.

test/unit/mock/ErrorHandlerMocks.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.mockito.Mockito
+import org.mockito.Mockito.when
+import org.scalatest.{BeforeAndAfterEach, Suite}
+import org.scalatest.mockito.MockitoSugar
+import play.api.mvc.Results.BadRequest
+import play.twirl.api.Html
+
+import scala.concurrent.Future
+
+trait ErrorHandlerMocks extends BeforeAndAfterEach { self: MockitoSugar with Suite =>
```
PatrykRudnicki: The same question like before. Why we need Suite?

test/base/MockExportsCacheService.scala
bambuchaAdm: BeforeAndAfterEach is posible to mix in only to `Suite`

app/controllers/actions/JourneyAction.scala
```diff
@@ -30,10 +30,9 @@ import uk.gov.hmrc.play.HeaderCarrierConverter
 
 import scala.concurrent.{ExecutionContext, Future}
 
-case class JourneyAction @Inject()(cacheService: ExportsCacheService, mcc: MessagesControllerComponents)
-    extends ActionRefiner[AuthenticatedRequest, JourneyRequest] with SessionIdAware {
-
-  implicit override val executionContext: ExecutionContext = mcc.executionContext
+case class JourneyAction @Inject()(cacheService: ExportsCacheService)(
```
ed0906: Unrelated to this change but this probably doesn't need to be a `case class` as we dont need any of the benefits that brings

test/base/MockExportsCacheService.scala
PatrykRudnicki: Cheers!

test/base/MockCustomsCacheService.scala
```diff
@@ -17,18 +17,19 @@
...
+import org.scalatest.{BeforeAndAfterEach, Suite}
 import org.scalatest.mockito.MockitoSugar
 import play.api.data.Form
 import services.CustomsCacheService
 import uk.gov.hmrc.http.cache.client.CacheMap
 
 import scala.concurrent.Future
 
-trait MockCustomsCacheService extends MockitoSugar {
+trait MockCustomsCacheService extends MockitoSugar with BeforeAndAfterEach { self: Suite =>
```
ed0906: What's the benefit of the `self: Suite => `?

test/base/MockExportsCacheService.scala
ed0906: It works in lots of test classes that don't have the `self: Suite =>`?

test/services/cache/ExportItemBuilder.scala
ed0906: Yep why have we duplicated this?

test/base/MockCustomsCacheService.scala
bambuchaAdm: It is required by `BeforeAndAfterEach`

test/services/cache/ExportItemBuilder.scala
bambuchaAdm: Fixed - i just did not expect that build could name like this

bambuchaAdm:
test this please

test/services/cache/ExportItemBuilder.scala
ed0906: Feel free to rename the existing one

test/unit/controllers/declaration/WarehouseIdentificationControllerSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+        "identificationType" -> exampleWarehauseIdentificationType,
+        "identificationNumber" -> exampleWarehauseIdentificationNumber,
+        "inlandModeOfTransportCode" -> exampleTransportMode
+      )
+
+      val result = controller.saveWarehouse().apply(postRequest(body))
+
+      await(result)
+
+      verify(mockExportsCacheService).update(any(), any())
```
ed0906: We should verify what we call update with.

There is some helper methods e.g. `theCacheModelUpdated` which uses argument captors allowing you tot assert on the content

bambuchaAdm: Maybe we should also start checking `id` of model we are going to update ?

ed0906: If we were being thorough then yes, but because it's generated it's a bit awkward. I guess we could mock the `journeySessionId` or whatever generates it

app/controllers/declaration/SealController.scala
```diff
@@ -108,15 +107,9 @@ class SealController @Inject()(
...
-    for {
-      _ <- getAndUpdateExportCacheModel(
-        sessionId,
-        model =>
-          exportsCacheService
-            .update(sessionId, model.copy(seals = formData))
-      )
-      _ <- customsCacheService.cache[Seq[Seal]](cacheId, formId, formData)
-    } yield Unit
+    getAndUpdateExportCacheModel(sessionId, model => {
```
PatrykRudnicki: Do we need curly brackets here?

app/controllers/declaration/SealController.scala
```diff
@@ -108,15 +107,9 @@ class SealController @Inject()(
...
+      exportsCacheService.update(sessionId, model.copy(seals = formData))
+    }).map(_ => ())
```
PatrykRudnicki: Do we need map this to Unit? is it necessary? It doesn't matter what this function return, because in code where we use it we just use `map( _ => ...)`

test/forms/declaration/WarehouseIdentificationSpec.scala
```diff
@@ -16,8 +16,61 @@
 
 package forms.declaration
 
+import base.TestHelper
 import forms.declaration.TransportCodes._
+import helpers.views.declaration.WarehouseIdentificationMessages
 import play.api.libs.json.{JsObject, JsString, JsValue}
+import unit.base.UnitSpec
+
+class WarehouseIdentificationSpec extends UnitSpec with WarehouseIdentificationMessages {
```
PatrykRudnicki: Please move this spec to `unit/forms/declaration/WarehouseIdentificationSpec`.

PR #383, opened 2019-08-06
CEDS-1466 CEDS-1249 Changed SealSpec TransportDetailsControllerSpec

PatrykRudnicki:


test/unit/forms/declaration/SealSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+  val form = Seal.form
+
+  "Seal object" should {
+
+    "has correct form id" in {
+
+      Seal.formId must be("Seal")
+    }
+
+    "has correct seals limit" in {
```
bambuchaAdm: i would put limit into test name

test/unit/forms/declaration/SealSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    "has errors" when {
+
+      "id field is empty" in {
+
+        val emptyForm = Map("id" -> "")
+
+        val result = form.bind(emptyForm)
+        val error = result.errors.head
```
bambuchaAdm: what is it contains more errors ?

test/unit/forms/declaration/SealSpec.scala
PatrykRudnicki: Noted, will change it in next PR

test/unit/forms/declaration/SealSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+
+    "has no errors" when {
+
+      "id field contains correct value" in {
+
+        val correctForm = Seal("12345")
+
+        val result = form.fill(correctForm)
+
+        result.hasErrors must be(false)
```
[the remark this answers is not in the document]
PatrykRudnicki: You're right, is `result.errors must be(Seq.empty)` better?

test/unit/forms/declaration/SealSpec.scala
PatrykRudnicki: It's only one field, it's not possibile to have more. But I can change this to compare `result.errors` with Sequence with this specific error or add case that length of errors must be 1

test/unit/forms/declaration/SealSpec.scala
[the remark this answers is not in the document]
bambuchaAdm: must be (empty)

test/unit/forms/declaration/SealSpec.scala
bambuchaAdm: We would have much more validation of errors - maybe some custom matcher would be nice

PR #384, opened 2019-08-06
CEDS-1476: Remove old caching from controllers and tests.

mattclark-zerogravit:


app/services/cache/ExportsCacheModelRepository.scala
```diff
@@ -69,6 +70,8 @@ class ExportsCacheModelRepository @Inject()(mc: ReactiveMongoComponent, appConfi
         updateResult.result[ExportsCacheModel]
       }
 
+  def remove(sessionId: String): Unit = collection.findAndRemove(selector = bySessionId(sessionId)).map(_ => ())
```
bambuchaAdm: Why we ignoring result future with this - even logging of errors ?

app/services/SubmissionService.scala
```diff
@@ -55,12 +56,11 @@ class SubmissionService @Inject()(
     auditService.auditAllPagesUserInput(getCachedData(exportsCacheModel))
     exportsConnector.submitExportDeclaration(data.ducr, data.lrn, data.payload).flatMap {
       case HttpResponse(ACCEPTED, _, _, _) =>
-        cacheService.remove(cacheId).map { _ =>
-          auditService.audit(AuditTypes.Submission, auditData(data.lrn, data.ducr, Success.toString))
-          exportsMetrics.incrementCounter(submissionMetric)
-          timerContext.stop()
-          data.lrn
-        }
+        cacheService.remove(sessionId)
```
bambuchaAdm: Here we have little change of semantic - what if in removing from cache there goes something wrong - we don't have even know abut it.

app/models/declaration/governmentagencygoodsitem/GovernmentAgencyGoodsItem.scala
```diff
@@ -90,23 +87,4 @@ case class GovernmentAgencyGoodsItemAdditionalDocumentSubmitter(
...
-case class DateTimeString(formatCode: String, value: String) {
-  import DateTimeFormats._
-
-  def time(): ZonedDateTime = formatCode match {
-    case "102" => handle102Format(value)
-    case "304" => handle304Format(value)
-  }
-}
-
-object DateTimeFormats {
```
bambuchaAdm: Why this glue-code goes out?

mattclark-zerogravit: I checked for usages and couldn't find any

PatrykRudnicki: Those formats was related with wco-dec. Hopefully now we're handling date in different (correct) way. Those two formats are formats that you can find in your declaration. I don't remember why we handled this in way above but I'm sure it was related with wco-dec

mattclark-zerogravit: We persisted wco models in the cache.. so I think these formats were used to convert the dates before they were cached.

app/services/SubmissionService.scala
mattclark-zerogravit: Have changed how we do this so all code in controller is handled inside the future.

app/services/cache/ExportsCacheModelRepository.scala
mattclark-zerogravit: Have changed how we do this so all code in controller is handled inside the future.

PR #386, opened 2019-08-07
Bau update controller specs

SteveSugden:


test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
```diff
@@ -113,9 +113,8 @@ class ConsignmentReferencesControllerSpec
     "redirect to 'Exporter Details' page" in {
 
       val result = route(app, postRequest(uri, correctConsignmentReferencesJSON)).get
-      val header = result.futureValue.header
 
-      header.headers.get("Location") must be(Some("/customs-declare-exports/declaration/exporter-details"))
+      verifyLocation(result, "/customs-declare-exports/declaration/exporter-details")
```
ed0906: What was wrong with just using `redirectLocation`?

I think it's better to have all the assertions in the test itself, custom `verify` methods just hide away code making it harder to debug

ed0906: If you're worried about duplicate code, you could always try doing a custom matcher so we can write
`result must have(location(routes.WhateverController.get()))`

The `ViewMatchers` trait has some examples of custom matchers (although I wouldnt add it that file as its specific to Views)

PR #387, opened 2019-08-07
CEDS-1446 Removed scala check from another specs

PatrykRudnicki:
PackageInformationSpec, only small tests, this page will have completely different behaviour than actual one, so just don't want to waste time on it. Minimum tests


app/controllers/declaration/CommodityMeasureController.scala
```diff
@@ -46,7 +46,8 @@ class CommodityMeasureController @Inject()(
       (packageInformation, commodityMeasure) match {
         case (Some(p), Some(data)) if p.nonEmpty => Ok(goodsMeasurePage(itemId, form().fill(data)))
         case (Some(p), _) if p.nonEmpty          => Ok(goodsMeasurePage(itemId, form()))
-        case _                                   => BadRequest(goodsMeasurePage(itemId, form().withGlobalError(ADD_ONE)))
+        case _ =>
+          BadRequest(goodsMeasurePage(itemId, form().withGlobalError("supplementary.commodityMeasure.global.addOne")))
```
bambuchaAdm: maybe new lines for every case ?

app/forms/declaration/PackageInformation.scala
```diff
@@ -22,6 +22,7 @@ import play.api.libs.json.Json
 import services.PackageTypes
 import utils.validators.forms.FieldValidator._
 
+//TODO Change those field to be non optional - all page is mandatory
```
bambuchaAdm: Ticket ID ?

PR #390, opened 2019-08-07
CEDS-1433 Align Remove Item Behaviour across all controllers

ed0906:
The desired behaviour is that clicking "Remove" on a multi-item list *only* removes that item and doesnt impact any other fields on the page.
As a result if there is any partially filled in fields on the same page, these should still be populated (but not submitted) after the "Remove" logic is complete.

The basic pattern I've implemented is
On submit:
- For `Save & Continue` and `Add` operations validate the form. Respond with a BadRequest if it is invalid
- For `Remove` operations, *don't* validate the form. Once removal is complete, "refresh"  the page with the same form as sent with the `Remove` request.

Previous logic was different in each controllers but involved validating the form first, removing the field, then redirecting back to the same page.

Also there was about 3 ways of displaying items in a list with a remove button. There's just one now.


app/views/components/buttons/save_and_continue_button.scala.html
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import controllers.util.SaveAndContinue
+
+@(messageKey: String = "site.save_and_continue")(implicit messages: Messages)
+
+<button id="submit" class="button" name="@SaveAndContinue.toString">@messages(messageKey)</button>
```
PatrykRudnicki: Hmm, I think we had save and contrinue button. What's this?

test/base/ViewValidator.scala
```diff
@@ -48,10 +52,13 @@ trait ViewValidator extends MustMatchers {
     elements.first()
   }
 
+  @deprecated("Please use 'page.getElementsBySelector must ...'")
   def getElementsByCss(html: Html, selector: String): Elements = asDocument(html).select(selector)
 
+  @deprecated("Please use 'page.getElementsBySelector must ...'")
   def getElementsByCss(page: String, selector: String): Elements = asDocument(page).select(selector)
```
PatrykRudnicki: Hmm, I think about not using anything based on the Css, just id.

but still `page.getElementsBySelector` is 100% better than `asDocument(page).select(selector)`

test/controllers/declaration/DeclarationHolderControllerSpec.scala
```diff
@@ -284,20 +284,6 @@ class DeclarationHolderControllerSpec
           verifyTheCacheIsUnchanged()
         }
       }
-
-      "try to remove not added Additional code" in {
```
PatrykRudnicki: What about this case?

test/controllers/declaration/DocumentsProducedControllerSpec.scala
```diff
@@ -187,21 +187,6 @@ class DocumentsProducedControllerSpec
 
       }
 
-      "try to remove a non existent document" in {
```
PatrykRudnicki: here the same and probably in more cases, but I'll just left comments here :)

test/views/declaration/TransportInformationContainersViewSpec.scala
```diff
@@ -107,13 +107,11 @@ class TransportInformationContainersViewSpec
         transportContainersPage(form, Seq(TransportInformationContainer("Test")))(fakeRequest, messages)
 
       // table header
-      getElementByCss(view, "form>div.field-group>table>thead>tr>th").text() must be(messages(ticTitle))
+      view.getElementById("removable_elements-heading") must containText(messages(ticTitle))
```
PatrykRudnicki: 👍

app/views/components/fields/field_multiple_values.scala.html
```diff
@@ -15,15 +15,18 @@
  *@
 
 @import controllers.util.{Add, Remove}
-
 @import views.html.components.add_button
+@import views.components.fields.LabelledValue
+
+@import sun.font.Decoration.Label
```
bambuchaAdm: ?

test/base/ViewValidator.scala
```diff
@@ -61,6 +68,7 @@ trait ViewValidator extends MustMatchers {
     element
   }
 
+  @deprecated("Please use 'page.getElementsByAttribute(name) must ...'")
```
bambuchaAdm: :+1:

app/views/components/fields/field_multiple_values.scala.html
ed0906: Removed wasnt used

app/views/components/buttons/save_and_continue_button.scala.html
ed0906: Hmmm there's `submit_button` but it is subtly different as it is wrapped in a `div class="section"` and doesn't have the `name` attribute set

test/controllers/declaration/DeclarationHolderControllerSpec.scala
ed0906: I dont see why it matters if we try and remove something that isnt there already. The outcome is still the same in that "the thing you asked us to remove doesnt exist" after the remove operation is complete.

If we can reduce the things that can lead a user to the "error!" page I think we should

test/controllers/declaration/DocumentsProducedControllerSpec.scala
ed0906: Similar to my response above

There's two scenarios:
- Remove something that exists
- Remove something that doesnt exist

The result of both is that the thing no longer exists. Does it matter if it ever existed? I don't think so

app/views/components/buttons/save_and_continue_button.scala.html
ed0906: I think the old submit button should be replaced as this one is more reusable.

It's not wrapped in a div so can be used in conjunction with the add button, or in any other context

ed0906:
> LGTM, I'll add approve when you answer on the questions :)
> 
> and I know, it's boring... but please run scalafmt :)

Done

PR #394, opened 2019-08-07
CEDS-1369 warehouse type and number validation

SteveSugden:


app/forms/declaration/WarehouseIdentification.scala
```diff
@@ -57,8 +57,16 @@ object WarehouseIdentification {
           )
       )
     )(WarehouseIdentification.apply)(WarehouseIdentification.unapply)
+    .verifying("supplementary.warehouse.identificationType.error", typeSelectedWhenNumberIsPopulated)
     .verifying("supplementary.warehouse.identificationNumber.empty", idNumberIsPopulatedWhenIDTypeIsSelected)
 
+  private def typeSelectedWhenNumberIsPopulated: WarehouseIdentification => Boolean =
+    w =>
+      w.identificationNumber match {
```
bambuchaAdm: w.identificationType.exists(_.nonEmpty) || w.identificationType.isEmpty

SteveSugden: I think you mean 
`w => w.identificationNumber.isEmpty || w.identificationType.exists(_.nonEmpty)`
I just "matched" the style of the existing `idNumberIsPopulatedWhenIDTypeIsSelected` method but I can change both if you prefer.

app/views/declaration/warehouse_identification.scala.html
```diff
@@ -54,7 +54,7 @@
                 hint = None,
                 inputs = WarehouseIdentification.IdentifierType.all.map(v =>
                     RadioOption(s"identification_type_${v.toLowerCase}", v, messages(s"supplementary.warehouse.identificationType.${v.toLowerCase}"))
-                )
+                ) :+ RadioOption("identification_type_none", "", "supplementary.warehouse.identificationType.none")
```
ed0906: 👍

app/forms/declaration/WarehouseIdentification.scala
ed0906: Whichever you think reads best

app/forms/declaration/WarehouseIdentification.scala
```diff
@@ -57,8 +57,16 @@ object WarehouseIdentification {
           )
       )
     )(WarehouseIdentification.apply)(WarehouseIdentification.unapply)
+    .verifying("supplementary.warehouse.identificationType.error", typeSelectedWhenNumberIsPopulated)
     .verifying("supplementary.warehouse.identificationNumber.empty", idNumberIsPopulatedWhenIDTypeIsSelected)
 
+  private def typeSelectedWhenNumberIsPopulated: WarehouseIdentification => Boolean =
+    w =>
```
PatrykRudnicki: I know that the scope of this `w` is small, but I think it's really better to use this `warehouse` or something different. In few places, where you have bigger part of the code or a lot of one letter variables it's time consuming to connect this with correct name

PR #396, opened 2019-08-08
CEDS-1433 - refresh list of seals to display after remove

SteveSugden:


app/controllers/declaration/SealController.scala
```diff
@@ -97,7 +97,7 @@ class SealController @Inject()(
   ): Future[Result] = {
     val updatedSeals = remove(ids.headOption, cachedSeals)
     updateCache(journeySessionId, updatedSeals).map { _ =>
-      Ok(sealPage(userInput.discardingErrors, cachedSeals, hasContainers))
+      Ok(sealPage(userInput.discardingErrors, updatedSeals, hasContainers))
```
ed0906: 👍 It's a shame we don't unit test what the view should look like

SteveSugden: Indeed.  I couldn't find any tests for add/remove functionality.  Maybe add a ticket to backlog?

PatrykRudnicki: I think sticker to tech board and discuss about it. We had view specs inside controller few months ago, after that moved out, in few places duplicated tests, now controller doesn't check it. If we want to do it, yeah, we can, why not. But it'll be better to be covered in acceptance tests or some integration tests/component tests.

PR #402, opened 2019-08-09
CEDS-1479 Submit ExportDeclarations to the Back End

ed0906:
This changes the Submission Service to POST to the JSON endpoint of the back end as well as it's current behaviour


app/controllers/ChoiceController.scala
```diff
@@ -63,10 +63,11 @@ class ChoiceController @Inject()(
             .update(
               authenticatedSessionId,
               ExportsDeclaration(
+                DeclarationStatus.DRAFT,
                 authenticatedSessionId,
                 UUID.randomUUID().toString,
-                createdDateTime = LocalDateTime.now,
-                updatedDateTime = LocalDateTime.now,
+                createdDateTime = Instant.now,
+                updatedDateTime = Instant.now,
```
ed0906: Just to highlight I've changed the way we store DateTimes. This wont affect existing data as we have a Mongo formatter for them.

project/AppDependencies.scala
```diff
@@ -31,5 +32,21 @@ object AppDependencies {
     "org.mockito" % "mockito-core" % "2.27.0" % "test"
   )
 
-  def apply(): Seq[ModuleID] = compile ++ test
+  private lazy val jettyVersion = "9.2.26.v20180806"
+  val overrides: Set[ModuleID] = Set(
```
ed0906: Wiremock and Selenium both have transitive dependencies on different versions of `jetty-http`. The latest version causes `ClassNotFoundException` so I've overridden it with a slightly older version.

ed0906: I'm introducing Wiremock as it appears we currently test the connector layer using a mocked out HTTP Client. 
To properly test all the JSON Serialization etc works I think we should use WireMock

project/AppDependencies.scala
```diff
@@ -20,9 +20,10 @@ object AppDependencies {
     "com.typesafe.play" %% "play-json-joda" % "2.6.10"
   )
 
-  val test = Seq(
+  val test: Seq[ModuleID] = Seq(
     //TODO remove hmrctest dependency
     "uk.gov.hmrc" %% "hmrctest" % "3.8.0-play-26" % "test",
+    "com.github.tomakehurst" % "wiremock" % "2.24.1" % "test",
```
bambuchaAdm: There is something like `wiremock-jre8` would allow you to not have all overrides

project/plugins.sbt
```diff
@@ -6,6 +6,8 @@ resolvers += "HMRC Releases" at "https://dl.bintray.com/hmrc/releases"
 
 addSbtPlugin("uk.gov.hmrc" % "sbt-auto-build" % "1.16.0")
 
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2")
```
ed0906: A  nice handy plugin for visualising project dependencies (and what they depend on)

I can remove this if required

project/AppDependencies.scala
ed0906: I'll give it a try

ed0906: Works, nice 🥇

test/connectors/WiremockTestServer.scala
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  override protected def beforeAll(): Unit = {
+    super.beforeAll()
+    wireMockServer.start()
+    WireMock.configureFor(wireHost, wirePort)
+  }
+
+  override protected def afterAll(): Unit = {
+    super.afterAll()
```
bambuchaAdm: to have stack order we should call `afterAll` as last one

I have in mind when two or more traits with override `BeforeAndAfterAll` are mixed into test. I have in mind that would have more expected behavior 

```
BeforeAll1
BeforeAll2
AfterAll2
AfterAll1
```

project/plugins.sbt
bambuchaAdm: Those type of utility plugin is nice to have in user sbt profile
```
[lukasz.dubiel@localhost customs-declare-exports-frontend]$ cat ~/.sbt/0.13/plugins/dependecy-graph.sbt 
addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2")
```
then all of them would be available in every project

test/connectors/WiremockTestServer.scala
ed0906: Done, I suspect there will be loads of these already in the code we'll need to fix

ed0906:
test this please

PR #404, opened 2019-08-12
Pass updated model to view

bambuchaAdm:


app/controllers/declaration/ItemTypeController.scala
```diff
@@ -103,8 +103,11 @@ class ItemTypeController @Inject()(
     val itemTypeUpdated = updateCachedItemTypeAddition(itemId, itemTypeInput, itemTypeCache)
     ItemTypeValidator.validateOnAddition(itemTypeUpdated) match {
       case Valid =>
-        updateExportsCache(itemId, itemTypeUpdated).flatMap { _ =>
-          refreshPage(itemId, itemTypeInput, hasFiscalReferences)
+        updateExportsCache(itemId, itemTypeUpdated).map {
+          case Some(model) =>
+            refreshPage(itemId, itemTypeInput, model)
+          case None =>
+            InternalServerError("Declaration not updated")
```
ed0906: I dont think we should ever throw an `InternalServerError`. This could happen the TTL removes a users session which is normal behaviour. I would argue we should just redirect to the Start Page, or handle it in a way that doesn't create a bad user experience

app/controllers/declaration/ItemTypeController.scala
```diff
@@ -196,33 +199,37 @@ class ItemTypeController @Inject()(
...
     }
-    updateExportsCache(itemId, itemTypeUpdated).flatMap { _ =>
-      val itemTypeInput: ItemType = ItemType.form().bindFromRequest().value.getOrElse(ItemType.empty)
-      refreshPage(itemId, itemTypeInput, hasFiscalReferences)
+    updateExportsCache(itemId, itemTypeUpdated).map {
+      case Some(model) =>
+        val itemTypeInput: ItemType = ItemType.form().bindFromRequest().value.getOrElse(ItemType.empty)
+        refreshPage(itemId, itemTypeInput, model)
+      case None =>
+        BadRequest("Model not updated")
```
ed0906: Similar here, lets just redirect somewhere sensible if this happens rather than risking a bad user experience

app/controllers/declaration/ItemTypeController.scala
ed0906: Could redirect to either `export-items` or `start` I guess depending on which you think is most suitable

bambuchaAdm: export-items would be better - if there are no declaration there should be second redirect

PR #405, opened 2019-08-12
Use model in additional fiscal information + reformat

bambuchaAdm:


app/models/ExportsDeclaration.scala
```diff
@@ -47,6 +47,12 @@ case class ExportsDeclaration(
   seals: Seq[Seal] = Seq.empty
 ) {
   def itemBy(itemId: String): Option[ExportItem] = items.find(_.id.equalsIgnoreCase(itemId))
+
+  def updatedItem(itemId: String, update: ExportItem => ExportItem): ExportsDeclaration =
```
ed0906: I'd name it `updateItem`. Also why does it pass an `ExportItem => ExportItem` rather than just an `ExportItem`?

bambuchaAdm: We have immutable structure - in `scala.collection.Map` there is method `updated` and here i am tend to mimic that convention.

Having just setter here would break hermetization - passing function for updating existing object does not reveal any information about internal structure of Declaration and ExporItems and could be used as setter simply by ignoring argument.

app/controllers/declaration/AdditionalFiscalReferencesController.scala
```diff
@@ -113,8 +112,10 @@ class AdditionalFiscalReferencesController @Inject()(
     cachedData: AdditionalFiscalReferencesData
   )(implicit request: JourneyRequest[_]): Future[Result] = {
     val updatedCache = MultipleItemsHelper.remove(values.headOption, cachedData.references)
-    updateExportsCache(itemId, journeySessionId, AdditionalFiscalReferencesData(updatedCache))
-      .map(_ => Ok(additionalFiscalReferencesPage(itemId, form.discardingErrors)))
+    updateExportsCache(itemId, AdditionalFiscalReferencesData(updatedCache)).map {
+      case Some(model) => Ok(additionalFiscalReferencesPage(itemId, form.discardingErrors))
+      case None        => Redirect(routes.ItemsSummaryController.displayPage())
```
SteveSugden: In what circumstance would the result of the call to updateExportsCache be Future(None)?

bambuchaAdm: Where there is delayed post after expiration for cache.

ed0906: If the item doesn't exist I think

SteveSugden: I don't see other controllers worrying about that.  Why would the item not exist?  It existed milliseconds earlier when the request was first handled.

PR #413, opened 2019-08-14
CEDS-1499 content review

SteveSugden:


app/assets/stylesheets/customsdecexfrontend-app.scss
```diff
@@ -32,3 +32,7 @@ a.button--gray, button.button--gray, .button--gray:visited {
 .mb-3 {
   margin-bottom: 3rem;
 }
+
+.margin-or {
```
ed0906: margin or?

app/views/components/address.scala.html
```diff
@@ -43,10 +43,11 @@
         label = messages("supplementary.address.postCode")
     )
 
-    @components.fields.field_autocomplete(
+    @components.fields.field_autocomplete_with_labelclass(
```
ed0906: Couldnt we just add an optional label class to the existing autocomplete?

app/assets/stylesheets/customsdecexfrontend-app.scss
SteveSugden: Its used to create space around the content "Or".

app/views/declaration/add_transport_containers.scala.html
```diff
@@ -44,8 +44,6 @@
             )
         }
 
-        <br/>
```
ed0906: 👍

app/assets/stylesheets/customsdecexfrontend-app.scss
ed0906: Can we name it similar to the others in here e.g. `mt-5`

app/views/cancel_declaration.scala.html
```diff
@@ -57,7 +57,7 @@
             labelClass = Some("bold-small")
         )
 
-        @components.fields.field_radio_with_hint(
```
ed0906: 👍

app/views/components/address.scala.html
SteveSugden: Thought of that.  Existing component has var-args as last param which precludes default arg value and most of the places that use that component (lots) don't use named arg lists.

Thought this was the simplest solution

ed0906: 15 usages would need to be fixed, thats not _thaaaat_ many

app/views/components/fields/removable_elements_table.scala.html
```diff
@@ -23,12 +23,14 @@
 
 <div class="field-group mb-2">
     <table id="@{id}">
-      @title.map { heading =>
-          <thead>
-              <tr>
-                <th id="@{id}-heading" colspan="2">@heading</th>
-              </tr>
-          </thead>
+      @if(elements.nonEmpty){
```
ed0906: So if the elements are empty we dont want to show the table title?

Would it not be better to have this `if` block around the entire `table/div`?

SteveSugden: Maybe around the table.  Getting rid of the div might affect the layout/spacing?   Also, there might be tests relying on finding a table by id.   

Thought this was the least invasive change

ed0906: I would presume the tests wouldn't look for the table by ID if it shouldn't be there though?

ed0906: I wouldn't worry about spacing, the margin/padding is on the wrong element if it is required even when the table isn't there

SteveSugden:
test this please

PR #415, opened 2019-08-15
CEDS-1504 Switch Submission to new Back End Endpoint

ed0906:


app/services/SubmissionService.scala
```diff
@@ -54,29 +53,23 @@ class SubmissionService @Inject()(
     val timerContext = exportsMetrics.startTimer(submissionMetric)
     val data = format(exportsDeclaration)
     auditService.auditAllPagesUserInput(getCachedData(exportsDeclaration))
-    (for {
+    for {
       _ <- exportsConnector.create(exportsDeclaration).recover {
-        case error =>
-          Logger.error("V2 Submission failed", error)
-          Future.successful((): Unit)
+        case error: Throwable =>
```
bambuchaAdm: [`NonFatal(e)`](https://www.scala-lang.org/api/2.9.3/scala/util/control/NonFatal$.html)

bambuchaAdm: Maybe extracting this `Future` as variable would make it little more readable

PR #418, opened 2019-08-15
CEDS-1504 Remove legacy mapping

PauloPortugal:
* Remove legacy declaration mapping
* Run ScalaFmt


app/connectors/CustomsDeclareExportsConnector.scala
```diff
@@ -115,5 +116,19 @@ class CustomsDeclareExportsConnector @Inject()(
...
       }
+}
 
+object CustomsDeclareExportsConnector {
+  def toXml(metaData: MetaData): String = {
+    import java.io.StringWriter
+
+    import javax.xml.bind.{JAXBContext, Marshaller}
+
+    val jaxbMarshaller = JAXBContext.newInstance(classOf[MetaData]).createMarshaller
```
bambuchaAdm: This is quite heavy operation - can we move this Val to object and not build marshaller every time ?

bambuchaAdm: Ok - from [Internet](https://timjansen.github.io/jarfiller/guide/jaxb/multithreading.xhtml) we must not share marshaller but Context is perfectly fine.

app/models/ExportsDeclaration.scala
```diff
@@ -46,13 +46,17 @@ case class ExportsDeclaration(
...
-
   def updatedItem(itemId: String, update: ExportItem => ExportItem): ExportsDeclaration =
     itemBy(itemId).fold(this) { item =>
       val updated = update(item)
       copy(items = items.filterNot(_.id.equalsIgnoreCase(itemId)) + updated)
     }
+
+  def itemBy(itemId: String): Option[ExportItem] = items.find(_.id.equalsIgnoreCase(itemId))
+
+  def lrn(): Option[String] = this.consignmentReferences.map { _.lrn }
```
bambuchaAdm: maybe just 'val' ?

PatrykRudnicki: For one line map better is use normal parenthesis instead of curly

PauloPortugal: addressed

PauloPortugal: addressed

PR #422, opened 2019-08-19
CEDS-1507 - date parse error

SteveSugden:


app/services/SubmissionService.scala
```diff
@@ -59,6 +59,7 @@ class SubmissionService @Inject()(
             AuditTypes.Submission,
             auditData(exportsDeclaration.lrn, exportsDeclaration.ducr, Failure.toString)
           )
+        throw error
```
bambuchaAdm: Do know know if `andThen` with `case Failure(exception)` would look nicer - is this align with case ?

SteveSugden: Like this?   Didn't know you can do this.   Much nicer that try-catch-throw.

PR #424, opened 2019-08-19
CEDS-1475 Make DUCR mandatory and personal UCR optional

PauloPortugal:
With this change the user:
 * Must supply a DUCR (known as 2/1 in the tariff)
 * Must supply an LRN (known as 2/5 in the tariff)
 * Can optionally supply a Reference Number (known as 2/4 in the tariff)


test/unit/controllers/declaration/ConsignmentReferencesControllerSpec.scala
```diff
@@ -63,22 +64,56 @@ class ConsignmentReferencesControllerSpec extends ControllerSpec {
...
+      }
+
+      "UCR length has been exceeded" in new SetUp {
+
+        val incorrectForm =
+          Json.toJson(ConsignmentReferences(Ducr("1234"), "", Some(TestHelper.createRandomAlphanumericString(36))))
+
+        val result = controller.submitConsignmentReferences()(postRequest(incorrectForm))
+
+        status(result) must be(BAD_REQUEST)
```
bambuchaAdm: this is more form spec rather then controller

test/unit/controllers/declaration/ConsignmentReferencesControllerSpec.scala
```diff
@@ -63,22 +64,56 @@ class ConsignmentReferencesControllerSpec extends ControllerSpec {
...
+      }
+
+      "UCR has special characters" in new SetUp {
```
PatrykRudnicki: The same here

test/unit/controllers/declaration/ConsignmentReferencesControllerSpec.scala
PauloPortugal: The Form/View Spec is already checking for the personal UCR validation. 
This test is just to assert that the HTTP response code is 400 and not 422, 409 or any other.
A response code needs to be asserted in the controller.

test/unit/controllers/declaration/ConsignmentReferencesControllerSpec.scala
PauloPortugal: The Form/View Spec is already checking for the personal UCR validation.
This test is just to assert that the HTTP response code, in this scenario, is 400 and not 422, 409 or any other.
A response code needs to be asserted in the controller.

PauloPortugal: How would you test that a 400 response code was replied by the controller?
One needs to mimic a failure scenario, ideally all of them that fall into a `Bad Request`. Please advise ;)

PauloPortugal:
@bambuchaAdm I disagree with your request :)
The ViewSpec is already testing all the form validations.
How would you test that a HTTP `400` response code is replied back to the client if a the same client provides some incorrect input?

test/unit/controllers/declaration/ConsignmentReferencesControllerSpec.scala
bambuchaAdm: Maybe we should provide some way to inject form and at some point in binding return errored one ?

PatrykRudnicki: but you checked 400 in tests above. This is just another case for form spec. So for Controller spec we just need one situation that will tests case like `if form has error return 400`, but not try to test every case if this value is longer, if this . value has special characters, etc

PatrykRudnicki: and as I see all tests checks 400 (BAD_REQUEST)

PatrykRudnicki: if you'll have different spec for 422 or 409 (one for every status) it's everything, but now three specs has 400

PauloPortugal: Ok, it seems to me that there was some thought put into this.
I am going to remove these extra controller tests.
 ;) 
Thanks for your passionate replies.

test/unit/controllers/declaration/ConsignmentReferencesControllerSpec.scala
```diff
@@ -63,22 +64,56 @@ class ConsignmentReferencesControllerSpec extends ControllerSpec {
...
+
+        val incorrectForm =
+          Json.toJson(ConsignmentReferences(Ducr("1234"), "", Some("special@&%")))
+
+        val result = controller.submitConsignmentReferences()(postRequest(incorrectForm))
+
+        status(result) must be(BAD_REQUEST)
+      }
```
PatrykRudnicki: Those tests still exists. I still think it's unnecessary, because every test cover the same part of controller codebase and it should be in form spec :)

PatrykRudnicki:
Please rebase your PR

PatrykRudnicki:
@PauloPortugal Please check build, there is a lot of errors after rebase

PR #425, opened 2019-08-19
CEDS-1484 Amend a rejected declaration

ed0906:


app/controllers/SubmissionsController.scala
```diff
@@ -45,4 +49,15 @@ class SubmissionsController @Inject()(
     } yield Ok(submissionsPage(result.toSeq))
   }
 
+  def amend(id: String): Action[AnyContent] = authenticate.async { implicit request =>
+    customsDeclareExportsConnector.findDeclaration(authenticatedSessionId, id) flatMap {
```
bambuchaAdm: there is `sessionId` on request - we don't have to use `SessionIdAware` trait

app/controllers/SubmissionsController.scala
```diff
@@ -45,4 +49,15 @@ class SubmissionsController @Inject()(
...
+      case Some(declaration) =>
+        exportsCacheService
+          .update(authenticatedSessionId, declaration)
```
bambuchaAdm: is here possible `sessionId` and `declaration.sessionId` could be different ?

app/services/SubmissionService.scala
```diff
@@ -52,7 +52,7 @@ class SubmissionService @Inject()(
     val timerContext = exportsMetrics.startTimer(submissionMetric)
     auditService.auditAllPagesUserInput(getCachedData(exportsDeclaration))
     for {
-      _ <- exportsConnector.create(exportsDeclaration).recover {
+      _ <- exportsConnector.createDeclaration(exportsDeclaration).recover {
```
bambuchaAdm: this is out of scope of this PR but here we would remove cache even connector would return an error

app/views/declaration/summary/summary_page.scala.html
```diff
@@ -74,14 +80,20 @@
 
     @items_section(suppDecData.items)
 
-    @helper.form(action = SummaryController.submitSupplementaryDeclaration(), 'id -> "summary") {
-        @helper.CSRF.formField
+    @if(mode == Mode.AmendMode) {
```
bambuchaAdm: @ed0906 what you think about 
```
mode.onAmend {
  <a class="button mt-3" href="@routes.DispatchLocationController.displayPage()">@messages("site.continue")</a>
}
```

test/unit/controllers/declaration/SealControllerSpec.scala
```diff
@@ -140,6 +139,7 @@ class SealControllerSpec extends ControllerSpec with ScalaFutures with ErrorHand
         val result = controller.submitForm()(postRequestAsFormUrlEncoded(body: _*))
 
         status(result) must be(SEE_OTHER)
+        redirectLocation(result) must be(Some(controllers.declaration.routes.SummaryController.displayPage(Mode.NormalMode).url))
```
PatrykRudnicki: 👍

app/views/submission_notifications.scala.html
```diff
@@ -17,33 +17,42 @@
...
-            }
-        </table>
-    </div>
+        </thead>
+
+        <tbody>
+        @for(notification <- notifications){
+            <tr class="table-row">
+                <td class="table-cell">@notification.status.toString</td>
+                <td class="table-cell">@DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm").format(notification.dateTimeIssued)</td>
```
bambuchaAdm: Please do not build formatter at each request.

conf/app.routes
```diff
@@ -10,12 +10,14 @@ GET        /unauthorised                                controllers.Unauthorised
 
 GET        /notifications                               controllers.NotificationsController.listOfNotifications()
 
-GET        /notifications/:mrn                          controllers.NotificationsController.listOfNotificationsForSubmission(mrn: String)
+GET        /notifications/:id                          controllers.NotificationsController.listOfNotificationsForSubmission(id: String)
```
bambuchaAdm: On backend we moved to `submissions/:id/notifications` maybe on frontend we should also

app/controllers/SubmissionsController.scala
ed0906: We probably dont need this trait any more at all

app/controllers/SubmissionsController.scala
ed0906: No, session id isn't stored in the back end so the only way it could be populated is from the request

ed0906: I'm not sure we actually need `declaration.sessionId` any more to be honest

app/services/SubmissionService.scala
ed0906: I think the "Save and Resume" story will mean we no longer need the cache. I discussed with Steve (who is planning to pick this work up) that we can potentially remove the cache as part of it

app/views/declaration/summary/summary_page.scala.html
ed0906: Hmmmm, I don't think its what we would intuitively think to use

app/views/submission_notifications.scala.html
ed0906: Sorted

conf/app.routes
ed0906: Good suggestion

ed0906: Done

app/views/declaration/summary/summary_page.scala.html
bambuchaAdm: It same concept like foreach on `Option` but for enumeration

app/controllers/SubmissionsController.scala
bambuchaAdm: sessionId was and old `Id` for declaration draft

PR #426, opened 2019-08-19
CEDS-1249 Migrated PreviousDocumentsControllerSpec to be Unit

PatrykRudnicki:


PatrykRudnicki:
@ed0906 I can inline it, it's not a problem, I'll prepare PR for it to inline in all places, because few specs contains something like this

PR #428, opened 2019-08-19
CEDS-1511 Remove deprecated cache service method

ed0906:
I expect as part of Save & Resume that this service will be converted to use the back end rather than Mongo.

This PR preps the service a little by removing some methods that wont be available via the backend

### DIFF
--- a/app/controllers/declaration/AdditionalInformationController.scala
+++ b/app/controllers/declaration/AdditionalInformationController.scala
@@ -61,7 +61,8 @@ class AdditionalInformationController @Inject()(
       val actionTypeOpt = FormAction.bindFromRequest()
 
       val cachedData = exportsCacheService
-        .getItemByIdAndSession(itemId, journeySessionId)
+        .get(journeySessionId)
+        .map(_.flatMap(_.itemBy(itemId)))
         .map(_.flatMap(_.additionalInformation).getOrElse(AdditionalInformationData(Seq())))
 
       cachedData.flatMap { cache =>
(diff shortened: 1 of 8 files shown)


app/services/cache/ExportsCacheService.scala
```diff
@@ -34,14 +34,7 @@ class ExportsCacheService @Inject()(journeyCacheModelRepo: ExportsDeclarationRep
...
-  def getItemByIdAndSession(itemId: String, sessionId: String): Future[Option[ExportItem]] =
-    get(sessionId).map {
-      case Some(model) => model.itemBy(itemId)
-      case _           => None
-    }
-
-  def remove(sessionId: String): Future[FindAndModifyCommand.FindAndModifyResult] =
-    journeyCacheModelRepo.remove(sessionId)
+  def remove(sessionId: String): Future[Unit] =
+    journeyCacheModelRepo.remove(sessionId).map(_ => (): Unit)
```
PatrykRudnicki: I think previous implementation was better. Here doesn't matter what we return, but mapping to Unit IMO doesn't make sense, non-blocking comment

ed0906: Services shouldn't be returning anything database related in my opinion, it breaks separation of concerns

app/controllers/declaration/AdditionalInformationController.scala
```diff
@@ -61,7 +61,8 @@ class AdditionalInformationController @Inject()(
       val actionTypeOpt = FormAction.bindFromRequest()
 
       val cachedData = exportsCacheService
-        .getItemByIdAndSession(itemId, journeySessionId)
+        .get(journeySessionId)
```
bambuchaAdm: request.cacheData

PR #430, opened 2019-08-19
CEDS-1249 Migrated ProcedureCodesControllerSpec to be Unit

PatrykRudnicki:


test/unit/controllers/declaration/ProcedureCodesControllerSpec.scala
```diff
@@ -0,0 +1,257 @@
...
+
+        val (responseForm, responseSeq) = theResponse
+        responseForm.value mustBe empty
+        responseSeq mustBe empty
+      }
+
+      "display page method is invoked with data in cache" in {
+
+        val item = ExportItem(itemId, procedureCodes = Some(ProcedureCodesData(Some("1234"), Seq("123"))))
+        withNewCaching(aDeclaration(withItem(item)))
```
bambuchaAdm: here you make stub two times - one in `beforeEach` and second time here.

I think if something does not fit for each test it should not be in before each

PatrykRudnicki: Fixed

PR #437, opened 2019-08-20
CEDS-1484 Amend Dec should create a new declaration

ed0906:
Fixed a bug where Amending a Dec actually tried to alter the old declaration.
It should create a new one with the same answers which the user can then alter.


app/controllers/SubmissionsController.scala
```diff
@@ -51,9 +51,12 @@ class SubmissionsController @Inject()(
...
     customsDeclareExportsConnector.findDeclaration(id) flatMap {
       case Some(declaration) =>
-        exportsCacheService
-          .update(declaration)
-          .map(_ => Redirect(controllers.declaration.routes.SummaryController.displayPage(Mode.AmendMode)))
+        customsDeclareExportsConnector
+          .createDeclaration(declaration.copy(status = DeclarationStatus.DRAFT))
+          .map { created =>
+            Redirect(controllers.declaration.routes.SummaryController.displayPage(Mode.AmendMode))
+              .addingToSession(ExportsSessionKeys.declarationId -> created.id.get)
```
SteveSugden: I thought this might be missing - adding id to session

PR #438, opened 2019-08-20
CEDS-1475 Fix personal ucr bug

PauloPortugal:
* Align PersonalUCR name with Backend
 * Change DUCR regex validation
 * Review content and validation on PreviousDocuments page


app/forms/Ducr.scala
```diff
@@ -26,7 +26,11 @@ case class Ducr(ducr: String)
 object Ducr {
   implicit val format = Json.format[Ducr]
 
-  private val ducrFormat = "^\\d[A-Z]{2}\\d{12}-[0-9A-Z]{1,19}$"
+  private val ducrFormat = """^[0-9][A-Z][A-Z][0-9A-Z\(\)\-/]{6,32}|
```
bambuchaAdm: Where is specification for this regexp ?

PauloPortugal: Is in a tariff document.
They accept 5 different DUCR formats.
I am just getting this confirmed with Ian.

We have used the same for the movements project.

bambuchaAdm: Would you like to put this as scaladoc over - just to not lose connection from where it comes.

PauloPortugal: This has been addressed.

PR #441, opened 2019-08-20
CEDS-1487 resume draft declaration

SteveSugden:
This implements the basic "Resume Draft Declaration" functionality.


app/config/AppConfig.scala
```diff
@@ -55,6 +55,7 @@ class AppConfig @Inject()(
 
   lazy val declarationsV2 = servicesConfig.getConfString(
     "customs-declare-exports.submit-declaration-v2",
+    // TODO - should that key be renamed as it is used to GET as well as POST/PUT?
     throw new IllegalStateException("Missing configuration for Customs Declarations Exports submit declaration URI")
```
SteveSugden: I can rename the key if there are no objections

app/controllers/SavedDeclarationsController.scala
```diff
@@ -0,0 +1,60 @@
...
+      Ok(savedDeclarationsPage(page))
+    }
+  }
+
+  def continueDeclaration(id: String): Action[AnyContent] = authenticate.async { implicit request =>
+    customsDeclareExportsConnector.findDeclaration(id) flatMap {
+      case Some(declaration) =>
+        Future.successful(
+          Redirect(controllers.declaration.routes.SummaryController.displayPage(Mode.SavedMode))
+            .addingToSession(ExportsSessionKeys.declarationId -> id)
```
SteveSugden: This re-uses the same summary form as used for the normal journey and also amend (failed) declarations.    This page could probably do with a re-design to make sure it provides a consistent user experience.

test/helpers/views/declaration/SavedDeclarationsMessages.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package helpers.views.declaration
+
+trait SavedDeclarationsMessages {
+
+  private val prefix: String = "saved.declarations"
+
+  val title: String = prefix + ".title"
```
PatrykRudnicki: I think it'll be better to not create val prefix, but e.g. title will have whole message key. The same with other messages. It'll be easier if you need to find what message is inside, so easier is to highlight messageKey and search for it

app/config/AppConfig.scala
SteveSugden: I think so too.  I'll change it

app/controllers/SavedDeclarationsController.scala
```diff
@@ -0,0 +1,60 @@
...
+class SavedDeclarationsController @Inject()(
+  authenticate: AuthAction,
+  customsDeclareExportsConnector: CustomsDeclareExportsConnector,
+  mcc: MessagesControllerComponents,
+  savedDeclarationsPage: saved_declarations
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  private val logger = Logger(this.getClass)
+  private val defaultPage = Page()
```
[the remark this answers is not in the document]
SteveSugden: This will need to change when implementing pagination (sep. ticket).  The back-end service already supports it.    I agree that we don't particularly need the 'defaultPage' constant declaring though.  I'll remove it and put a default in the connector.

app/controllers/SavedDeclarationsController.scala
SteveSugden: Yes it would.   But they could come back and continue the declaration they just abandoned.

app/models/Mode.scala
```diff
@@ -22,11 +22,13 @@ sealed trait Mode
 object Mode {
   case object NormalMode extends Mode
   case object AmendMode extends Mode
+  case object SavedMode extends Mode
 
   implicit val jsLiteral: JavascriptLiteral[Mode] = new JavascriptLiteral[Mode] {
     override def to(value: Mode): String = value match {
       case NormalMode => "NormalMode"
       case AmendMode  => "AmendMode"
+      case SavedMode  => "SavedMode"
```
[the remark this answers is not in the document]
SteveSugden: I wasn't sure about this.  I thought maybe "ContinueMode" as in "continuing to work on a saved draft"

test/helpers/views/declaration/SavedDeclarationsMessages.scala
SteveSugden: I agree.  I was just following the pattern set by the other tests.   I'll change it and we can change the other ones sometime

app/connectors/CustomsDeclareExportsConnector.scala
```diff
@@ -82,6 +82,17 @@ class CustomsDeclareExportsConnector @Inject()(appConfig: AppConfig, httpClient:
       .map(_.map(_.toExportsDeclaration))
   }
 
+  def findSavedDeclarations(
```
ed0906: Could potentially reuse the find declarations method (with optional filters & sort params) but this is fine too

app/controllers/SavedDeclarationsController.scala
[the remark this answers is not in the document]
ed0906: I think it's fine where it is tbh, makes it easier to implement pagination later on having it at this level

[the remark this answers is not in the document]
ed0906: I maybe wouldnt use default pagination though as you ideally want `pageSize=Integer.MAX`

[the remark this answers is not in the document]
SteveSugden: The ticket calls for 50 items on page (at this point).    Testers will be bound to test that!

app/controllers/SavedDeclarationsController.scala
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  def continueDeclaration(id: String): Action[AnyContent] = authenticate.async { implicit request =>
+    customsDeclareExportsConnector.findDeclaration(id) flatMap {
+      case Some(declaration) =>
+        Future.successful(
+          Redirect(controllers.declaration.routes.SummaryController.displayPage(Mode.SavedMode))
+            .addingToSession(ExportsSessionKeys.declarationId -> id)
+        )
+      case _ => Future.successful(Redirect(controllers.routes.SavedDeclarationsController.displayDeclarations()))
```
ed0906: 👍

app/models/Mode.scala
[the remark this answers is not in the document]
ed0906: `DraftMode`?

Tbh I regret sticking `Mode` at the end already, it reads better just as `Mode.Normal`, `Mode.Amend`

app/views/saved_declarations.scala.html
```diff
@@ -0,0 +1,69 @@
...
+
+@main_template(
+    title = messages("saved.declarations.title")
+) {
+
+    @components.back_link(routes.ChoiceController.displayPage().url)
+
+    <h1>@messages("saved.declarations.title")</h1>
+
+    <div>
```
ed0906: Any reason for the outer `div`?

app/views/saved_declarations.scala.html
```diff
@@ -0,0 +1,69 @@
...
+        <table class="govuk-table">
```
ed0906: I don't think these `govuk` styles do anything

app/views/saved_declarations.scala.html
```diff
@@ -0,0 +1,69 @@
...
+@import views.components.inputs.RadioOption
+@import forms.cancellation.CancellationChangeReason
+@import forms.cancellation.CancellationChangeReason.{NoLongerRequired, Duplication, OtherReason}
+
+@this(main_template: views.html.main_template)
+
+@(declarations: Paginated[ExportsDeclaration])(implicit request: Request[_], messages: Messages)
+
+@main_template(
+    title = messages("saved.declarations.title")
```
ed0906: FYI you can use `fullWidth = true` if the table looks crammed

app/views/saved_declarations.scala.html
SteveSugden: I based this page on 'submissions' which has the div

app/views/saved_declarations.scala.html
SteveSugden: It was.  Seems to look ok.

test/unit/controllers/SavedDeclarationsControllerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    "return 303 (SEE_OTHER)" when {
+       "continue declaration method is invoked" in new SetUp {
+
+         getDeclaration("123")
+
+         val result = controller.continueDeclaration("123")(getRequest())
+
+         status(result) must be(SEE_OTHER)
```
ed0906: Can we assert that the session property is added please :)

`session(result).get(ExportsSessionKeys.declarationId) must be(Some("123"))

test/unit/controllers/SavedDeclarationsControllerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+
+        listOfDraftDeclarations()
+
+        val result = controller.displayDeclarations()(getRequest())
+
+        status(result) must be(OK)
+      }
+    }
+
+    "return 303 (SEE_OTHER)" when {
```
ed0906: What about when the declaration is not found?

app/views/saved_declarations.scala.html
ed0906: I don't think either of them should have it tbh (it doesn't offer anything extra)

app/views/saved_declarations.scala.html
ed0906: I bet if all the `govuk` classes were removed it would look no different :)

app/views/declaration/summary/summary_page.scala.html
```diff
@@ -84,10 +85,6 @@
 
     @items_section(suppDecData.items)
 
-    @if(mode == Mode.AmendMode) {
```
bambuchaAdm: why we delete amend section ?

SteveSugden: It's not deleted.  Its moved to the else block

app/views/ViewDates.scala
```diff
@@ -16,13 +16,17 @@
...
 import java.time.temporal.TemporalAccessor
 
 object ViewDates {
 
   val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm")
+  val dateFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("d MMM uuu").withZone(ZoneId.systemDefault)
+  val timeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("HH:mm").withZone(ZoneId.systemDefault)
 
   def format(temporal: TemporalAccessor): String = formatter.format(temporal)
+  def formatDateAtTime(temporal: TemporalAccessor): String = s"${dateFormatter.format(temporal)} at ${timeFormatter.format(temporal)}"
```
bambuchaAdm: This could be done with single formatter
```
DateTimeFormatter.ofPattern("d MMM uuu 'at' HH:mm").withZone(ZoneId.systemDefault)
```

SteveSugden: Didn't know you could do that (quote the literal text).  Thanks

app/views/declaration/summary/summary_page.scala.html
ed0906: It's moved into an `else` block

app/views/saved_declarations.scala.html
SteveSugden: Should I use the same css classes as the 'submissions' table then?

SteveSugden:
test this please

PR #442, opened 2019-08-21
CEDS-1475 Rename HTML ids to `Related document`

PauloPortugal:
This was a request from @leszeklorek so that the CI tests can search
the html elements by their new name, making the domain language in code,
test and business consistent.

### DIFF
--- a/test/unit/controllers/declaration/SummaryControllerSpec.scala
+++ b/test/unit/controllers/declaration/SummaryControllerSpec.scala
@@ -121,11 +121,8 @@ class SummaryControllerSpec extends ControllerSpec with ErrorHandlerMocks with O
         val declaration = aDeclaration()
         withNewCaching(aDeclaration())
         when(
-          mockSubmissionService.submit(any[ExportsDeclaration])(
-            any[JourneyRequest[_]],
-            any[HeaderCarrier],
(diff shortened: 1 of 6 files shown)


test/unit/controllers/declaration/SummaryControllerSpec.scala
```diff
@@ -134,7 +131,11 @@ class SummaryControllerSpec extends ControllerSpec with ErrorHandlerMocks with O
         session(result).get(ExportsSessionKeys.declarationId) must be(None)
         redirectLocation(result) must be(Some(controllers.declaration.routes.ConfirmationController.displayPage().url))
         flash(result).get("LRN") must be(Some("123LRN"))
-        verify(mockSubmissionService).submit(any[ExportsDeclaration])(any[JourneyRequest[_]], any[HeaderCarrier], any[ExecutionContext])
+        verify(mockSubmissionService).submit(any[ExportsDeclaration])(
+          any[JourneyRequest[_]],
```
PatrykRudnicki: In general I'm thinking that it's good to add those types for any. I mean `JourneyRequest[_]` etc, there is no option to invoke it with different type there. Just any() will be shorter, but I can be wrong about it.

PR #443, opened 2019-08-21
CEDS-1249 The last spec migrated to be Unit

PatrykRudnicki:


test/unit/controllers/declaration/ItemTypeControllerSpec.scala
```diff
@@ -0,0 +1,245 @@
...
+
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
+    authorizedUser()
+    setupErrorHandler()
+    when(mockItemTypePage.apply(any(), any(), any(), any(), any())(any(), any()))
+      .thenReturn(HtmlFormat.empty)
+  }
+
+  val itemId = "itemId"
```
bambuchaAdm: Maybe item id generation service here ?

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
```diff
@@ -0,0 +1,245 @@
...
+    "return 200 (OK)" when {
+
+      "item exists in the cache and item type is defined" in {
+
+        withNewCaching(aDeclaration(withItem(anItem(withItemId(itemId), withItemType()))))
+
+        val result = controller.displayPage(itemId)(getRequest())
+
+        status(result) mustBe OK
+        verify(mockItemTypePage, times(1)).apply(any(), any(), any(), any(), any())(any(), any())
```
bambuchaAdm: it is done in next line - if there was

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
```diff
@@ -0,0 +1,245 @@
...
+
+        val correctForm = Seq(
+          ("combinedNomenclatureCode", "code"),
+          ("taricAdditionalCode[]", ""),
+          ("nationalAdditionalCode[]", ""),
+          ("descriptionOfGoods", "description"),
+          ("cusCode", ""),
+          ("unDangerousGoodsCode", ""),
+          ("statisticalValue", "1234"),
+          addActionUrlEncoded
```
MaciejRewera: Why is there supposed to be 303 code if the action is for adding? Should it not the same page be refreshed?

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
```diff
@@ -0,0 +1,245 @@
...
+
+        withNewCaching(aDeclaration())
+
+        val result = controller.displayPage(itemId)(getRequest())
+
+        status(result) mustBe SEE_OTHER
+        verify(mockItemTypePage, times(0)).apply(any(), any(), any(), any(), any())(any(), any())
+      }
+
+      "correct item is added and declaration model exist in the cache" in {
```
MaciejRewera: This is confusing, because in our case the `item` is something bigger. And you are actually not adding the whole `item`, but only some information about it. Instead of `item` I'd suggest even the simplest `data`.

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
```diff
@@ -0,0 +1,245 @@
...
+          saveAndContinueActionUrlEncoded
+        )
+
+        val result = controller.submitItemType(itemId)(postRequestAsFormUrlEncoded(correctForm: _*))
+
+        status(result) mustBe SEE_OTHER
+        verify(mockItemTypePage, times(0)).apply(any(), any(), any(), any(), any())(any(), any())
+      }
+
+      "item has been removed with model in cache" in {
```
MaciejRewera: I think "piece of data has been removed and there is model in cache" explains better.

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
PatrykRudnicki: Will do it

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
PatrykRudnicki:
```
private def handleAddition(
    itemId: String,
    itemTypeInput: ItemType,
    itemTypeCache: ItemType,
    hasFiscalReferences: Boolean
  )(implicit request: JourneyRequest[AnyContent]): Future[Result] = {
    val itemTypeUpdated = updateCachedItemTypeAddition(itemId, itemTypeInput, itemTypeCache)
    ItemTypeValidator.validateOnAddition(itemTypeUpdated) match {
      case Valid =>
        updateExportsCache(itemId, itemTypeUpdated).map {
          case Some(model) =>
            refreshPage(itemId, itemTypeInput, model)
          case None =>
            Redirect(routes.ItemsSummaryController.displayPage())
        }
      case Invalid(errors) =>
        val formWithErrors =
          errors.foldLeft(ItemType.form().fill(itemTypeInput))((form, error) => form.withError(adjustErrorKey(error)))
        Future.successful(
          BadRequest(
            itemTypePage(
              itemId,
              adjustDataKeys(formWithErrors),
              hasFiscalReferences,
              itemTypeCache.taricAdditionalCodes,
              itemTypeCache.nationalAdditionalCodes
            )
          )
        )
    }
  }
```

PatrykRudnicki: status it's OK (200) I don't understand your question

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
PatrykRudnicki: You're right, changing....

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
PatrykRudnicki: sorted, thanks :)

PR #444, opened 2019-08-21
CEDS-1511 POC for Save & Resume

ed0906:
I haven't done tests yet so the PR builder will fail


app/controllers/declaration/AdditionalDeclarationTypeController.scala
```diff
@@ -48,16 +51,22 @@ class AdditionalDeclarationTypeController @Inject()(
...
+        decType
+          .fold(
+            formWithErrors => Future.successful(BadRequest(declarationTypePage(formWithErrors))),
+            validAdditionalDeclarationType =>
+              updateCache(validAdditionalDeclarationType).map { _ =>
+                Redirect(controllers.declaration.routes.ConsignmentReferencesController.displayPage())
+            }
+          )
+      case Some(SaveAndReturn) =>
+        Future.successful(navigator.goToDraftConfirmation())
```
ed0906: This discards all  partially completed answers and saves

app/controllers/declaration/AdditionalDeclarationTypeController.scala
```diff
@@ -48,15 +51,21 @@ class AdditionalDeclarationTypeController @Inject()(
...
         formWithErrors => Future.successful(BadRequest(declarationTypePage(formWithErrors))),
         validAdditionalDeclarationType =>
-          updateCache(validAdditionalDeclarationType)
-            .map(_ => Redirect(controllers.declaration.routes.ConsignmentReferencesController.displayPage()))
+          updateCache(validAdditionalDeclarationType).map { _ =>
+            action match {
+              case Some(SaveAndContinue) =>
+                Redirect(controllers.declaration.routes.ConsignmentReferencesController.displayPage())
+              case Some(SaveAndReturn) =>
+                navigator.goToDraftConfirmation()
```
ed0906: So the only way to "save and resume" is to complete the page with valid data (agreed with Ian & Ben M)

bambuchaAdm: Is SaveAndReturn always goes to this page ?

app/controllers/navigation/Navigator.scala
```diff
@@ -0,0 +1,35 @@
...
+
+package controllers.navigation
+
+import config.AppConfig
+import javax.inject.Inject
+import models.requests.JourneyRequest
+import models.responses.FlashKeys
+import play.api.mvc.{Result, Results}
+
+class Navigator @Inject()(appConfig: AppConfig) {
```
bambuchaAdm: Can we do this with provider and `play.api.Configuraiton` ?

app/controllers/declaration/AdditionalDeclarationTypeController.scala
```diff
@@ -48,15 +51,21 @@ class AdditionalDeclarationTypeController @Inject()(
...
     decType
-      .form()
-      .bindFromRequest()
       .fold(
         formWithErrors => Future.successful(BadRequest(declarationTypePage(formWithErrors))),
         validAdditionalDeclarationType =>
-          updateCache(validAdditionalDeclarationType)
-            .map(_ => Redirect(controllers.declaration.routes.ConsignmentReferencesController.displayPage()))
+          updateCache(validAdditionalDeclarationType).map { _ =>
+            action match {
```
bambuchaAdm: What if action is None ?

What do you think about link navigator and action together ?

app/controllers/navigation/Navigator.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def goToDraftConfirmation()(implicit req: JourneyRequest[_]): Result = {
+    val updatedDateTime = req.cacheModel.updatedDateTime
+    val expiry = updatedDateTime.plusSeconds(appConfig.draftTimeToLive.toSeconds)
+    Results.Redirect(
+      controllers.declaration.routes.ConfirmationController.displayDraftConfirmation()
+    ).flashing(FlashKeys.expiryDate -> expiry.toEpochMilli.toString)
```
SteveSugden: Is this the best way of getting config data into the view?   Why not just give the view the AppConfig?

app/models/responses/FlashKeys.scala
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.responses
+
+object FlashKeys {
```
PatrykRudnicki: 👍

app/controllers/navigation/Navigator.scala
ed0906: So I think the actual declaration is meant to be out of memory at the point of the confirmation controller, but it needs to know the expiry date. 
So options are: 
- use session storage
- use flash storage (one request session storage)
- don't remove declaration from memory

app/controllers/navigation/Navigator.scala
ed0906: I cant store it as a reusable variable that way. The Time to Live needs to be used in multiple places

app/controllers/navigation/Navigator.scala
SteveSugden: Didn't realise you'd lost the declaration at that point.

app/controllers/declaration/AdditionalDeclarationTypeController.scala
ed0906: Action = None means you've hacked the UI, i'm not sure we are too bothered :)

app/controllers/navigation/Navigator.scala
bambuchaAdm: It always could be bind and injected to multiple places

app/models/responses/FlashKeys.scala
bambuchaAdm: Why not use TypeClass for those ? In most cases we don't store multiple items with same type in Flash ?

app/views/declaration/submission_confirmation_page.scala.html
```diff
@@ -35,10 +37,10 @@ <h1 class="heading-medium">@messages("supplementary.confirmation.whatHappensNext
...
     <p>
         @messages("supplementary.confirmation.explanation")
-        <a href="/customs-declare-exports/submissions">
+        <a href="@routes.SubmissionsController.displayListOfSubmissions()">
             @messages("supplementary.confirmation.explanation.linkText")
         </a>
     </p>
 
-    @components.button_link(messageKey = "supplementary.confirmation.submitAnotherDeclaration", href = "/customs-declare-exports/choice")
+    @components.button_link(messageKey = "supplementary.confirmation.submitAnotherDeclaration", href=routes.ChoiceController.displayPage().url)
```
bambuchaAdm: Maybe button_link should take `Call` instead `href` string

app/controllers/declaration/AdditionalDeclarationTypeController.scala
bambuchaAdm: So maybe we should move away from `Option` here - if it is meaningless why we should bother about it.

app/views/declaration/submission_confirmation_page.scala.html
ed0906: Probably, it's used in quite a few places though so I dont want to change it now

app/controllers/declaration/AdditionalDeclarationTypeController.scala
ed0906: Yep

ed0906: Perhaps we can expand the navigator eventually to redirect based on the action

app/controllers/navigation/Navigator.scala
ed0906: Well, I'd missed the code to pull it from the session but I've added it now :D

app/controllers/declaration/AdditionalDeclarationTypeController.scala
ed0906: Good question, one for the tech debt board

app/models/responses/FlashKeys.scala
ed0906: Can you show me an example?

app/controllers/navigation/Navigator.scala
ed0906: I think we need to agree we as a team if we do/don't want to go with that approach in the future. I personally have never found `AppConfig` causes any issues

app/models/responses/FlashKeys.scala
bambuchaAdm:
```
case class LRN(value: String)

object LRN {
   implicit val flashBindalbe = FlashBindable[LNR]("LRN", _.value)
}
```
On request site
```
.flashing(FlashFor(lnr))
````

Comment (ed0906) [reply app/models/responses/FlashKeys.scala]: I agree that it'd be nice to make it type safe but Flash is a play class which doesn't seem to have any binder interface as far as I can see

File: app/controllers/util/FormAction.scala [hunk 8]
```diff
@@ -25,13 +25,15 @@ sealed trait FormAction {
...
   private val saveAndContinueLabel = "SaveAndContinue"
+  private val saveAndReturnLabel = "SaveAndReturn"
   private val continueLabel = "Continue"
   private val removeLabel = "Remove"
 
   private def fromUrlEncoded(input: Map[String, Seq[String]]): FormAction =
     input.flatMap {
       case (`addLabel`, _)             => Some(Add)
       case (`saveAndContinueLabel`, _) => Some(SaveAndContinue)
+      case (saveAndReturnLabel, _)     => Some(SaveAndReturn)
```
PatrykRudnicki: Can you add ``` like in rest of the cases to be consistent with the actual code.

ed0906: Pretty sure this was already fixed wasnt it?

PR #446, opened 2019-08-22
CEDS-1511 Add Save & Resume to Declaration Journey

ed0906:


ed0906:
test this please

test/unit/controllers/declaration/AdditionalInformationControllerSpec.scala
```diff
@@ -197,8 +198,8 @@ class AdditionalInformationControllerSpec extends ControllerSpec with ErrorHandl
 
         val result = controller.saveAdditionalInfo("itemId")(postRequestAsFormUrlEncoded(correctForm: _*))
 
-        status(result) must be(SEE_OTHER)
-        redirectLocation(result) must be(Some(routes.DocumentsProducedController.displayPage("itemId").url))
+        await(result) mustBe aRedirectToTheNextPage
```
PatrykRudnicki: Why await? Why we're checking this differently? 

If we want to do it I prefer to extends ScalaFutures and use `.futureValue` on result. This comment is for all changes like this below

ed0906: Why not use await? :)

We are checking it differently because the result is returned from the navigator which is mocked

ed0906: In hindsight there is nicer ways to assert on this e.g. by mocking the navigator so it returns a `Redirect(the route that was passed in)`

But Paulo is working on this too so I don't want to cause all conflicts for him

PatrykRudnicki: ah, okay.

Probably `await` is from hmrctest, as I remember, that is deprecated (repo not method), so it's good to avoid more places with this, because we need to ged rid of this library. Moreover, `futureValue` is from ScalaFutures. It's part of scalatest, so IMO it's better choice, but the second answer is related with my personal thinking, so it's not something that must be or not

ed0906: It's from `play.api.test.Helpers` I believe not hmrctest

Although I think theres a similar one in hmrctest

app/controllers/declaration/DestinationCountriesController.scala
```diff
@@ -97,7 +99,7 @@ class DestinationCountriesController @Inject()(
 
     actionTypeOpt match {
       case Some(Add) if !boundForm.hasErrors             => addRoutingCountry(cache)
-      case Some(SaveAndContinue) if !boundForm.hasErrors => saveAndContinue(cache)
+      case Some(SaveAndContinue) | Some(SaveAndReturn) if !boundForm.hasErrors => saveAndContinue(cache)
```
PauloPortugal: :+1:

app/controllers/declaration/DocumentsProducedController.scala
```diff
@@ -65,14 +67,14 @@ class DocumentsProducedController @Inject()(
 
     actionTypeOpt match {
       case Some(Add) if !boundForm.hasErrors             => addItem(itemId, boundForm.get, cache)
-      case Some(SaveAndContinue) if !boundForm.hasErrors => saveAndContinue(itemId, boundForm.get, cache)
+      case Some(SaveAndContinue) |  Some(SaveAndReturn) if !boundForm.hasErrors => saveAndContinue(itemId, boundForm.get, cache)
```
PauloPortugal: :+1:

PauloPortugal:
@ed0906 I think we should add ViewSpec unit tests here, to ensure a button is present (so to keep consistent with the rest of the tests)

ed0906:
> @ed0906 I think we should add ViewSpec unit tests here, to ensure a button is present (so to keep consistent with the rest of the tests)

I was unsure if it would add much value as we'd be asserting on an ID thats always there. 
I guess we could write a matcher e.g `doc must contain(saveAndReturnButton)`

test/unit/controllers/declaration/AdditionalInformationControllerSpec.scala
bambuchaAdm: @PatrykRudnicki await is also from play-test-helpers

ed0906:
View Specs added @PauloPortugal

PR #447, opened 2019-08-22
CEDS-1524 draft dec pagination

SteveSugden:


app/config/AppConfig.scala
```diff
@@ -105,6 +105,8 @@ class AppConfig @Inject()(
   lazy val draftTimeToLive: FiniteDuration =
     servicesConfig.getDuration("draft.timeToLive").asInstanceOf[FiniteDuration]
 
+  lazy val draftItemPerPage: Int = servicesConfig.getInt("draft.itemsPerPage")
```
ed0906: I imagine the 50 limit will be service wide rather than just for drafts

ed0906: But no harm keeping it as it is :)

SteveSugden: You're probably right.  I'll rename it.

app/views/components/PaginationUtil.scala
```diff
@@ -0,0 +1,47 @@
...
+    if (max+1 <= pageCount && min-1 >= 1) {
+      (min-1) +: list :+ (max+1)
+    } else if (max+1 <= pageCount) {
+      list :+ (max+1)
+    } else if (min-1 >= 1) {
+      (min-1) +: list
+    } else {
+      list
+    }
+  }
```
ed0906: What a fantastically well written, readable method 💯 I wonder who wrote it...

SteveSugden: I did say it was ripped off in the PR request.   Take a bow Mr W

app/views/components/PaginationUtil.scala
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package views.components
+
+import scala.collection.immutable.Seq
+
+object PaginationUtil {
+
+  def surroundingPages(pageIndex: Int, pageCount: Int, available: Int): Seq[Int] = {
```
PatrykRudnicki: Can we make this function without vars? just recursive function

app/views/components/PaginationUtil.scala
PatrykRudnicki: Hmm, I think removing parenthesis and putting in line with if will be more readable
```
if (max+1 <= pageCount && min-1 >= 1) (min-1) +: list :+ (max+1)
else if (max+1 <= pageCount) list :+ (max+1)
else if (min-1 >= 1) (min-1) +: list
else list
```

or just stay with current  format but without parenthesis. difficult to understand method's behaviour

app/views/components/PaginationUtil.scala
SteveSugden: This class was lifted from binding-tariffs.  I haven't really studied it to understand how it works, but it's tested/proven code and I'd rather just leave it as-is.  Unless we have nothing else to do.

PR #452, opened 2019-08-23
CEDS-1487 - fix couple of issues with 'choice' page

SteveSugden:
- order (and text) of choice items changed to reflect latest prototype
- cleared the declarationId from session when landing on the "saved drafts" page so that when you navigate back to the choice page the previously viewed dec type is not pre-selected.

### DIFF
--- a/app/controllers/SavedDeclarationsController.scala
+++ b/app/controllers/SavedDeclarationsController.scala
@@ -40,7 +40,7 @@ class SavedDeclarationsController @Inject()(
 
   def displayDeclarations(pageNumber: Int = 1): Action[AnyContent] = authenticate.async { implicit request =>
     customsDeclareExportsConnector.findSavedDeclarations(Page(pageNumber, appConfig.paginationItemsPerPage)).map { page =>
-      Ok(savedDeclarationsPage(page))
+      Ok(savedDeclarationsPage(page)).removingFromSession(ExportsSessionKeys.declarationId)
     }
   }
 
--- a/app/views/choice_page.scala.html
+++ b/app/views/choice_page.scala.html
@@ -39,11 +39,11 @@
                 field = form("value"),
                 legend = "",
                 inputs = Seq(
-                    RadioOption("Supplementary declaration", SupplementaryDec, messages("declaration.choice.SMP")),
                     RadioOption("Standard declaration", StandardDec, messages("declaration.choice.STD")),
+                    RadioOption("Supplementary declaration", SupplementaryDec, messages("declaration.choice.SMP")),
+                    RadioOption("Submissions", Submissions, messages("declaration.choice.SUB")),
                     RadioOption("Continue declaration", ContinueDec, messages("declaration.choice.CON")),
-                    RadioOption("Cancel declaration", CancelDec, messages("declaration.choice.CAN")),
-                    RadioOption("Submissions", Submissions, messages("declaration.choice.SUB"))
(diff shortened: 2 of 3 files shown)


app/controllers/SavedDeclarationsController.scala
```diff
@@ -40,7 +40,7 @@ class SavedDeclarationsController @Inject()(
 
   def displayDeclarations(pageNumber: Int = 1): Action[AnyContent] = authenticate.async { implicit request =>
     customsDeclareExportsConnector.findSavedDeclarations(Page(pageNumber, appConfig.paginationItemsPerPage)).map { page =>
-      Ok(savedDeclarationsPage(page))
+      Ok(savedDeclarationsPage(page)).removingFromSession(ExportsSessionKeys.declarationId)
```
ed0906: Why have this here?

PauloPortugal:
I am a bit of a git nerd, so take this with a pinch of salt.
I love your description on the PR. It would be awesome if you added those to the commit, so that the next time we annotate the code or do `git blame` we understand **why** we did something in the past.

`git commit --amend` and add those comments.
`git push origin +CEDS-1487-choice-issues --force-with-lease` should do it

app/controllers/SavedDeclarationsController.scala
SteveSugden: Why not?  Seemed like a good place to clear up session state.  We go from the list of saved decs to editing a dec (setting the dec id) landing on the summary page.  If we click "back"  we return to the saved dec list and back again takes us to the choice page with the dec type pre-selected.

Alternatively, we could clear the id when loading the choice page.   But then someone would complain when coming all the way back from doing a declaration.

PauloPortugal: > cleared the declarationId from session when landing on the "saved drafts" page so that when you navigate back to the choice page the previously viewed dec type is not pre-selected.
Probably we should add a test for this functionality.

SteveSugden: Adding test... But are we happy that this is the right place to clear id?

SteveSugden: Any ideas how to test this?  I thought of creating a new controller test like this 
```
    "clear declarationId from session" when {
      "returning to display saved declarations" in new SetUp {
        getDeclaration("123")
        val editResult = controller.continueDeclaration("123")(getRequest())
        session(editResult).get(ExportsSessionKeys.declarationId) must be(Some("123"))

        listOfDraftDeclarations()
        val listResult = controller.displayDeclarations()(getRequest())
        session(listResult).get(ExportsSessionKeys.declarationId) must be(None)
      }
```
But that passes even without my change.  The second session is not related to the first.

ed0906: Ah it’s for back button behaviour 👍🏽 I’d prefer this kind of thing to be on the “exiting editing a declaration” actions e.g submit, save for later but I guess we need to handle back button logic somehow too

ed0906:
I agree with Paulos comment *minus* the force with lease bit. Perhaps we should have a team chat about git hygiene at some point. You can always copy the description into the Squash and Merge commit body 👍🏽

SteveSugden:
I very rarely manage to complete a ticket with a single commit and therefore always do a squash and merge into master.  (So my nice commit commit comments would be lost wouldn't they).  I'll make sure the merge commit message is descriptive

PR #453, opened 2019-08-27
CEDS-1544 Removed hmrctest

PatrykRudnicki:


test/base/ViewValidator.scala
```diff
@@ -32,7 +32,7 @@ trait ViewValidator extends MustMatchers with ViewMatchers {
   private def asDocument(page: String): Document = Jsoup.parse(page)
 
   //TODO Remove methods based on the CSS, stay and use those based on the ID
-  @deprecated("Please use 'page.getElementsBySelector must ...'")
+  @deprecated("Please use 'page.getElementsBySelector must ...'", since = "2019-08-07")
```
PatrykRudnicki: Using `deprecated` with only one argument is deprecated :D

test/connectors/CustomsDeclareExportsConnectorIntegrationSpec.scala
```diff
@@ -46,6 +47,9 @@ class CustomsDeclareExportsConnectorIntegrationSpec
   private val notification = Notification("conv-id", "mrn", LocalDateTime.now, "f-code", None, Seq.empty, "payload")
   private val connector = new CustomsDeclareExportsConnector(config, httpClient)
 
+  implicit val defaultPatience: PatienceConfig =
```
PatrykRudnicki: First test take more time than default timeout

test/connectors/CustomsDeclareExportsConnectorIntegrationSpec.scala
```diff
@@ -62,9 +66,9 @@ class CustomsDeclareExportsConnectorIntegrationSpec
           )
       )
 
-      val response = await(connector.createDeclaration(newDeclaration))
```
PatrykRudnicki: I can't find await method in scalatest, this one was from hmrc test

test/services/model/AutoCompleteItemSpec.scala
```diff
@@ -18,64 +18,64 @@ package services.model
...
     "map from document type" in {
-      AutoCompleteItem.fromDocumentType(List(DocumentType("description", "code"))) shouldBe List(
+      AutoCompleteItem.fromDocumentType(List(DocumentType("description", "code"))) mustBe List(
         AutoCompleteItem("description - code", "code")
       )
     }
 
     "map from national additional code" in {
-      AutoCompleteItem.fromNationalAdditionalCode(List(NationalAdditionalCode("code"))) shouldBe List(
+      AutoCompleteItem.fromNationalAdditionalCode(List(NationalAdditionalCode("code"))) mustBe List(
```
PatrykRudnicki: Our base spec extends MustMatchers, so probably better to stay with one approach

test/connectors/CustomsDeclareExportsConnectorIntegrationSpec.scala
ed0906: `import play.api.test.Helpers._`

PatrykRudnicki: ah, okay, I'll change it here :)

PatrykRudnicki: Changed, waiting for build and merge

test/connectors/CustomsDeclareExportsConnectorIntegrationSpec.scala
```diff
@@ -64,7 +69,7 @@ class CustomsDeclareExportsConnectorIntegrationSpec
 
       val response = await(connector.createDeclaration(newDeclaration))
 
-      response shouldBe existingDeclaration
+      response mustBe existingDeclaration
```
PauloPortugal: why `mustBe`?

PatrykRudnicki: `UnitSpec` form `hmrctest` extends `Matchers`, our `UnitSpec` extends `MustMatchers` so we don't have access to `shouldBe`.

To have it we need to extends `Matchers` but IMO it doesn't make any sense to add additional `Matchers`. `mustBe` is consistent with actual approach in other tests.

PR #454, opened 2019-08-27
CEDS-1490 First round of cleanup

PatrykRudnicki:


test/base/CustomExportsBaseSpec.scala
```diff
@@ -151,54 +120,4 @@ trait CustomExportsBaseSpec
...
-    )
-
-    FakeRequest("POST", uri)
-      .withHeaders(Map(cfg.headerName -> token).toSeq: _*)
-      .withSession(session.toSeq: _*)
-      .withFormUrlEncodedBody(body: _*)
-      .withCSRFToken
-  }
-
-  // FIXME should be promoted to MockExportsCacheService trait
```
bambuchaAdm: Did you address this comment? AFAIR those two versions are different

PatrykRudnicki: Those methods are mocked in the `MockExportCacheService` in exactly the same way

PR #455, opened 2019-08-27
CEDS-1511 Render Draft Expiry Date Properly

ed0906:


app/views/declaration/draft_confirmation_page.scala.html
```diff
@@ -28,17 +30,26 @@
         <h1 class="heading-xlarge">
             @messages("confirmation.draft.title")
         </h1>
-        <p class="font-large">@messages("confirmation.draft.info", flash.get(FlashKeys.expiryDate).getOrElse("-"))</p>
+        <p class="font-large">
+         @messages("confirmation.draft.info")
+        </p>
+        <p id="draft_confirmation-expiry" class="font-large">
+        @{
+            flash.get(FlashKeys.expiryDate)
```
PauloPortugal: are you storing it in the `flash` store?

test/views/declaration/DraftConfirmationViewSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+import play.api.mvc.Flash
+import play.twirl.api.Html
+import views.declaration.spec.ViewSpec
+import views.html.declaration.draft_confirmation_page
+import views.tags.ViewTest
+
+@ViewTest
+class DraftConfirmationViewSpec extends ViewSpec with CommonMessages {
+
+  private val page = app.injector.instanceOf[draft_confirmation_page]
```
bambuchaAdm: What do you think about move this to unit test ?

PatrykRudnicki: I prefer to create page manually or use mock.

To create:
`val page = new draft_confirmation_page(mainTemplate)`
You should have `mainTemplate` in scope. We will avoid using this app and injector. It'll make future easier when we'll start removing app from view specs

app/views/declaration/draft_confirmation_page.scala.html
```diff
@@ -28,17 +30,26 @@
         <h1 class="heading-xlarge">
             @messages("confirmation.draft.title")
         </h1>
-        <p class="font-large">@messages("confirmation.draft.info", flash.get(FlashKeys.expiryDate).getOrElse("-"))</p>
+        <p class="font-large">
+         @messages("confirmation.draft.info")
```
SteveSugden: That message key needs the {0} param removing I think.     You've also increased the separation between the message and the date by introducing a new paragraph.  Why not <p> Message <br/> Date </p> ?

app/views/declaration/draft_confirmation_page.scala.html
ed0906: Yes, the Navigator handles that aspect

app/views/declaration/draft_confirmation_page.scala.html
ed0906: I removed the `{0}` from the message and made it into two lines so the date is always on the same line.

Not sure `<br/>` gives us anything over just having two `<p>` tags, unless one is more accessible?

SteveSugden: Missed the change to message.en.    Difference is single line spacing.

test/views/declaration/DraftConfirmationViewSpec.scala
ed0906: Perhaps, all the view specs appear to follow this pattern though

bambuchaAdm: injector != application - you don't have to run application to have only injector in scope

PR #457, opened 2019-08-27
BAU - TamperMonkey script - select "Standard Declaration" on choice page

SteveSugden:


ed0906:
Presume this was meant to be merged? 
We need it for the demo tomorrow so I’ve merged for you

PR #458, opened 2019-08-27
CEDS-1506 Ensure Each Page knows what "Mode" it is in

ed0906:
This allows us to have custom "Back", "Save & Continue" & "Save & Return" behaviour depending on which "Mode" the page is in e.g. "Amending a Declaration", "Checking Answers", "Editing a Draft"

It is represented as a query param on each Journey page `mode=Amend` 
When the param is not present in the URL it defaults to `NormalMode`


bambuchaAdm:
What do you think about taking mode only as `JourneyRequest` field not as actionParameter ?

ed0906:
test this please

ed0906:
> What do you think about taking mode only as `JourneyRequest` field not as actionParameter ?

It needs to be possible (and obvious) to link to controllers in a certain mode e.g.

`Redirect(controllers.declaration.routes.SummaryController.displayPage(someMode)`

Therefore it needs to be a parameter on the controller method rather than the request

app/views/components/back_link.scala.html
```diff
@@ -14,8 +14,18 @@
  * limitations under the License.
  *@
 
-@(href: String)(implicit messages: Messages)
+@import models.Mode
+@import models.Mode._
+ 
+@(href: Call, mode: Mode = Mode.Normal)(implicit messages: Messages)
```
SteveSugden: 👍

PatrykRudnicki: Do we need this default parameter? In practice probably always we're passing this value to the view

ed0906: Mode is only required on the journey pages, other pages e.g. Submissions/Drafts all have back buttons too and dont care about the mode

ed0906:
test this please

PR #461, opened 2019-08-27
CEDS-1533 new declaration bug

SteveSugden:


test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -44,6 +50,11 @@ class ChoiceControllerSpec extends ControllerSpec with ErrorHandlerMocks {
     authorizedUser()
   }
 
+  override def postRequest(body: JsValue): Request[AnyContentAsJson] =
```
PatrykRudnicki: This method is inside `ControllerSpec`

Path: `test/unit/base/ControllerSpec.scala`

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -82,41 +93,75 @@ class ChoiceControllerSpec extends ControllerSpec with ErrorHandlerMocks {
...
         created.status mustBe DeclarationStatus.DRAFT
         created.choice mustBe "SMP"
       }
 
-      "user chooses Standard Dec" in new SetUp {
+      "user chooses Supplementary Dec for existing Standard Dec" in new SetUp {
+        val existingDec = existingDeclaration(Choice.AllowedChoiceValues.StandardDec)
+        when(mockExportsCacheService.get(any())(any())).thenReturn(Future.successful(Some(existingDec)))
+        when(mockExportsCacheService.update(any())(any()))
+          .thenReturn(Future.successful(Some(existingDec)))
```
PatrykRudnicki: Those mocks are in `MockExportCacheService`

`withNewCaching(existingDec)`

app/controllers/ChoiceController.scala
```diff
@@ -89,4 +88,23 @@ class ChoiceController @Inject()(
       )
   }
 
+  private def updateChoice(id: String, choice: Choice)(implicit hc: HeaderCarrier) =
+    exportsCacheService.get(id).map(_.map(_.copy(choice = choice.value))).flatMap {
```
PatrykRudnicki: I'll prefer to create method in cacheService to retrieve this if necessary, `map -> map -> copy -> flatMap` it can be more readable when you will wrap this in a method

app/controllers/ChoiceController.scala
```diff
@@ -89,4 +88,23 @@ class ChoiceController @Inject()(
...
+  private def create(choice: Choice)(implicit hc: HeaderCarrier) =
+    exportsCacheService
+      .create(
+        ExportsDeclaration(
+          None,
+          DeclarationStatus.DRAFT,
+          createdDateTime = Instant.now,
+          updatedDateTime = Instant.now,
+          choice.value
+        )
```
PatrykRudnicki: Similar here, I'll add method to `ExportsDeclaration` which will take choice as argument and return this declaration, it's some kind of empty declaration so it can be useful in the future

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -23,18 +23,24 @@ import models.{DeclarationStatus, ExportsDeclaration}
...
+import utils.FakeRequestCSRFSupport._
 
 import scala.concurrent.Future
 
 class ChoiceControllerSpec extends ControllerSpec with ErrorHandlerMocks {
   import ChoiceControllerSpec._
 
-  private val existingDeclaration =
-    aDeclaration(withId("declarationId"), withChoice(Choice.AllowedChoiceValues.SupplementaryDec))
+  private def existingDeclaration(choice: String = Choice.AllowedChoiceValues.SupplementaryDec) =
```
PatrykRudnicki: I know it's old code, but can you change import from just Choice to `Choice.AllowedChoiceValues._`?

Usage will be shorter and more redable

test/unit/controllers/ChoiceControllerSpec.scala
```diff
@@ -82,41 +93,75 @@ class ChoiceControllerSpec extends ControllerSpec with ErrorHandlerMocks {
...
         created.id mustBe None
         created.status mustBe DeclarationStatus.DRAFT
         created.choice mustBe "STD"
       }
+
+      "user chooses Standard Dec for existing Supplementary Dec" in new SetUp {
+        val existingDec = existingDeclaration(Choice.AllowedChoiceValues.SupplementaryDec)
+        when(mockExportsCacheService.get(any())(any())).thenReturn(Future.successful(Some(existingDec)))
+        when(mockExportsCacheService.update(any())(any()))
+          .thenReturn(Future.successful(Some(existingDec)))
```
PatrykRudnicki: The same about mocks here

app/controllers/ChoiceController.scala
SteveSugden: Not sure about this.  More readable here perhaps, but the service method would(or could) look the same.    Took me ages to get the right combination of maps as well.

test/unit/controllers/ChoiceControllerSpec.scala
SteveSugden: Thanks.  I'll take a look

test/unit/controllers/ChoiceControllerSpec.scala
SteveSugden: I wanted a version that did not set a value for decId on the session.

PatrykRudnicki: Ah, okay, sorry.

What do you think about moving this to `ControllerSpec`? Can it be useful or it'll be just part of this spec?

What do you think about changing name of this method and use it without `override`?

SteveSugden: I could do either.   I don't think many other controllers would need it as they are mostly dealing with requests involving existing draft decs.   Maybe a more useful name would be a good idea.

ed0906: It's an exceptional circumstance so I'd just create the request manually, but upto you

app/controllers/ChoiceController.scala
SteveSugden: I agree that some sort of factory method to create a new draft dec of a given type would be useful.   But perhaps we could create a ticket to create a new `DecType` value and separate the valid values from 'Choice'.   Seems a bit odd that you can create a declaration from a `continue saved declaration`

test/unit/controllers/ChoiceControllerSpec.scala
SteveSugden: Just had a look.  `withNewCaching` mocks all three of the get/create/update methods.  I specifically want to test that update (not create) is called when dealing with an existing draft dec (and vise-versa for a new one).    "Blanket" mocking all three methods might have hidden a bug.  (Accept that I could use verify to check what mocked methods are actually called but a null-pointer tends to grab your attention)

SteveSugden: In fact, there are quite a few tests here that are needlessly mocking `mockExportsCacheService.get` (i.e. tests pass without mocking that method - as they should)

SteveSugden:
test this please

PR #462, opened 2019-08-27
CEDS-1549 Fix not removeble additional fiscal reference for item

bambuchaAdm:
Plus introducing ItemAction - would generalize flow for actions where we want to have item withId but such does not exist.

Introduce UnitViewSpec and rename old ViewSpec to AppViewSpec.

### DIFF
--- a/app/controllers/declaration/SealController.scala
+++ b/app/controllers/declaration/SealController.scala
@@ -82,9 +82,9 @@ class SealController @Inject()(
   private def badRequest(mode: Mode, formWithErrors: Form[Seal], cachedSeals: Seq[Seal])(
     implicit request: JourneyRequest[_]
   ) = {
-    val d = request.cacheModel.transportDetails
-    val s = request.cacheModel.seals
-    BadRequest(sealPage(mode, formWithErrors, s, d.fold(false)(_.container)))
+    val transportDetails = request.cacheModel.transportDetails
+    val seals = request.cacheModel.seals
+    BadRequest(sealPage(mode, formWithErrors, seals, transportDetails.fold(false)(_.container)))
   }
 
   private def removeSeal(
--- a/app/controllers/declaration/SealController.scala
+++ b/app/controllers/declaration/SealController.scala
@@ -70,7 +70,7 @@ class SealController @Inject()(
     implicit request: JourneyRequest[AnyContent]
   ): Future[Result] =
     saveAndContinue(boundForm, cachedSeals, isMandatory = false, elementLimit).fold(
-      formWithErrors => badRequest(mode, formWithErrors, cachedSeals),
+      formWithErrors => Future.successful(badRequest(mode, formWithErrors, cachedSeals)),
       updatedCache =>
         if (updatedCache != cachedSeals) {
           updateCache(updatedCache).map { _ =>
@@ -79,15 +79,10 @@ class SealController @Inject()(
         } else Future.successful(navigator.continueTo(routes.SummaryController.displayPage(Mode.Normal)))
     )
 
-  private def badRequest(mode: Mode, formWithErrors: Form[Seal], cachedSeals: Seq[Seal])(
-    implicit request: JourneyRequest[_]
-  ) = {
-    val declaration = exportsCacheService.get(request.declarationId)
-    declaration.map(_.flatMap(_.transportDetails)).flatMap { data =>
-      declaration.map(_.map(_.seals)).map { seals =>
-        BadRequest(sealPage(mode, formWithErrors, seals.getOrElse(Seq.empty), data.fold(false)(_.container)))
-      }
-    }
+  private def badRequest(mode: Mode, formWithErrors: Form[Seal], cachedSeals: Seq[Seal])(implicit request: JourneyRequest[_]) = {
+    val d = request.cacheModel.transportDetails
+    val s = request.cacheModel.seals
+    BadRequest(sealPage(mode, formWithErrors, s, d.fold(false)(_.container)))
   }
 
   private def removeSeal(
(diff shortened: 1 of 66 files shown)


app/controllers/declaration/SealController.scala
```diff
@@ -80,12 +80,9 @@ class SealController @Inject()(
     )
 
   private def badRequest(formWithErrors: Form[Seal], cachedSeals: Seq[Seal])(implicit request: JourneyRequest[_]) = {
-    val declaration = exportsCacheService.get(request.declarationId)
-    declaration.map(_.flatMap(_.transportDetails)).flatMap { data =>
-      declaration.map(_.map(_.seals)).map { seals =>
-        BadRequest(sealPage(formWithErrors, seals.getOrElse(Seq.empty), data.fold(false)(_.container)))
-      }
-    }
+    val d = request.cacheModel.transportDetails
```
PatrykRudnicki: I prefer to have better naming than `d` or `s`. I know the scope is small, but it'll be easier to read when you will have whole word

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -43,6 +43,38 @@
 
 @(itemId:String, form: Form[AdditionalFiscalReference], additionalReferences: Seq[AdditionalFiscalReference] = Seq.empty)(implicit request: Request[_], messages: Messages)
 
+@referenceTable = {
```
PatrykRudnicki: Why you're not using `removable_elements_table` from components? We should have one approach to show this part of the page

app/controllers/declaration/SealController.scala
bambuchaAdm: Fixed in later commit

app/views/declaration/additional_fiscal_references.scala.html
bambuchaAdm: Now it continas submits instead fo buttons - i will try think about unification

app/controllers/declaration/SealController.scala
PatrykRudnicki: Still `d` and `s`

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -43,26 +43,57 @@
...
+        </div>
+    }
+}
+
 @main_template(
     title = messages("declaration.additionalFiscalReferences.title")
 ) {
 
-    @components.back_link(routes.FiscalInformationController.displayPage(mode, itemId), mode)
+    @components.back_link(routes.FiscalInformationController.displayPage(mode, itemId))
```
ed0906: ?

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -43,26 +43,57 @@
...
-    @components.fields.removable_elements_table(
-        title = Some(messages("declaration.additionalFiscalReferences.numbers.header")),
-        elements = additionalReferences.map(_.asString).map(ref => LabelledValue(ref))
-    )
-
     @helper.form(routes.AdditionalFiscalReferencesController.saveReferences(mode, itemId), 'autoComplete -> "off") {
         @helper.CSRF.formField
 
+
+        @referenceTable
```
ed0906: Seems unnecessary to pull this out as it’s not reused

test/controllers/navigation/NavigatorTest.scala
```diff
@@ -46,14 +46,14 @@ class NavigatorTest extends WordSpec with Matchers with MockitoSugar with Export
...
-    def authenticatedRequest(action: Option[FormAction]) = AuthenticatedRequest[AnyContent](
+    def authenticatedRequest(action: Option[FormAction]) = new AuthenticatedRequest[AnyContent](
       FakeRequest("GET", "uri")
         .withFormUrlEncodedBody(action.getOrElse("other-field").toString -> "")
         .withSession(ExportsSessionKeys.declarationId -> "declarationId"),
       mock[SignedInUser]
     )
     def request(action: Option[FormAction]): JourneyRequest[AnyContent] =
-      JourneyRequest[AnyContent](authenticatedRequest(action), declaration)
+      new JourneyRequest[AnyContent](authenticatedRequest(action), declaration)
```
ed0906: Why are all these using new now?

bambuchaAdm: Now they inherited form them salve - this required to move out form case classes and as consequence `apply` on companion object is not generated

app/views/declaration/additional_fiscal_references.scala.html
bambuchaAdm: This is more about to keep high level point of view whole page - we have components earlier - and later - this lot of low level tags just breaks abstraction level here

This is why it why i extract those just as variable in same view

app/views/declaration/additional_fiscal_references.scala.html
bambuchaAdm: Probably by accident - moreover in `back_link` template `mode` seems not used at all
https://github.com/hmrc/customs-declare-exports-frontend/blob/CEDS-1549/app/views/components/back_link.scala.html#L21

ed0906: Yep but it will be and this will likely cause a bug as we will forget it was taken off

bambuchaAdm: Fixed - in this case maybe not putting default one would be much better ?

bambuchaAdm:
@ed0906 Jest to not stuck with this PR and issue - should i withdraw solution and propose minimal fix for bug ?

bambuchaAdm:
test this please

bambuchaAdm:
@MaciejRewera for export-frontend pr validator would fail if code wont be formatted properly

PR #471, opened 2019-08-29
CEDS-1490 Add trait Injector which provide injector with clearing SharedMetrics

PatrykRudnicki:


test/base/WithInjector.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package base
+
+import com.codahale.metrics.SharedMetricRegistries
+import play.api.inject.guice.GuiceApplicationBuilder
+
+trait WithInjector {
+
+  SharedMetricRegistries.clear()
```
ed0906: Maybe add a comment too explain why this is here

test/base/WithInjector.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  val injector = GuiceApplicationBuilder().injector()
```
ed0906: I think some helper methods could be useful here,

with `MockitoSugar` we get `mock[T]`, maybe a method that follows the same pattern e.g. `real[T]` or just `instanceOf[T]` so we can declutter the tests a little? What do you think?

test/connectors/CustomsDeclareExportsConnectorSpec.scala
```diff
@@ -39,12 +37,9 @@ import unit.base.UnitSpec
 
 import scala.concurrent.ExecutionContext.global
 
-class CustomsDeclareExportsConnectorSpec extends UnitSpec with ScalaFutures {
+class CustomsDeclareExportsConnectorSpec extends UnitSpec with ScalaFutures with WithInjector {
```
ed0906: `with WithInjector` seems a little strange, why not just `with Injector`?

Or if that has name clashes maybe `with InjectorSugar` or something?

test/base/WithInjector.scala
PatrykRudnicki: Done, changes in separate commit :)

PR #472, opened 2019-08-29
CEDS-1511 Fix Save & Resume Button Location & Summary Page Links

ed0906:
When in Draft mode, at the end of the journey you would be taken to the `/saved-summary`, similar in Amend mode.
If you click "Save and continue" on the final page of the journey you are now always taken to `/summary` (your mode reverts to Normal)


ed0906:
test this please

PR #473, opened 2019-08-29
CEDS-1552 Consistent way of using loggers

PatrykRudnicki:


app/controllers/ChoiceController.scala
```diff
@@ -45,7 +45,7 @@ class ChoiceController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
-  val logger = Logger.apply(this.getClass)
+  private val logger = Logger(this.getClass)
```
ed0906: 👍

ed0906:
Is that all there was :o

PatrykRudnicki:
We discussed this few months ago and change loggers to be used in this way like in this PR :)

PR #476, opened 2019-08-29
CEDS-1558 Mode in action refiner spike

bambuchaAdm:
Done only on single page. Working and pass first page

Props: Mode is hidden in request
Cons: Manual form generation - could be overcome with little our infrastructure.

We discuss with @ed0906  and there could be stay with `Mode` in `Controllers` signature and routes because they require it but pass it with request for view as meet in middle

### DIFF
--- a/app/controllers/ChoiceController.scala
+++ b/app/controllers/ChoiceController.scala
@@ -20,6 +20,7 @@ import java.time.Instant
 
 import controllers.actions.AuthAction
 import controllers.declaration.ModelCacheable
+import controllers.navigation.StandardDeclarationJorney
 import forms.Choice
 import forms.Choice.AllowedChoiceValues._
 import forms.Choice._
@@ -70,11 +71,11 @@ class ChoiceController @Inject()(
               request.declarationId match {
                 case Some(id) =>
                   updateChoice(id, choice).map { _ =>
-                    Redirect(controllers.declaration.routes.DispatchLocationController.displayPage(Mode.Normal))
+                    StandardDeclarationJorney.start
                   }
                 case _ =>
                   create(choice) map { created =>
-                    Redirect(controllers.declaration.routes.DispatchLocationController.displayPage(Mode.Normal))
+                    StandardDeclarationJorney.start
                       .addingToSession(ExportsSessionKeys.declarationId -> created.id.get)
                   }
               }
--- a/app/controllers/actions/JourneyAction.scala
+++ b/app/controllers/actions/JourneyAction.scala
@@ -17,6 +17,7 @@
 package controllers.actions
 
 import com.google.inject.Inject
+import models.Mode
 import models.requests.{AuthenticatedRequest, JourneyRequest}
 import play.api.Logger
 import play.api.mvc.{ActionRefiner, Result, Results}
@@ -36,13 +37,18 @@ class JourneyAction @Inject()(cacheService: ExportsCacheService)(
     implicit val hc: HeaderCarrier =
       HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
 
-    request.declarationId match {
-      case Some(id) =>
-        cacheService.get(id).map {
-          case Some(declaration) => Right(JourneyRequest(request, declaration))
-          case _                 => handleMissingDeclarationId(request)
+    Mode.binder.bind("mode", request.queryString) match {
+      case Some(Right(mode)) =>
+        request.declarationId match {
+          case Some(id) =>
+            cacheService.get(id).map {
+              case Some(declaration) => Right(JourneyRequest(request, declaration, mode))
+              case _                 => handleMissingDeclarationId(request)
+            }
+          case None => Future.successful(handleMissingDeclarationId(request))
         }
-      case None => Future.successful(handleMissingDeclarationId(request))
+
+      case Some(Left(_)) | None => Future.successful(Left(Results.BadRequest("mode query parameter is required")))
     }
   }
 
--- a/app/controllers/declaration/DispatchLocationController.scala
+++ b/app/controllers/declaration/DispatchLocationController.scala
@@ -42,14 +42,14 @@ class DispatchLocationController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
-  def displayPage(): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
     request.cacheModel.dispatchLocation match {
       case Some(data) => Ok(dispatchLocationPage(DispatchLocation.form().fill(data)))
       case _          => Ok(dispatchLocationPage(DispatchLocation.form()))
     }
   }
 
-  def submitForm(): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+  def submitForm(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
     DispatchLocation
       .form()
       .bindFromRequest()
--- a/app/controllers/declaration/DispatchLocationController.scala
+++ b/app/controllers/declaration/DispatchLocationController.scala
@@ -42,30 +42,29 @@ class DispatchLocationController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
-  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+  def displayPage(): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
     request.cacheModel.dispatchLocation match {
-      case Some(data) => Ok(dispatchLocationPage(mode, DispatchLocation.form().fill(data)))
-      case _          => Ok(dispatchLocationPage(mode, DispatchLocation.form()))
+      case Some(data) => Ok(dispatchLocationPage(DispatchLocation.form().fill(data)))
+      case _          => Ok(dispatchLocationPage(DispatchLocation.form()))
     }
   }
 
-  def submitForm(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+  def submitForm(): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
     DispatchLocation
       .form()
       .bindFromRequest()
       .fold(
-        (formWithErrors: Form[DispatchLocation]) =>
-          Future.successful(BadRequest(dispatchLocationPage(mode, formWithErrors))),
+        (formWithErrors: Form[DispatchLocation]) => Future.successful(BadRequest(dispatchLocationPage(formWithErrors))),
         validDispatchLocation =>
           updateCache(validDispatchLocation)
-            .map(_ => navigator.continueTo(nextPage(mode, validDispatchLocation)))
+            .map(_ => navigator.continueTo(nextPage(validDispatchLocation)))
       )
   }
 
-  private def nextPage(mode: Mode, providedDispatchLocation: DispatchLocation): Call =
+  private def nextPage(providedDispatchLocation: DispatchLocation)(implicit request: JourneyRequest[_]): Call =
     providedDispatchLocation.dispatchLocation match {
       case AllowedDispatchLocations.OutsideEU =>
-        controllers.declaration.routes.AdditionalDeclarationTypeController.displayPage(mode)
+        controllers.declaration.routes.AdditionalDeclarationTypeController.displayPage(request.mode)
       case AllowedDispatchLocations.SpecialFiscalTerritory =>
         controllers.declaration.routes.NotEligibleController.displayPage()
     }
--- a/app/controllers/navigation/StandardDeclarationJorney.scala
+++ b/app/controllers/navigation/StandardDeclarationJorney.scala
@@ -14,15 +14,16 @@
  * limitations under the License.
  */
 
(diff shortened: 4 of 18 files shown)


app/controllers/actions/JourneyAction.scala
```diff
@@ -36,13 +37,18 @@ class JourneyAction @Inject()(cacheService: ExportsCacheService)(
...
-          case _                 => handleMissingDeclarationId(request)
+    Mode.binder.bind("mode", request.queryString) match {
+      case Some(Right(mode)) =>
+        request.declarationId match {
+          case Some(id) =>
+            cacheService.get(id).map {
+              case Some(declaration) => Right(JourneyRequest(request, declaration, mode))
+              case _                 => handleMissingDeclarationId(request)
+            }
+          case None => Future.successful(handleMissingDeclarationId(request))
```
PatrykRudnicki: I'll at least combine those two options with `handleMissingDeclarationId` to one, but it's old code

app/controllers/navigation/StandardDeclarationJorney.scala
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package controllers.navigation
+
+import models.Mode
+import play.api.mvc.Result
+
+import play.api.mvc.Results._
+
+object StandardDeclarationJorney {
```
PatrykRudnicki: I'm just worried about this object, the size of it when we move here whole journeys, some duplications between Supplementary and Standard

PatrykRudnicki:
> We discuss with ed and there could be stay with Mode in Controllers signature and routes because they require it but pass it with request for view

but in practice you removed it, so what is the approach, with it or without?

app/controllers/navigation/StandardDeclarationJorney.scala
bambuchaAdm: This for now is palceholder which hides manual parameter building - I don't know which would i like more - put all logic here and having it spread over views and controllers

app/controllers/actions/JourneyAction.scala
bambuchaAdm: You know that one is before database hit and second is after ?

bambuchaAdm:
> but in practice you removed it, so what is the approach, with it or without?
This Spike is example for initial approch with action refiner not applied yet @ed0906 ideas. I can create another one but i would like to take response from team.

app/controllers/actions/JourneyAction.scala
PatrykRudnicki: yeah, you're right

app/views/declaration/additionaldeclarationtype/declaration_type.scala.html
```diff
@@ -27,9 +27,10 @@
     title = messages("declaration.declarationType.title")
 ) {
 
-    @components.back_link(controllers.declaration.routes.DispatchLocationController.displayPage(mode), mode)
+    @components.back_link(controllers.declaration.routes.DispatchLocationController.displayPage(), request.mode)
```
bambuchaAdm: if back_link would take `(mode: Mode) => call` this would even simplify to
``` 
@components.back_link(controllers.declaration.routes.DispatchLocationController.displayPage)
```
Even having mode in controller signature
Even if controller would have more then one parameter 
```
@components.back_link(MuplitpleParamterPath.displayPage(_, otherParamter)) 
```
still look good

app/controllers/navigation/StandardDeclarationJorney.scala
ed0906: Same, I don't think we need this object. It is likely to grow large and uses static method references which cant be mocked out in tests

app/views/declaration/dispatch_location.scala.html
```diff
@@ -18,26 +18,30 @@
...
-            @components.back_link(controllers.routes.ChoiceController.displayPage(), mode)
+            @components.back_link(controllers.routes.ChoiceController.displayPage(), request.mode)
         }
     }
 
-  @helper.form(controllers.declaration.routes.DispatchLocationController.submitForm(mode), 'autoComplete -> "off") {
-      @helper.CSRF.formField
+  @defining(controllers.declaration.routes.DispatchLocationController.submitForm()){ call =>
+
+  <form method="@call.method" action="@call.url?mode=@request.mode.name" autocomplete="off">
```
ed0906: Not keen on building URLs like this in  HTML

I much prefer the existing approach of using Play Routes to do it

conf/declaration.routes
```diff
@@ -1,8 +1,8 @@
 # Declaration type
 
-GET         /dispatch-location                   controllers.declaration.DispatchLocationController.displayPage(mode: models.Mode ?= models.Mode.Normal)
+GET         /dispatch-location                   controllers.declaration.DispatchLocationController.displayPage()
 
-POST        /dispatch-location                   controllers.declaration.DispatchLocationController.submitForm(mode: models.Mode ?= models.Mode.Normal)
+POST        /dispatch-location                   controllers.declaration.DispatchLocationController.submitForm()
```
ed0906: I prefer being forced to think about which Mode the page needs to be in, it is easy to create bugs without this

app/views/declaration/summary/summary_page.scala.html
```diff
@@ -131,6 +131,6 @@
     @if(mode == Mode.Draft) {
         <p class="mt-5">@messages("saved.declarations.summary.paragraph")</p>
 
-        <a class="button mt-3" href="@routes.DispatchLocationController.displayPage(mode)">@messages("saved.declarations.summary.continue")</a>
+        <a class="button mt-3" href="@routes.DispatchLocationController.displayPage()?mode=@mode.name">@messages("saved.declarations.summary.continue")</a>
```
ed0906: As mentioned above, not keen on having this query param everywhere. What is the route for display page already had query params?

app/controllers/declaration/DispatchLocationController.scala
```diff
@@ -42,30 +42,29 @@ class DispatchLocationController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
-  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+  def displayPage(): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
```
ed0906: I don't mind having the mode on the `request`, but I still think it needs to be here so we can construct the URLs using Play routing. It will be too easy to miss it out otherwise.

ed0906:
Apologies but I'm not keen on this. The `mode` is an important part of the journey that we should be forced to consider.
- It is important for all `Redirect`s In both controllers & views
- It is important for rendering Back Links in views

If we take it out of the controller method signature it will be easy to forget about and will cause bugs in every place it is missed.

I don't mind making it a field on the `JourneyRequest` *as well*, but not *instead* of the controller argument. This would mean we wouldn't need all the manual URL building I see in the views which needs additional tests.

app/controllers/ChoiceController.scala
```diff
@@ -70,11 +71,11 @@ class ChoiceController @Inject()(
               request.declarationId match {
                 case Some(id) =>
                   updateChoice(id, choice).map { _ =>
-                    Redirect(controllers.declaration.routes.DispatchLocationController.displayPage(Mode.Normal))
+                    StandardDeclarationJorney.start
```
PauloPortugal: :+1:

app/controllers/actions/JourneyAction.scala
```diff
@@ -36,13 +37,18 @@ class JourneyAction @Inject()(cacheService: ExportsCacheService)(
     implicit val hc: HeaderCarrier =
       HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
 
-    request.declarationId match {
-      case Some(id) =>
-        cacheService.get(id).map {
-          case Some(declaration) => Right(JourneyRequest(request, declaration))
-          case _                 => handleMissingDeclarationId(request)
+    Mode.binder.bind("mode", request.queryString) match {
```
PauloPortugal: :+1:

app/controllers/declaration/DispatchLocationController.scala
PauloPortugal: :+1:

app/views/declaration/additionaldeclarationtype/declaration_type.scala.html
PauloPortugal: :+1:

app/views/declaration/additionaldeclarationtype/declaration_type.scala.html
```diff
@@ -27,9 +27,10 @@
...
 
-    @helper.form(controllers.declaration.routes.AdditionalDeclarationTypeController.submitForm(mode), 'autoComplete -> "off") {
+    @defining(controllers.declaration.routes.AdditionalDeclarationTypeController.submitForm()){ call =>
```
PauloPortugal: This is the thing that puts me off, this would be a /POST with a query parameter, which is a bit weird.

app/views/declaration/dispatch_location.scala.html
```diff
@@ -18,26 +18,30 @@
 @import forms.declaration.DispatchLocation.AllowedDispatchLocations._
 @import uk.gov.hmrc.play.views.html._
 @import views.components.inputs.RadioOption
+@import models.requests.JourneyRequest
 
 @this(main_template: views.html.main_template)
 
-@(mode: Mode, form: Form[DispatchLocation])(implicit request: Request[_], messages: Messages)
+@(form: Form[DispatchLocation])(implicit request: JourneyRequest[_], messages: Messages)
```
PauloPortugal: :+1:

app/views/declaration/summary/summary_page.scala.html
```diff
@@ -112,7 +112,7 @@
     @items_section(suppDecData.items)
 
     @if(mode == Mode.Amend) {
-        <a class="button mt-3" href="@routes.DispatchLocationController.displayPage(mode)">@messages("site.continue")</a>
+        <a class="button mt-3" href="@routes.DispatchLocationController.displayPage()?mode=@mode.name">@messages("site.continue")</a>
```
PauloPortugal: This is Okish

conf/declaration.routes
```diff
@@ -1,8 +1,8 @@
 # Declaration type
 
-GET         /dispatch-location                   controllers.declaration.DispatchLocationController.displayPage(mode: models.Mode ?= models.Mode.Normal)
+GET         /dispatch-location                   controllers.declaration.DispatchLocationController.displayPage()
```
PauloPortugal: simpler routes definition :+1:

app/controllers/ChoiceController.scala
ed0906: If we are pulling this code out I think it should go into the Navigator

app/views/declaration/additionaldeclarationtype/declaration_type.scala.html
ed0906: I think the existing solution already does that

I think we just have to bare in mind that this is a UI and not a REST API, so the same rules of convention don’t necessarily apply

conf/declaration.routes
ed0906: I think this is going to make it easy to forget that `mode` needs to be considered 👎

conf/declaration.routes
bambuchaAdm: Fixed - now mode backs to constroller signature

app/views/declaration/summary/summary_page.scala.html
bambuchaAdm: Sorry i miss this again - i think mode should not have default value. We can remove this from router - it is already done on level of binder.

app/views/components/declaration_form.scala.html
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import models.requests.JourneyRequest
+@(callFactory: Mode => Call)(formBody: Html)(implicit request: JourneyRequest[_], messages: Messages)
+
+@helper.form(callFactory(request.mode), 'autoComplete -> "off"){
```
PatrykRudnicki: What's this? Some kind of the wrapper for all forms?

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -38,16 +38,17 @@
...
 
-@(mode: Mode, itemId:String, form: Form[AdditionalFiscalReference], additionalReferences: Seq[AdditionalFiscalReference] = Seq.empty)(implicit request: Request[_], messages: Messages)
+@(mode: Mode, itemId:String, form: Form[AdditionalFiscalReference], additionalReferences: Seq[AdditionalFiscalReference] = Seq.empty)(implicit request: JourneyRequest[_], messages: Messages)
 
 @main_template(
     title = messages("declaration.additionalFiscalReferences.title")
 ) {
 
-    @components.back_link(routes.FiscalInformationController.displayPage(mode, itemId), mode)
+    @components.journey_back(routes.FiscalInformationController.displayPage(_, itemId))
```
PatrykRudnicki: This underscore is as mode?

bambuchaAdm: journey_back - takes `Model => Call` function - here we use currying to have partial applied function

PatrykRudnicki:
What about this draft? Can we close it @bambuchaAdm ?

PatrykRudnicki:
Closing... Old PR probably not worth to keep it open

PR #481, opened 2019-08-29
CEDS-1511 Fix Summary Page Back Link

ed0906:


app/views/declaration/summary/summary_page.scala.html
```diff
@@ -32,20 +32,22 @@
...
-                    components.back_link(controllers.declaration.routes.TransportDetailsController.displayPage(mode), mode)
-            }
-        }
+  @mode match {
+      case Mode.Draft => {
+          @components.back_link(controllers.routes.SavedDeclarationsController.displayDeclarations(), mode)
+      }
+      case Mode.Amend => {
+          @components.back_link(controllers.routes.SavedDeclarationsController.displayDeclarations(), mode)
+      }
```
SteveSugden: Aren't those two cases doing the same thing?

test/unit/tools/Stubs.scala
```diff
@@ -64,6 +64,7 @@ trait Stubs {
       |metrics.showSamples=false
       |metrics.jvm=false
       |metrics.logback=false
+      |draft.timeToLive=1d
```
SteveSugden: What;s this change for?

app/views/declaration/summary/summary_page.scala.html
ed0906: Good spot, think that was a copy and paste error

test/unit/tools/Stubs.scala
ed0906: To render the summary page for a draft we need that adding into the config apparently,
otherwise it breaks

ed0906:
Turns out we cant fix the back link behaviour for Amends until we change the Amend Logic from "copy the rejected dec, then edit" to "edit the rejected dec" as we would need to know the ID of the rejected declaration to be able to go back to it...

SteveSugden:
So does this PR still need to be merged?  I'm struggling to see any change in the view logic

ed0906:
> So does this PR still need to be merged? I'm struggling to see any change in the view logic

Well, there was some missing tests that exist now so it's probably still worth merging

PR #487, opened 2019-08-30
CEDS-1567 Stop using app on viewspecs

PauloPortugal:
This will speed up our view specs and our tests will be prepared for
internationalisation as we will be testing that the message keys are
present in the templates.


PatrykRudnicki:
One scala file is not formatted, this is the reason of build failure: `ProcedureCodesViewSpec.scala`

PR #489, opened 2019-09-02
CEDS-1570 Remove usage of deprecated methods from ViewValidator

PatrykRudnicki:


test/views/declaration/CommodityMeasureViewSpec.scala
```diff
@@ -93,9 +93,9 @@ class CommodityMeasureViewSpec extends AppViewSpec with CommodityMeasureMessages
 
       val view = createView(CommodityMeasure.form().fillAndValidate(CommodityMeasure(Some(""), "", "")))
 
-      checkErrorsSummary(view)
-      checkErrorLink(view, 1, netMassEmpty, "#netMass")
-      checkErrorLink(view, 2, grossMassEmpty, "#grossMass")
+      view.getElementById("error-summary-heading").text() mustNot be(empty)
+      view.getElementById("netMass-error").attr("href") mustBe "#netMass"
```
bambuchaAdm: you have matcher for this in `ViewMatchers` - `hasError` or something like this

test/views/declaration/CommodityMeasureViewSpec.scala
```diff
@@ -105,8 +105,8 @@ class CommodityMeasureViewSpec extends AppViewSpec with CommodityMeasureMessages
 
       val view = createView(CommodityMeasure.form().fillAndValidate(CommodityMeasure(Some("99.123"), "", "")))
 
-      checkErrorsSummary(view)
-      checkErrorLink(view, 1, supplementaryUnitsError, "#supplementaryUnits")
+      view.getElementById("error-summary-heading").text() mustNot be(empty)
+      view.getElementById("supplementaryUnits-error").attr("href") mustBe "#supplementaryUnits"
```
bambuchaAdm: `haveHref`

test/views/cancellation/CancelDeclarationViewSpec.scala
```diff
@@ -53,7 +53,7 @@ class CancelDeclarationViewSpec extends AppViewSpec with CommonMessages {
         )
     )
 
-    checkErrorsSummary(view)
+    view.getElementById("error-summary-heading").text() mustNot be(empty)
```
MaciejRewera: Is it enough to just test that it is not empty? It confuses me always on view tests, to what extent do we want our frontend to be tested?

test/views/declaration/CommodityMeasureViewSpec.scala
```diff
@@ -115,8 +115,8 @@ class CommodityMeasureViewSpec extends AppViewSpec with CommodityMeasureMessages
 
       val view = createView(CommodityMeasure.form().fillAndValidate(CommodityMeasure(Some("99.99"), "", "10.00")))
 
-      checkErrorsSummary(view)
-      checkErrorLink(view, 1, netMassEmpty, "#netMass")
+      view.getElementById("error-summary-heading").text() mustNot be(empty)
+      view.getElementById("netMass-error").attr("href") mustBe "#netMass"
```
bambuchaAdm: same here - `haveHref` but i think it would be nice to add here also something `haveHrefToAnchor` with additional check to if document contains specified id

test/base/ViewValidator.scala
```diff
@@ -73,17 +32,6 @@ trait ViewValidator extends MustMatchers with ViewMatchers {
     errorLink.attr("href") must be(href)
   }
 
-  @deprecated("Please use 'page must haveFieldErrorLink(...)'", since = "2019-08-07")
```
MaciejRewera: Do we have methods like `haveFieldErrorLink` defined already or is it part of the future works?

test/views/cancellation/CancelDeclarationViewSpec.scala
SteveSugden: I'm confused.  This change seems quite repetitive.  Why not change `checkErrorSummary` to do the same thing as above?    Or am I missing something?

bambuchaAdm: Problem was in way how scala resolved implicits - it resolve earlier `MustMatchers` implicit defs rather then which comes form `ViewMatchers` and cause errors in usage. Hopefully problem was solved.

test/base/ViewValidator.scala
bambuchaAdm: Yes we have but in different trait - `ViewMatchers`

PatrykRudnicki:
Thanks guys for the comments, yeah, I didn't know about implicit conversion to Document in ViewMatchers, so I'll be using methods from ViewMatchers. I based this solution on few changes in Ed's PR, but it was wrong, good to make some kind of the draft :)

PatrykRudnicki:
Closing PR... I'll raise another one

PR #492, opened 2019-09-03
CEDS-1567 Stop using app on viewspecs

PauloPortugal:
This will speed up our view specs and our tests will be prepared for
internationalisation as we will be testing that the message keys are
present in the templates.


PauloPortugal:
closing PR due to https://github.com/hmrc/customs-declare-exports-frontend/pull/497/files

PR #495, opened 2019-09-04
BAU - quick POC for custom mapper for required radio fields

SteveSugden:


PatrykRudnicki:
Spec for this model failed, is it mean that it's different behaviour?

SteveSugden:
TransportDetailsSpec passes now that I've included "verify isContainedIn"

Should I work this up into a proper PR?    We can replace the Option[String] fields with String as Patryk suggested.

PatrykRudnicki:
I think the solution is good :)

PR #503, opened 2019-09-06
CEDS-1608 - add legal declaration questions to the declaration submit…

SteveSugden:
… page


app/utils/validators/forms/FieldValidator.scala
```diff
@@ -53,6 +53,8 @@ object FieldValidator {
 
   val nonEmpty: String => Boolean = (input: String) => input.trim.nonEmpty
 
+  val isTrue: Boolean => Boolean = (input: Boolean) => input.equals(true)
```
PatrykRudnicki: Hmm, `input.equals(true)` => if true return true, if false return false, so can you just return `input`? There is no point to check if it's equal to `true`

test/forms/declaration/LegalDeclarationSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+ * limitations under the License.
+ */
+
+package forms.declaration
+
+import helpers.views.declaration.WarehouseIdentificationMessages
+import play.api.libs.json._
+import unit.base.UnitSpec
+
+class LegalDeclarationSpec extends UnitSpec with WarehouseIdentificationMessages {
```
PatrykRudnicki: Why are you extending `WarehouseIdentificationMessages`?

PatrykRudnicki:
Looks good, will approve it after answers :)

app/utils/validators/forms/FieldValidator.scala
SteveSugden: That's true.  Readability?  :-)

test/forms/declaration/LegalDeclarationSpec.scala
SteveSugden: Because I forgot to delete from the Spec I copied to create this one.

PR #504, opened 2019-09-09
CEDS-1506 Add Source ID & Fix Back Logic for Amending a Declaration

ed0906:


app/controllers/SubmissionsController.scala
```diff
@@ -51,8 +52,15 @@ class SubmissionsController @Inject()(
   def amend(id: String): Action[AnyContent] = authenticate.async { implicit request =>
     customsDeclareExportsConnector.findDeclaration(id) flatMap {
       case Some(declaration) =>
+        val currentDateTime = Instant.now()
+        val newDeclaration = declaration.copy(
```
bambuchaAdm: can we move this logic to companion object of declaration ?

ed0906: done

test/unit/controllers/SubmissionsControllerSpec.scala
```diff
@@ -105,6 +117,17 @@ class SubmissionsControllerSpec extends ControllerSpec {
         redirectLocation(result) must be(Some(controllers.routes.SubmissionsController.displayListOfSubmissions().url))
       }
 
+      def inTheLast(timeout: Timeout): BeMatcher[Instant] = new BeMatcher[Instant] {
```
SteveSugden: Should we not be in control of time so that we can avoid potentially assertions like this?  I though AppConfig had a clock defined, but apparently not.

ed0906: Good point, I'll make it unit testable with a fixed clock

PR #510, opened 2019-09-13
CEDS-1651 Remove /v2 prefix from REST endpoints

PauloPortugal:
As part of the mapping story we had to create a second version of the
endpoints, to consume a JSON payload as opposed to a XML payload,
therefore every single conflicting endpoint ended up being prefixed
by /v2.
This task is to remove this prefix.

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -79,7 +79,7 @@ microservice {
       host = localhost
       port = 6792
       submit-declaration = "/declaration"
-      declaration-v2 = "/v2/declarations"
+      declarations = "/declarations"
       fetch-notifications = "/notifications"
       fetch-submission-notifications = "/submission-notifications"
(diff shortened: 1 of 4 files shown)


conf/application.conf
```diff
@@ -79,7 +79,7 @@ microservice {
       host = localhost
       port = 6792
       submit-declaration = "/declaration"
-      declaration-v2 = "/v2/declarations"
+      declarations = "/declarations"CustomsDeclareExportsConnectorIntegrationSpec
```
PatrykRudnicki: You paste name of test in this line

PauloPortugal: :man_facepalming:

PauloPortugal: This has been addressed! :)

PR #512, opened 2019-09-16
CEDS-1611 Audit declaration cancellation events

PauloPortugal:


app/controllers/CancelDeclarationController.scala
```diff
@@ -54,13 +63,31 @@ class CancelDeclarationController @Inject()(
...
         form => {
+          auditService.auditAllPagesUserInput(Json.toJson(form).as[JsObject])
           val context = exportsMetrics.startTimer(cancelMetric)
-          customsDeclareExportsConnector.createCancellation(form).map { _ =>
-            exportsMetrics.incrementCounter(cancelMetric)
-            context.stop()
-            Ok(cancelConfirmationPage())
-          }
+          for {
+            _ <- customsDeclareExportsConnector.createCancellation(form) andThen {
```
bambuchaAdm: Are we able to move this from for-comprehension ?

bambuchaAdm: It would looks like
```
customsDeclareExportsConnector.createCancellation(form).andThen {
  case Failure(exception) =>
    logger.error(s"Error response from backend $exception")
    auditService.audit(AuditTypes.Cancellation, auditData(form, Failure.toString))
  casee Success(_) => 
    auditService.audit(AuditTypes.Cancellation, auditData(form, Success.toString))
    exportsMetrics.incrementCounter(cancelMetric)
    context.stop()
}.map(_ => Ok(cancelConfirmationPage())
```

PauloPortugal: This has been addressed.

app/controllers/CancelDeclarationController.scala
```diff
@@ -54,13 +63,30 @@ class CancelDeclarationController @Inject()(
...
+            case Success(_) =>
+              auditService.audit(AuditTypes.Cancellation, auditData(form, Success.toString))
+              exportsMetrics.incrementCounter(cancelMetric)
+              context.stop()
+          } map (_ => Ok(cancelConfirmationPage()))
         }
       )
   }
+
+  private def auditData(form: CancelDeclaration, result: String)(
```
ed0906: We could move this kind of method to the audit service potentially

app/controllers/CancelDeclarationController.scala
```diff
@@ -54,13 +63,30 @@ class CancelDeclarationController @Inject()(
         (formWithErrors: Form[CancelDeclaration]) =>
           Future.successful(BadRequest(cancelDeclarationPage(formWithErrors))),
         form => {
+          auditService.auditAllPagesUserInput(Json.toJson(form).as[JsObject])
```
ed0906: Could change this method to accept a generic argument and an implicit Writes to hide some of the json boilerplate

PauloPortugal: Sure. I would like to do this with you.

PR #513, opened 2019-09-17
CEDS-1657 DMS Rejected errors and support for it

PatrykRudnicki:


app/services/DmsRejErrors.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import com.github.tototoshi.csv._
+import services.model.DmsRejError
+
+object DmsRejErrors {
+
+  private val errors: List[DmsRejError] = {
+    val reader = CSVReader.open(new File("conf/code-lists/errors-dms-rej-list.csv"))
+
+    val errors: List[List[String]] = reader.all()
```
bambuchaAdm: why shaddowing here ?

app/services/model/DmsRejError.scala
```diff
@@ -0,0 +1,36 @@
...
+
+case class DmsRejError(code: String, description: String)
+
+object DmsRejError {
+
+  private val logger = Logger(this.getClass)
+
+  implicit val format = Json.format[DmsRejError]
+
+  def apply(list: List[String]): DmsRejError = list match {
```
bambuchaAdm: Try/Option - ?

app/services/DmsRejErrors.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package services
+
+import java.io.File
+
+import com.github.tototoshi.csv._
+import services.model.DmsRejError
+
+object DmsRejErrors {
```
bambuchaAdm: What is `Rej` ?

app/services/DmsRejErrors.scala
PatrykRudnicki: Mistake, fixed

app/services/DmsRejErrors.scala
PatrykRudnicki: Those errors are `DMSREJ`, changed to better name

app/services/model/DmsRejError.scala
PatrykRudnicki: It should fail at runtime, I'm not sure that there is a value of using try or option here.

List is prepared by us, so if something is wrong it must fail

app/services/DmsRejectedErrors.scala
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package services
+
+import java.io.File
+
+import com.github.tototoshi.csv._
+import services.model.DmsRejectedError
+
+object DmsRejectedErrors {
```
bambuchaAdm: Maybe move to componanion - would be in one place

PatrykRudnicki: Approach from different code list in our application. I can move it there. Let's first wait for the decision if we can use this library

conf/code-lists/errors-dms-rej-list.csv
```diff
@@ -0,0 +1,142 @@
...
+CDS11006,Relation error: The NetMass per item or the sum of all in the declaration is larger than the gross mass.
+CDS11007,Relation error: The submitter in the request/supplementary declaration differs from the related declaration.
+CDS11012,Relation error: element is not allowed to be amended through this additional message type.
+CDS12003,Uniqueness error: Reuse of submitter reference is not (yet) allowed.
+CDS12005,Authorization error: Party ID unknown or invalid.
+CDS12006,Authorization warning: invalid import/export license number.
+CDS12007,Authorization error: authorization not valid for declaration.
+CDS12008,Authorization error: requested amount or quantity is not available for authorization reservation.
+CDS12014,"Sequence error: Sequence does not comply with one of the following conditions:
+- A Goods Presentation Notification is only permitted after an advance declaration or an advance incomplete declaration.
```
ed0906: How does the CSV Reader handle new lines like this?

app/services/model/DmsRejectedError.scala
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services.model
+
+import play.api.Logger
+import play.api.libs.json.Json
+
+case class DmsRejectedError(code: String, description: String)
```
ed0906: Any chance we could call it something a bit more generic that doesn't imply its tied to a downstream service?

e.g. `RejectionReason`, `RejectionError`

test/unit/services/DmsRejectedErrorsSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import services.DmsRejectedErrors
+import services.model.DmsRejectedError
+import unit.base.UnitSpec
+
+class DmsRejectedErrorsSpec extends UnitSpec {
+
+  import DmsRejectedErrors._
+
+  "Dms Rejected Errors" should {
```
ed0906: I'd maybe have a test to double check those multi line messages are read in properly

conf/code-lists/errors-dms-rej-list.csv
PatrykRudnicki: In the same way like you can see here, so when you retrieve this you will receive the same format and everything in string

ed0906: As long as it works :D

app/services/DmsRejectedErrors.scala
PatrykRudnicki: Extracted

test/unit/services/model/RejectionReasonSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    "correctly read multiline values" in {
+
+      val expectedMessages =
+        """Sequence error: The referred declaration does not comply with one of the following conditions:
+          |- The AdditionalMessage.declarationReference must refer to an existing declaration (Declaration.reference),
+          |- have been accepted,
+          |- not be invalidated.""".stripMargin
+      val expectedRejectionReason = RejectionReason("CDS12015", expectedMessages)
+
+      allRejectedErrors must contain(expectedRejectionReason)
```
ed0906: 👍

conf/code-lists/errors-dms-rej-list.csv
PatrykRudnicki: Added test for this case

app/services/model/DmsRejectedError.scala
PatrykRudnicki: Changed

test/unit/services/DmsRejectedErrorsSpec.scala
PatrykRudnicki: Added test

PR #514, opened 2019-09-18
CEDS-1656 Rejected notification errors page

PatrykRudnicki:


app/services/model/RejectionReason.scala
```diff
@@ -44,4 +46,22 @@ object RejectionReason {
 
     errors.map(RejectionReason.apply _)
   }
+
+  def getErrorDescription(errorCode: String): String =
+    allRejectedErrors.find(_.code == errorCode).map(_.description).getOrElse("Unknown error")
+
+  def fromNotifications(notifications: Seq[Notification]): Seq[RejectionReason] = {
+
+    val rejectionNotification = notifications.filter(_.functionCode == SubmissionStatus.Rejected.fullCode).headOption
```
bambuchaAdm: find

app/views/submissions.scala.html
```diff
@@ -46,7 +46,13 @@ <h1>@messages("submissions.title")</h1>
 
         @for( (submission, notifications) <- submissions ){
             <tr class="table-row">
-                <td class="table-cell">@submission.ducr</td>
+                <td class="table-cell">
+                    @if(notifications.map(_.functionCode).contains(SubmissionStatus.Rejected.fullCode)) {
```
bambuchaAdm:
```
notifications.map(_.functionCode).contains(SubmissionStatus.Rejected.fullCode)
```
Can we hide this structural detail under some object method ?

app/services/model/RejectionReason.scala
PatrykRudnicki: I want to avoid having Option, but I used `headOption` 🤦‍♂ Fixed

app/views/submissions.scala.html
PatrykRudnicki: Yes, done

PR #515, opened 2019-09-18
CEDS-1661 - some content changes

SteveSugden:
Mostly simple changes as outlined in the Google doc attached to the ticket.  Only real "code" changes are the ability to render a list of EU countries and related territories and the ability to have html content as part of a "hint" (in this case a link to an eu web-site).


conf/code-lists/eu-special-fiscal-territories.csv
```diff
@@ -0,0 +1,14 @@
...
+Guernsey
+Guadeloupe
+Jersey
+Martinique
+Mayotte
+Mount Athos
+Réunion
+Saint Martin
+San Marino
+Turkey
```
PauloPortugal: Turkey???? Really!

test/services/CountriesSpec.scala
```diff
@@ -16,19 +16,33 @@
...
 package services
 
-import services.Countries.allCountries
+import org.scalatest.{Matchers, WordSpec}
+import services.Countries._
 import services.model.Country
-import unit.base.UnitSpec
 
-class CountriesSpec extends UnitSpec {
+class CountriesSpec extends WordSpec with Matchers {
```
bambuchaAdm: Why not unit ?

app/views/declaration/goods_measure.scala.html
```diff
@@ -44,16 +44,14 @@
         )
 
         @components.fields.field_text(
-          field = form("netMass"),
-          label = messages("supplementary.commodityMeasure.netMass"),
-          hint = Some(messages("supplementary.commodityMeasure.netMass.hint")),
-          labelClass = Some("bold-small")
+            field = form("grossMass"),
+            label = messages("supplementary.commodityMeasure.grossMass"),
```
bambuchaAdm: why flipped ?

app/views/declaration/dispatch_location.scala.html
```diff
@@ -62,6 +63,24 @@
       )
     )
 
+      <details class="govuk-details" role="group">
```
bambuchaAdm: Do we have test for this section ?

conf/code-lists/eu-special-fiscal-territories.csv
SteveSugden: I obtained the contents of the list(s) from the ticket (and double checked with James W)

test/services/CountriesSpec.scala
SteveSugden: I was experimenting with Matchers.  Have now discovered that you can replace 'should' with 'must' (didn't seem to work before) and will therefore revert to UnitSpec

app/views/declaration/dispatch_location.scala.html
SteveSugden: What would you test?  It's just a block of static content and I've tested that the lists of countries are correctly produced from the csv files.    There aren't any tests for the details/summary section in fiscal_information either.

One thing worth testing is that the details hide/reveal works in IE/Edge because from memory I don't think this standard html 5 tag is supported in these browsers without some javascript trickery.

test/services/CountriesSpec.scala
PatrykRudnicki: `Matchers -> shouldBe`, `MustMatchers -> mustBe`, but if you're talking about description:

`"A Controller" should/must` both, should and must is supported with WordSpec

PR #518, opened 2019-09-19
CEDS-1676 - Back link for declaration 'error list view'

SteveSugden:
- change the url and text for the "Back" button when on summary screen when amending a rejected declaration.

### DIFF
--- a/app/views/declaration/summary/summary_page.scala.html
+++ b/app/views/declaration/summary/summary_page.scala.html
@@ -36,7 +36,7 @@
 
   @mode match {
       case Mode.Amend if request.cacheModel.sourceId.isDefined => {
-          @components.back_link(controllers.routes.NotificationsController.listOfNotificationsForSubmission(request.cacheModel.sourceId.get), mode)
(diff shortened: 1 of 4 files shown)


app/views/declaration/summary/summary_page.scala.html
```diff
@@ -36,7 +36,7 @@
 
   @mode match {
       case Mode.Amend if request.cacheModel.sourceId.isDefined => {
-          @components.back_link(controllers.routes.NotificationsController.listOfNotificationsForSubmission(request.cacheModel.sourceId.get), mode)
+          @components.back_link(controllers.routes.SubmissionsController.displayListOfSubmissions(), mode, "supplementary.summary.back")
```
ed0906: Looks like we don't need `sourceId` any more, I think the only reason we have that field was for this link

SteveSugden: I see what you mean now.  I thought it was just not needed for this case test; but it's now not referenced at all.

I'm going to leave it as part of the model, however.  This change to "back" functionality might not be the final design - depends on user feedback.  Plus, I think that at some point we will need to know the "heritage" of an amended declaration.

PR #520, opened 2019-09-20
CEDS-1691 - notification field rename

SteveSugden:
This PR requires both of the changes to the BE - i.e. the new `notification.actionId` and `action.id` fields

### DIFF
--- a/app/models/declaration/submissions/SubmissionStatus.scala
+++ b/app/models/declaration/submissions/SubmissionStatus.scala
@@ -15,136 +15,36 @@
  */
 
 package models.declaration.submissions
+import play.api.libs.json.Format
+import utils.EnumJson
 
-import play.api.libs.json._
+object SubmissionStatus extends Enumeration {
+  type SubmissionStatus = Value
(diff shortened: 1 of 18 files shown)


app/models/declaration/submissions/SubmissionStatus.scala
```diff
@@ -15,136 +15,36 @@
...
-  }
-
-  case object AdditionalDocumentsRequired extends SubmissionStatus {
-    val fullCode: String = "06"
-
-    override def toString(): String = "Additional Documents Required"
+import play.api.libs.json.Format
+import utils.EnumJson
+
+object SubmissionStatus extends Enumeration {
```
PatrykRudnicki: I can't agree with this change. Case object and sealed trait it's the best approach to use with Scala for now.

SteveSugden: But you approved same change on BE.  And look at all that useless code I got rid of.

PatrykRudnicki: I approved changed with removing function code and name code, not changes to change sealed traits to case objects

PR #524, opened 2019-09-25
CEDS-1699 Use LRN on cancelation declaration

bambuchaAdm:
We send UCR instead of LNR in cancelation request.

### DIFF
--- a/app/forms/Lrn.scala
+++ b/app/forms/Lrn.scala
@@ -31,9 +31,9 @@ object Lrn {
 
   private val lrnMaxLength = 22
 
-  val mapping = text()
-    .verifying("supplementary.consignmentReferences.lrn.error.empty", _.trim.nonEmpty)
-    .verifying("supplementary.consignmentReferences.lrn.error.length", noLongerThan(lrnMaxLength))
-    .verifying("supplementary.consignmentReferences.lrn.error.specialCharacter", isAlphanumeric)
+  def mapping(prefix: String) = text()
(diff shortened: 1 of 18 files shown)


app/forms/Lrn.scala
```diff
@@ -0,0 +1,40 @@
...
+  def nonEmpty: Boolean = !isEmpty
+}
+
+object Lrn {
+  implicit val format: Format[Lrn] =
+    Format[Lrn](Reads.StringReads.map(Lrn.apply), Writes[Lrn](lrn => JsString(lrn.value)))
+
+  private val lrnMaxLength = 22
+
+  def mapping(prefix: String) =
```
MaciejRewera: Nit: Just thinking... would it be giving any value to use an Enum for prefixes? It is easy for the client to make a mistake/mistype. 
And the Enum could be even inside `Lrn` object.

app/models/ExportsDeclaration.scala
```diff
@@ -47,8 +47,8 @@ case class ExportsDeclaration(
 ) {
   def isComplete: Boolean = status == DeclarationStatus.COMPLETE
 
-  val lrn: Option[String] = this.consignmentReferences.map(_.lrn)
```
MaciejRewera: 👍

app/forms/Lrn.scala
bambuchaAdm: This is used on level on other level mapping - it complete to user for this API it hard to tell what are good solution here. From my site i followed KISS here.

PR #525, opened 2019-09-25
CEDS-1612 Audit userInput when SaveAndReturn is chosen

PauloPortugal:


app/controllers/navigation/Navigator.scala
```diff
@@ -21,14 +21,19 @@ import controllers.util.{FormAction, SaveAndReturn}
...
+class Navigator @Inject()(appConfig: AppConfig, auditService: AuditService) {
 
-  def continueTo(call: Call)(implicit req: JourneyRequest[AnyContent]): Result =
+  def continueTo(call: Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
     FormAction.bindFromRequest match {
-      case SaveAndReturn => goToDraftConfirmation()
-      case _             => Results.Redirect(call)
+      case SaveAndReturn =>
+        auditService.auditAllPagesUserInput(Json.toJson(req.cacheModel).as[JsObject])
+        goToDraftConfirmation()
```
ed0906: Can we stick this Json.toJson stuff inside the audit service? It seems very specific to the Audit Service so probably belongs in there

test/base/MockNavigator.scala
```diff
@@ -21,21 +21,26 @@ import models.requests.JourneyRequest
...
 import play.api.mvc.{AnyContent, Call, Result}
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.http.logging.Authorization
 
 trait MockNavigator extends MockitoSugar with BeforeAndAfterEach { self: MockitoSugar with Suite =>
 
   protected val navigator: Navigator = mock[Navigator]
   protected val aRedirectToTheNextPage: Result = mock[Result]
+  protected val hc: HeaderCarrier = HeaderCarrier(
+    authorization = Some(Authorization(TestHelper.createRandomString(255)))
```
ed0906: Pretty sure you can just do HeaderCarrier() to create a new one

test/controllers/navigation/NavigatorTest.scala
```diff
@@ -39,7 +45,9 @@ class NavigatorTest extends WordSpec with Matchers with MockitoSugar with Export
 
   private val call = Call("GET", "url")
   private val config = mock[AppConfig]
-  private val navigator = new Navigator(config)
+  private val auditService = mock[AuditService]
+  private val hc: HeaderCarrier = HeaderCarrier(authorization = Some(Authorization(TestHelper.createRandomString(255))))
```
ed0906: Similar here

ed0906: Also given Audit Service is mocked you should be able to just use a mock HeaderCarrier if you want

test/base/MockNavigator.scala
PatrykRudnicki: I checked it previously and unfortunately `authorization` is necessary, because in other cases you won't be authorised. For us, it's test so it will be easier to use proposal from @ed0906 . Can we add this `Some(Authorization(TestHelper.createRandomString(255)))` as default value if we not provide anything?

Second question, I remember that there was some problems with `createRandomString` method, I'm not sure that exactly with this, but we had Chinese characters or something similar, should we use here method from `TestHelper.createRandomAlphanumericString`?

app/controllers/declaration/AdditionalDeclarationTypeController.scala
```diff
@@ -70,7 +70,7 @@ class AdditionalDeclarationTypeController @Inject()(
 
   private def updateCache(
     formData: AdditionalDeclarationType
-  )(implicit request: JourneyRequest[_]): Future[Option[ExportsDeclaration]] =
+  )(implicit request: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] =
```
PatrykRudnicki: 👍

PauloPortugal: Thanks for your opinion! :)

test/controllers/navigation/NavigatorTest.scala
PauloPortugal: This has been addressed

test/base/MockNavigator.scala
PauloPortugal: This has been addressed.

app/controllers/navigation/Navigator.scala
PauloPortugal: This involves a bigger change on code. I am going to merge in the work for QA to test the TxM audit requirements and then address this point.

PR #526, opened 2019-09-25
CEDS-1691 - re-order action fields

SteveSugden:
Something Ed suggested when I did the "conversationId" field rename but forgot to push.

### DIFF
--- a/app/models/declaration/submissions/Action.scala
+++ b/app/models/declaration/submissions/Action.scala
@@ -20,7 +20,7 @@ import java.time.LocalDateTime
 
 import play.api.libs.json.Json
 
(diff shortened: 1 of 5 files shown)


app/models/declaration/submissions/Action.scala
```diff
@@ -20,7 +20,7 @@ import java.time.LocalDateTime
 
 import play.api.libs.json.Json
 
-case class Action(requestType: RequestType, id: String, requestTimestamp: LocalDateTime = LocalDateTime.now())
+case class Action(id: String, requestType: RequestType, requestTimestamp: LocalDateTime = LocalDateTime.now())
```
PatrykRudnicki: 👍

PR #528, opened 2019-09-26
CEDS-1669 - make declarant eoir mandatory & and remove address fields…

SteveSugden:
… from form


app/forms/declaration/DeclarantDetails.scala
```diff
@@ -25,7 +28,18 @@ object DeclarantDetails {
...
 
-  val mapping = Forms.mapping("details" -> EntityDetails.mapping)(DeclarantDetails.apply)(DeclarantDetails.unapply)
+  val declarantMapping = Forms
+    .mapping(
+      "eori" -> optional(
+        text()
+          .verifying("supplementary.eori.empty", nonEmpty)
+          .verifying("supplementary.eori.error", noLongerThan(17) and isAlphanumeric)
+      ),
+      "address" -> optional(Address.mapping)
```
PauloPortugal: `"address" -> optional(Address.mapping)`
I would delete this line, as it is no longer necessary

SteveSugden: Good suggestion.  Done!

app/forms/declaration/DeclarantDetails.scala
```diff
@@ -25,7 +27,15 @@ object DeclarantDetails {
...
   val id = "DeclarantDetails"
 
-  val mapping = Forms.mapping("details" -> EntityDetails.mapping)(DeclarantDetails.apply)(DeclarantDetails.unapply)
+  val declarantMapping = Forms
+    .mapping(
+      "eori" ->
+        text()
+          .verifying("supplementary.eori.empty", nonEmpty)
+          .verifying("supplementary.eori.error", noLongerThan(17) and isAlphanumeric)
+    )((eori => EntityDetails(Some(eori), None)))(entityDetails => entityDetails.eori)
```
ed0906: Minor but looks like some extra brackets here

ed0906: `((eori => EntityDetails(Some(eori), None)))`

PatrykRudnicki: This part is little tricky. I think it will be better to prepare new class for DeclarantDetails page. We don't need to reuse old component. Now this page doesn't have address, so in the future it will be difficult to read, why we changed this or go with this way

SteveSugden: As I mentioned last week, there is some discussion about whether this is the final solution for this problem.  We might end up removing this page altogether and extracting the EORI number of the declarant from the logged in user details.  

I therefore decided that changing the model (and the BE) wasn't worth the effort.

ed0906: I don't think we should couple our model to our form structure. I know it already is but let's not make it worse.

I think what @SteveSugden has done is fine although could maybe be done like:
`(EntityDetails.apply)(_.eori)`
You'd need a separate apply method that just accepts an `Option[String]` though

PR #529, opened 2019-09-26
CEDS-1687 Make reference in transport details and border transport mandatory

bambuchaAdm:
This PR disallow empty reference. Some fixes in messages keys, some clean up in model.

There is new extracted base class for testing `FormSpec`. Have some matcher and scala test Type Classes which make testing of play forms much easier.

Include some content changes address around this issue.


PatrykRudnicki:
Please fix test: `views.declaration.TransportDetailsViewSpec`

app/models/declaration/wcodeclarationModels.scala
```diff
@@ -16,6 +16,7 @@
 
 package models.declaration
 
+//FIXME deadcode
 case class BorderTransportMeans(
```
PatrykRudnicki: It's wcodecalrationModels, so probably it's unused. In this case please remove it

test/services/cache/ExportsDeclarationBuilder.scala
```diff
@@ -159,7 +159,7 @@ trait ExportsDeclarationBuilder {
   def withBorderTransport(
     borderModeOfTransportCode: String = "",
     meansOfTransportOnDepartureType: String = "",
-    meansOfTransportOnDepartureIDNumber: Option[String] = None
+    meansOfTransportOnDepartureIDNumber: String = "1"
```
PatrykRudnicki: Why not empty string?

test/services/cache/ExportsDeclarationBuilder.scala
```diff
@@ -227,7 +227,7 @@ trait ExportsDeclarationBuilder {
     meansOfTransportCrossingTheBorderNationality: Option[String] = None,
     container: Boolean = false,
     meansOfTransportCrossingTheBorderType: String = "",
-    meansOfTransportCrossingTheBorderIDNumber: Option[String] = None,
+    meansOfTransportCrossingTheBorderIDNumber: String = "1",
```
PatrykRudnicki: The same here

test/services/cache/ExportsDeclarationBuilder.scala
bambuchaAdm: Count not be empty

test/services/cache/ExportsDeclarationBuilder.scala
bambuchaAdm: Could not be empty - it is now mandatory field

test/services/cache/ExportsDeclarationBuilder.scala
PatrykRudnicki: Yes, but `borderModeOfTransportCode` cannot be empty too and it contain empty string

PR #534, opened 2019-09-30
CEDS-1682 Error summary for Legal Declaration Form

PatrykRudnicki:


test/views/declaration/summary/SummaryPageViewSpec.scala
```diff
@@ -38,6 +38,8 @@ class SummaryPageViewSpec
     with ViewMatchers {
 
   private val form: Form[LegalDeclaration] = LegalDeclaration.form()
+  private val formWithErrors: Form[LegalDeclaration] =
```
bambuchaAdm: was used only once - why not declare it in test case

PatrykRudnicki: Moved

PR #535, opened 2019-09-30
CEDS-1721 Fix page headings to support screen readers

PauloPortugal:
* **Remove redundant `page_title` template**. We basically had two templates doing very similar things. This ensures we only use one and stick to a pattern, where we control in a single place the `section header`, `page title` and `page hint`.
* **Add `HTML` tag ids**
* **Add `CSS` classes** to match `GDS Design System`  - This was discussed with @christophercameron-hmrc
* **Fix Headings support screen readers**


app/views/components/heading.scala.html
```diff
@@ -14,11 +14,18 @@
...
-    <legend class="heading-large" role="heading" aria="" level="1">
-        <span class="heading-secondary">@messages(title)</span>
-        <h1 style="margin:0">@messages(heading)</h1>
-    </legend>
-</fieldset>
\ No newline at end of file
+<header class="page-heading">
+    @if(sectionHeader.nonEmpty) {
+        <span class="visuallyhidden">This section is </span>
+        <span id="section-header" class="heading-secondary">@sectionHeader.get</span>
```
bambuchaAdm: are you sure that heading is used only once per page ?

MaciejRewera: Nit: To avoid using `.get` and to be more scala-stylish you could use `map` on `sectionHeader`

test/views/declaration/BorderTransportViewSpec.scala
```diff
@@ -82,10 +78,6 @@ class BorderTransportViewSpec extends UnitViewSpec with CommonMessages with Stub
       view.getElementById("title").text() mustBe messages("supplementary.transportInfo.title")
     }
 
-    "display header" in {
-      view.select("legend>h1").text() mustBe messages("supplementary.transportInfo.title")
```
bambuchaAdm: is this a regression ?

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -83,17 +83,15 @@
 
     @components.error_summary(form.errors)
 
-    @components.section_header(messages("declaration.fiscalInformation.header"))
-
-    @components.page_title(Some(messages("declaration.additionalFiscalReferences.title")))
+    @components.heading(
+        Some(messages("declaration.fiscalInformation.header")),
```
MaciejRewera: Would be more clear if you used `sectionHeader = Some(...)`

app/views/cancel_declaration.scala.html
```diff
@@ -37,7 +37,7 @@
 
         @components.error_summary(form.errors)
 
-        @components.page_title(Some("cancellation.title"))
+        @components.heading(title = Some("cancellation.title"))
```
ed0906: 👍

app/views/components/heading.scala.html
```diff
@@ -14,11 +14,18 @@
...
+    }
+
+    @if(title.nonEmpty) {
+        <h1 id="title" style="margin: 0px">@title.get</h1>
```
ed0906: Is there a `m-0` class you can use instead of the inline style?  If not can we create one please :)

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -83,17 +83,15 @@
 
     @components.error_summary(form.errors)
 
-    @components.section_header(messages("declaration.fiscalInformation.header"))
-
-    @components.page_title(Some(messages("declaration.additionalFiscalReferences.title")))
```
ed0906: Can `page_title` be deleted or is the component still used? Similar for `section_header`

test/views/declaration/BorderTransportViewSpec.scala
PauloPortugal: No, this is no longer valid and it was being tested twice. Redundant

app/views/components/heading.scala.html
PauloPortugal: This has been addressed.

app/views/declaration/additional_fiscal_references.scala.html
PauloPortugal: Yes it would! :) Done

app/views/components/heading.scala.html
PauloPortugal: yes it is :).
Thank for notificing that!

app/views/declaration/additional_fiscal_references.scala.html
PauloPortugal: Yes good point I was thinking about that either! :stuck_out_tongue:

app/assets/stylesheets/partials/_base.scss
```diff
@@ -27,6 +27,25 @@ padding: 0;
...
+  margin-top: 0px;
+  margin-bottom: 20px;
+}
+
+.page-heading .heading-secondary {
+  line-height: 1.25;
+}
+
+.page-heading .title {
+  margin: 0px;
```
PauloPortugal: @ed0906 here is the CSS styling

PR #538, opened 2019-10-02
CEDS-1523 - retain the previously selected option...

SteveSugden:
...when returning to the choice page

### DIFF
--- a/app/controllers/ChoiceController.scala
+++ b/app/controllers/ChoiceController.scala
@@ -47,10 +47,10 @@ class ChoiceController @Inject()(
 
   private val logger = Logger(this.getClass)
 
-  def displayPage(previousChoice: Option[String] = None): Action[AnyContent] = authenticate.async { implicit request =>
-    def pageForPreviousChoice(previousChoice: Option[String]) = {
+  def displayPage(previousChoice: Option[Choice]): Action[AnyContent] = authenticate.async { implicit request =>
+    def pageForPreviousChoice(previousChoice: Option[Choice]) = {
       val form = Choice.form()
-      choicePage(previousChoice.fold(form)(choice => form.fill(Choice(choice))))
+      choicePage(previousChoice.fold(form)(form.fill))
     }
 
     request.declarationId match {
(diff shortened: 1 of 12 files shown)


app/controllers/ChoiceController.scala
```diff
@@ -47,14 +47,19 @@ class ChoiceController @Inject()(
 
   private val logger = Logger(this.getClass)
 
-  def displayPage(): Action[AnyContent] = authenticate.async { implicit request =>
+  def displayPage(previousChoice: Option[String] = None): Action[AnyContent] = authenticate.async { implicit request =>
+    def pageForPreviousChoice(previousChoice: Option[String]) = previousChoice match {
+      case Some(choice) => choicePage(Choice.form().fill(Choice(choice)))
```
bambuchaAdm: what do you think about
```
val form = Choice.form()
previousChoice.fold(form)(form.fill)
```
?

conf/app.routes
```diff
@@ -42,6 +42,6 @@ GET         /start                                         controllers.StartCont
 
 # Declaration Choice Page
 
-GET         /choice                                        controllers.ChoiceController.displayPage()
+GET         /choice                                        controllers.ChoiceController.displayPage(previousChoice: Option[String] ?= None)
```
bambuchaAdm: Can we add `QueryStringBindable` for `Choice` ? 

I am thinking about usage of Null Object Pattern here with object `NoChoice` as explicit null but maybe it is too much complication ?

app/controllers/ChoiceController.scala
SteveSugden: Not sure it adds to readability.  But because I don't do much folding I've done the change

conf/app.routes
SteveSugden: I agree! (That it's too much complication)

bambuchaAdm: Depending of which code - little more complicated at at `Choice` definition but all clients are not have to deal with Option everywhere

SteveSugden: I've implemented Lukasz'z suggestion, if you want to take a 2nd look

SteveSugden:
test this please

PR #539, opened 2019-10-02
CEDS-1773 - additional holder of authorisation codes for No Deal 

SteveSugden:


test/services/HolderOfAuthorisationCodeSpec.scala
```diff
@@ -41,5 +41,11 @@ class HolderOfAuthorisationCodeSpec extends UnitSpec {
       val codes = HolderOfAuthorisationCode.all.filter(code => Set("ACE", "UKCS").contains(code.value))
       codes mustBe List(HolderOfAuthorisationCode("ACE"), HolderOfAuthorisationCode("UKCS"))
     }
+
+    "contain values for CEDS-1773" in {
```
PatrykRudnicki: I think it should be differen name for the test

SteveSugden: Got any suggestions?

PR #540, opened 2019-10-02
CEDS-1772 - align page title with GOV.UK standard

SteveSugden:
Initial commit for comment


app/views/Title.scala
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package views
+
+import play.api.i18n.Messages
+
+object Title {
+
+  def createTitle(headingKey: String, sectionKey: Option[String] = None)(implicit messages: Messages): String =
```
bambuchaAdm: can we goes with default in `sectionKey` as empty string instead of Option - having empty would means not use it and leads to much easier call site

SteveSugden: Sure. I thought "" was the devils work?

ed0906: This could be a case class e.g 
```
case class Title(headingKey: String, sectionKey: Option[String] = None) {
    def toString(implicit messages: Messages) = ...
}
```

Then main template can be changed to only accept a `Title` type

SteveSugden: Adopted Ed's suggestion

app/views/declaration/items_summary.scala.html
```diff
@@ -18,49 +18,44 @@
...
 @this(main_template: views.html.main_template)
 
 
 @(mode: Mode, items: List[ExportItem], itemsErrors: Seq[FormError] = Seq.empty)(implicit request: Request[_], messages: Messages)
 
-@main_template(
-    title = messages("supplementary.items")
-) {
-    @components.back_link(routes.PreviousDocumentsController.displayPage(mode), mode)
+@defining(if(items.isEmpty) { messages("supplementary.itemsAdd.title") }
```
bambuchaAdm: can we extract this as local twirl variable ?

bambuchaAdm: I think use case for non, singular, plural use case would be quite ofter - maybe having simple apply method like
```
def apply(count: Int, none: String, singular: String, plural: String): Title = count match {
  case 0 => none
  case 1 => singular
  case _ => plural
}
```
As helper method for this case would simplify this code.

Extra: AFAIR java.text.format have support for conditional messages - maybe it would be worth to have quick look over topic.

app/views/declaration/seal_summary.scala.html
```diff
@@ -21,27 +21,30 @@
...
 @(mode: Mode, form: Form[YesNoAnswer], container: Option[Container])(implicit request: Request[_], messages: Messages)
 
 @defining(container.map(_.id).getOrElse("")) { containerId =>
 
-    @main_template(
-        title = messages("standard.seal.title", containerId)
-    ) {
-
-        @components.back_link(controllers.declaration.routes.TransportContainerController.displayContainerSummary(mode), mode)
+    @defining(container.map(_.seals).getOrElse(Seq.empty)) { seals: Seq[Seal] =>
```
bambuchaAdm: Same as items but with two arguments apply ?

app/views/declaration/summary/summary_page.scala.html
```diff
@@ -25,13 +25,18 @@
...
+@import views.Title
 
 @this(main_template: views.html.main_template)
 
 @(mode: Mode, suppDecData: SupplementaryDeclarationData, legalDeclarationForm: Form[LegalDeclaration])(implicit request: JourneyRequest[_], messages: Messages, appConfig: AppConfig)
 
 @main_template(
-    title = messages("supplementary.summary.title")
+    title = mode match {
+        case Mode.Normal => Title("supplementary.summary.normal-header")
```
bambuchaAdm: Can we move this logic to Title companion ?

test/views/TitleSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package views
+
+import org.scalatest.MustMatchers
+import views.declaration.spec.UnitViewSpec
+
+class TitleSpec extends UnitViewSpec with MustMatchers {
+
+  override val messages = realMessagesApi.preferred(request)
```
bambuchaAdm: can we do this with messages stub ?

PR #541, opened 2019-10-03
CEDS-1543 Added validation for additional fiscal references

PatrykRudnicki:


app/services/cache/ExportItem.scala
```diff
@@ -43,8 +43,13 @@ case class ExportItem(
     fiscalInformation.exists(_.onwardSupplyRelief == FiscalInformation.AllowedFiscalInformationAnswers.yes)
 
   def isCompleted: Boolean =
-    procedureCodes.isDefined && fiscalInformation.isDefined && itemType.isDefined &&
+    procedureCodes.isDefined && isFiscalInformationCompleted && itemType.isDefined &&
       packageInformation.nonEmpty && commodityMeasure.isDefined && additionalInformation.isDefined
+
+  private def isFiscalInformationCompleted: Boolean =
```
bambuchaAdm: can we move this to fiscal information object ?

app/services/cache/ExportItem.scala
```diff
@@ -43,8 +43,13 @@ case class ExportItem(
...
+    if (fiscalInformation.isDefined && fiscalInformation.map(_.onwardSupplyRelief) == Some(yes))
```
bambuchaAdm: fiscalInformation.exists(_.isDefined)

PatrykRudnicki: Changed implementation

app/services/cache/ExportItem.scala
PatrykRudnicki: this method based on information in fiscal information and additional fiscal information. Decision about completed is connected only from whole item perspective. I think this is better place for this method

PR #547, opened 2019-10-07
CEDS-1673 Remove unnecessary PUT

ed0906:
- Remove WCO library as its no longer used
- Remove Request from service layer

### DIFF
--- a/app/controllers/declaration/SummaryController.scala
+++ b/app/controllers/declaration/SummaryController.scala
@@ -70,7 +70,7 @@ class SummaryController @Inject()(
             BadRequest(summaryPage(Mode.Normal, SupplementaryDeclarationData(request.cacheModel), formWithErrors))
         ),
         legalDeclaration => {
-          submissionService.submit(request.authenticatedRequest.user.eori, request.cacheModel, legalDeclaration).map {
+          submissionService.submit(request.eori, request.cacheModel, legalDeclaration).map {
             case Some(lrn) =>
               Redirect(controllers.declaration.routes.ConfirmationController.displaySubmissionConfirmation())
                 .flashing(Flash(Map("LRN" -> lrn)))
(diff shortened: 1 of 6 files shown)


app/controllers/declaration/SummaryController.scala
```diff
@@ -70,7 +70,7 @@ class SummaryController @Inject()(
             BadRequest(summaryPage(Mode.Normal, SupplementaryDeclarationData(request.cacheModel), formWithErrors))
         ),
         legalDeclaration => {
-          submissionService.submit(request.cacheModel, legalDeclaration).map {
+          submissionService.submit(request.authenticatedRequest.user.eori, request.cacheModel, legalDeclaration).map {
```
PatrykRudnicki: I'm not sure but we should have `eori` field in `JourneyRequest` so you don't need to go through `authenticatedRequest.user.eori`

ed0906: It wasn't on the request but it is now 👍

app/models/requests/JourneyRequest.scala
```diff
@@ -23,4 +23,5 @@ import play.api.mvc.WrappedRequest
 class JourneyRequest[A](val authenticatedRequest: AuthenticatedRequest[A], val cacheModel: ExportsDeclaration)
     extends AuthenticatedRequest[A](authenticatedRequest, authenticatedRequest.user) {
   val choice: Choice = Choice(cacheModel.choice)
+  def eori: String = authenticatedRequest.user.eori
```
PatrykRudnicki: It can be a `val`, but it's okay :)

app/services/SubmissionService.scala
```diff
@@ -37,55 +35,71 @@ import scala.util.{Failure, Success}
...
   exportsMetrics: ExportsMetrics
 ) {
 
   private val logger = Logger(this.getClass)
 
-  def submit(
-    exportsDeclaration: ExportsDeclaration,
-    legalDeclaration: LegalDeclaration
-  )(implicit request: JourneyRequest[_], hc: HeaderCarrier, ec: ExecutionContext): Future[Option[String]] = {
+  def submit(eori: String, exportsDeclaration: ExportsDeclaration, legalDeclaration: LegalDeclaration)(
```
bambuchaAdm: why eori as string ?

PR #550, opened 2019-10-08
BAU - fix to prevent modified container being added to the end of the…

SteveSugden:
… container list

### DIFF
--- a/app/controllers/declaration/SealController.scala
+++ b/app/controllers/declaration/SealController.scala
@@ -153,15 +153,15 @@ class SealController @Inject()(
           Future.successful(navigator.continueTo(routes.SealController.displaySealSummary(mode, cachedContainer.id)))
     )
 
-  private def updateCache(container: Container)(implicit req: JourneyRequest[AnyContent]) =
+  private def updateCache(updatedContainer: Container)(implicit req: JourneyRequest[AnyContent]) =
     updateExportsDeclarationSyncDirect(model => {
 
       val updatedContainers =
         model.containerData
           .map(_.containers)
-          .fold(Seq(container))(_.map {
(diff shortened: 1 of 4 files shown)


app/controllers/declaration/SealController.scala
```diff
@@ -159,7 +159,10 @@ class SealController @Inject()(
       val updatedContainers =
         model.containerData
           .map(_.containers)
-          .fold(Seq(container))(_.filterNot(_.id == container.id) :+ container)
+          .fold(Seq(container))(_.map {
+            case c if (c.id == container.id) => container
+            case x                           => x
```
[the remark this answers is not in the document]
SteveSugden: Good point.  And I've removed the unnecessary () before Ed gets on my case again!

app/controllers/declaration/SealController.scala
```diff
@@ -153,13 +153,16 @@ class SealController @Inject()(
...
 
-  private def updateCache(container: Container)(implicit req: JourneyRequest[AnyContent]) =
+  private def updateCache(updatedContainer: Container)(implicit req: JourneyRequest[AnyContent]) =
     updateExportsDeclarationSyncDirect(model => {
 
       val updatedContainers =
         model.containerData
           .map(_.containers)
-          .fold(Seq(container))(_.filterNot(_.id == container.id) :+ container)
+          .fold(Seq(updatedContainer))(_.map {
```
bambuchaAdm: Please move map logic for Containers object

SteveSugden: Good idea.

SteveSugden: Can you take a 2nd look @bambuchaAdm

PR #552, opened 2019-10-10
CEDS-1695 Showing data field for errors in rejected notifications screen

PatrykRudnicki:


test/views/RejectedNotificationErrorsViewSpec.scala
```diff
@@ -51,31 +52,48 @@ class RejectedNotificationErrorsViewSpec extends UnitViewSpec with Stubs with In
...
     "have correct title" in {
 
-      view.getElementById("title").text() mustBe messages("rejected.notification.title")
+      view().getElementById("title").text() mustBe messages("rejected.notification.title")
     }
 
     "have correct back link" in {
 
-      val backLink = view.getElementById("link-back")
+      val backLink = view().getElementById("link-back")
```
bambuchaAdm: can we extract default arg view render as variable and don't render it for each case ?

app/models/Pointer.scala
```diff
@@ -0,0 +1,42 @@
...
+      case Failure(_) => s
+    }
+  }.mkString(".")
+
+  override def toString: String = value
+}
+
+object Pointer {
+  implicit val format: Format[Pointer] = Format(
+    Reads(js => js.validate[JsString].map(string => Pointer(string.value))),
```
bambuchaAdm: Why not validate just asString ?

app/models/Pointer.scala
```diff
@@ -0,0 +1,42 @@
...
+    Writes(pointer => JsString(pointer.value))
+  )
+
+  def apply(value: String): Pointer = Pointer(value.split("\\.").toList)
```
bambuchaAdm: Please use option with `Char` not pattern based

test/views/RejectedNotificationErrorsViewSpec.scala
PatrykRudnicki: yup

app/models/Pointer.scala
PatrykRudnicki: don't know what you mean

test/views/RejectedNotificationErrorsViewSpec.scala
ed0906: Whats the harm?

app/models/Pointer.scala
bambuchaAdm: Sry i miss this with `replace(Char)` - split does not have split

test/views/RejectedNotificationErrorsViewSpec.scala
bambuchaAdm: Rendering 10x allocates big string 10x and puts unnecessary GC pressure on tests. Unless we expect that view render with same parameters give use different outcomes.

app/models/Pointer.scala
PatrykRudnicki: Changed

PR #553, opened 2019-10-10
CEDS-1802 Change order of columns on rejected notification page

PatrykRudnicki:
Added `thead` and `tbody` tags after discussion with design.

### DIFF
--- a/app/views/rejected_notification_errors.scala.html
+++ b/app/views/rejected_notification_errors.scala.html
@@ -35,21 +35,25 @@
     <span id="information">@messages("rejected.notification.information")</span>
 
     <table>
-        <tr class="table-row">
-            <th class="table-cell table-head">@messages("rejected.notification.header.fieldName")</th>
-            <th class="table-cell table-head">@messages("rejected.notification.header.errorCode")</th>
-            <th class="table-cell table-head">@messages("rejected.notification.header.errorDescription")</th>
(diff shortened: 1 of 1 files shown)


app/views/rejected_notification_errors.scala.html
```diff
@@ -35,21 +35,25 @@
...
-            <th class="table-cell table-head" id="rejected_notifications-row-@index-name">
-                @error.pointer.map(messages(_))
-            </th>
-            <th class="table-cell" id="rejected_notifications-row-@index-code">@error.code</th>
-            <th class="table-cell" id="rejected_notifications-row-@index-description">@error.description</th>
-          </tr>
-        }
+        <thead>
+            <tr class="table-row">
+                <th class="table-cell table-head">@messages("rejected.notification.header.fieldName")</th>
```
bambuchaAdm: please add class for columns names - like "field name" would be much easier to manage test

PatrykRudnicki: I'll add class name, but I can't agree that it'll be easier, those elements has unique id and `field-name` can be used somewhere in the css. I'll add `rejected-field-name` to be sure that this change doesn't change the view or doesn't apply styles that we don't want.

PatrykRudnicki: Changed

PR #554, opened 2019-10-10
CEDS-1804 Remove Option from id in ExportsDeclaration

bambuchaAdm:


app/connectors/CustomsDeclareExportsConnector.scala
```diff
@@ -50,14 +50,11 @@ class CustomsDeclareExportsConnector @Inject()(appConfig: AppConfig, httpClient:
 
   private val createTimer: Timer = metrics.defaultRegistry.timer("declaration.create.timer")
 
-  def createDeclaration(declaration: ExportsDeclaration)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[ExportsDeclaration] = {
+  def createDeclaration(declaration: ExportsDeclarationExchange)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[ExportsDeclaration] = {
```
ed0906: Hmmm this means other layers are now exposed to our request/response object

bambuchaAdm: To be true only layer expose for it is service layer which does not have logic at all (expect update modification time which should be moved to backend) - having brainless forwarding layer just to have it is little wired

bambuchaAdm: Moreover only thing that frontend require to pass full declaration is for creating amendment - if e would have endpoint on backend to manage this (eg. `POST /declarations/:id/amendment`) we won't have to transport all declaration things back and forward between backend and frontend.

PR #556, opened 2019-10-11
CEDS-1726 add and remove button hidden text

SteveSugden:


app/views/components/add_save_button.scala.html
```diff
@@ -18,10 +18,14 @@
...
     saveKey: String = "site.save_and_continue",
-    addKey: String = "site.add"
+    addKey: String = "site.add",
+    addHint: String = "site.add"
 )(implicit messages: Messages)
 
 <div class="section align='right'">
-    <button id="add" class="button--secondary" name="@Add.toString">@messages(addKey)</button>
+    <button id="add" class="button--secondary" name="@Add.toString">
+    <span aria-hidden="true">@messages(addKey)</span>
```
MaciejRewera: Nit: I think `span` should be indented

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -60,13 +60,10 @@
...
-                                type="submit"
-                                class="button--secondary"
-                                formmethod="post"
-                                formaction="@routes.AdditionalFiscalReferencesController.removeReference(mode, itemId, elem)"
-                                value="@messages("site.remove")"
-                            >
+                            <button class="remove button--secondary" type="submit" formmethod="post" formaction="@routes.AdditionalFiscalReferencesController.removeReference(mode, itemId, elem)">
+                                <span aria-hidden="true">@messages("site.remove")</span>
+                                <span class="visuallyhidden">@{messages("declaration.additionalFiscalReferences.remove.hint", elem)}</span>
+                            </button>
```
MaciejRewera: Nit: It might be worth considering to create a dedicated template for secondary buttons?

app/views/components/fields/field_multiple_values.scala.html
```diff
@@ -27,7 +27,9 @@
     labelClass: Option[String] = None,
     input: Html,
     includeAddButton: Boolean = true,
-    valuePrefix: Option[String] = None
+    valuePrefix: Option[String] = None,
+    addHint: String = "site.add",
+    removeHint: String = "site.remove.hint"
 )(implicit messages: Messages)
```
MaciejRewera: Nit: This component's arguments list grew signifficantly. We might need to think about either splitting the component (as it might be doing too much things), or to somehow reduce the amount of parameters.

app/views/declaration/additional_fiscal_references.scala.html
SteveSugden: We already have the remove_button template that I tried to use here but that got rejected.  This 'button' is unique in that it defines a formaction.

app/views/components/fields/field_multiple_values.scala.html
SteveSugden: It's a sign that it had lots of configuration options.  Not sure what the solution is though.  Either the component is re-usable or we create multiple components for use on specific forms/pages

test/views/declaration/AdditionalInformationViewSpec.scala
```diff
@@ -143,7 +143,7 @@ class AdditionalInformationViewSpec extends UnitViewSpec with AdditionalInformat
       view.select("table>tbody>tr>th:nth-child(1)").text() mustBe "12345-12345"
 
       val removeButton = view.select("table>tbody>tr>th:nth-child(2)>button")
-      removeButton.text() mustBe messages(removeCaption)
+      removeButton.text() mustBe "site.remove site.remove.hint"
```
MaciejRewera: I got tricked by this, thinking it is single message. Could you split it into 2 lines and use for e.g. `include("...")`?

test/views/declaration/PreviousDocumentsViewSpec.scala
```diff
@@ -230,14 +230,14 @@ class PreviousDocumentsViewSpec extends UnitViewSpec with ExportsTestData with S
       view.select("form>table>thead>tr>th:nth-child(2)").text() must be("supplementary.previousDocuments.documentType.label")
       view.select("form>table>thead>tr>th:nth-child(3)").text() must be("supplementary.previousDocuments.documentReference.label")
       view.select("form>table>thead>tr>th:nth-child(4)").text() must be("supplementary.previousDocuments.goodsItemIdentifier.label")
-      view.select("form>table>thead>tr>th:nth-child(5)").text() must be("supplementary.packageInformation.remove")
+      view.select("form>table>thead>tr>th:nth-child(5)").text() must be("supplementary.previousDocuments.remove")
```
MaciejRewera: Do we have a better CSS selector for this, or not yet?

test/views/declaration/AdditionalInformationViewSpec.scala
SteveSugden: Could do.  But isn't it better to be more precise?  (I.e. use mustBe).   To be honest, I'm not 100% convinced of the value of tests like this.  I just fixed what I broke.

app/views/components/fields/field_multiple_values.scala.html
MaciejRewera: So maybe we could wrap all configuration params into one class and pass it then?

test/views/declaration/AdditionalInformationViewSpec.scala
SteveSugden: If we think it's worth testing the hint text separately then maybe we need to extract the text of the two spans contained within the button and test them separately?

test/views/declaration/PreviousDocumentsViewSpec.scala
SteveSugden: Don't know.  Was just fixing a typo.

PR #557, opened 2019-10-11
Add possibility to have multiple messages files for one language

MaciejRewera:


MaciejRewera:
This is not the consequence of a need, but to add the possibility. It is also to encourage us to start following this path.

PR #563, opened 2019-10-14
CEDS-1448 remove col heading

SteveSugden:


app/views/declaration/package_information.scala.html
```diff
@@ -59,7 +59,7 @@
                         <th scope="col">@messages("supplementary.packageInformation.typesOfPackages")</th>
                         <th scope="col">@messages("supplementary.packageInformation.numberOfPackages")</th>
                         <th scope="col">@messages("supplementary.packageInformation.shippingMarks")</th>
-                        <th scope="col">@messages("supplementary.packageInformation.remove")</th>
+                        <td></td>
```
PatrykRudnicki: If you remove whole td, it should work similar, but not sure

SteveSugden: Followed the advice in the article Chris linked in Slack.  Empty <td> used to pad un-used header cells.

PR #564, opened 2019-10-14
CEDS-1404 Add DocumentIdentifer hint

PauloPortugal:
also removed unused text messages


test/views/declaration/DocumentsProducedViewSpec.scala
```diff
@@ -56,10 +56,11 @@ class DocumentsProducedViewSpec
 
       val messages = instanceOf[MessagesApi].preferred(request)
       // FIXME Not used on view -> for remove this
```
PatrykRudnicki: Can you fix this TODO? Probably it's about unused messages in the view. If all below messages in view are used please remove this TODO

PauloPortugal: Ahhh! I did that, I just forgot to remove the comment.

PR #565, opened 2019-10-14
CEDS-1675 Change the case of TXM ids to camelcase

PauloPortugal:
This was the feedback from TxM team.

### DIFF
--- a/app/services/audit/AuditService.scala
+++ b/app/services/audit/AuditService.scala
@@ -98,5 +98,5 @@ object AuditTypes extends Enumeration {
 }
 object EventData extends Enumeration {
   type Data = Value
-  val EORI, LRN, MRN, DUCR, DecType, ChangeReason, ChangeDescription, FullName, JobRole, Email, Confirmed, SubmissionResult, Success, Failure = Value
+  val eori, lrn, mrn, ducr, decType, changeReason, changeDescription, fullName, jobRole, email, confirmed, submissionResult, Success, Failure = Value
 }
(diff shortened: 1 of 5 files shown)


app/services/audit/AuditService.scala
```diff
@@ -98,5 +98,5 @@ object AuditTypes extends Enumeration {
 }
 object EventData extends Enumeration {
   type Data = Value
-  val EORI, LRN, MRN, DUCR, DecType, ChangeReason, ChangeDescription, FullName, JobRole, Email, Confirmed, SubmissionResult, Success, Failure = Value
+  val eori, lrn, mrn, ducr, decType, changeReason, changeDescription, fullName, jobRole, email, confirmed, submissionResult, Success, Failure = Value
```
PatrykRudnicki: The same question like in movements, what about Success and Failure?

test/services/audit/AuditServiceSpec.scala
```diff
@@ -107,10 +107,10 @@ trait AuditTestSupport extends UnitSpec with ExportsDeclarationBuilder with Scal
   val mockAuditConnector = mock[AuditConnector]
 
   val auditData = Map(
-    EORI.toString -> "eori1",
+    eori.toString -> "eori1",
     LRN.toString -> "lrn1", // FIXME refere to bad thing
     DUCR.toString -> "ducr1",
```
PatrykRudnicki: Here you're using wrong labels, that contains probably correct value, but this LRN and DUCR are not from EventData

app/services/audit/AuditService.scala
ed0906: I think they get used as a value, rather than a key in the JSON

PauloPortugal: That is a value not a key. TxM wanted the keys to be `camelCase`

test/services/audit/AuditServiceSpec.scala
PauloPortugal: Those are fine, here but I am going to explicitly refer to EvenData.

PR #566, opened 2019-10-14
CEDS-1718 Handle Sequence Pointers e.g Additional Doc ID

ed0906:


conf/messages.en
```diff
@@ -877,3 +877,4 @@ rejected.notification.continue = Continue declaration (opens in a new window)
 summary.amend.information = You will start at the beginning of the declaration
 
 field.declaration.consignmentReferences.lrn=LRN
+field.declaration.goodsShipment.governmentAgencyGoodsItem.$.additionalDocument.$.id=Item {0} - Additional Document {1} - ID
```
SteveSugden: Is this just a single example of a message derived from a Pointer?   Will these messages go in their own file eventually?

ed0906: Yeh at the moment we just convert the pointer to a "field name" which we show in the UI.
This will probably change depending on the requirements on how we display errors

PR #567, opened 2019-10-15
CEDS 1718 - Display Hierachical Field Pointers

ed0906:
This allows us to have Field pointers that look like this in the UI

```
Item 1
 - Document 2
  - ID
```
It's done using CSS so we can be consistent across fields & change the design easily.


conf/messages.en
```diff
@@ -871,4 +871,4 @@ rejected.notification.continue = Continue declaration (opens in a new window)
 summary.amend.information = You will start at the beginning of the declaration
 
 field.declaration.consignmentReferences.lrn=LRN
-field.declaration.goodsShipment.governmentAgencyGoodsItem.$.additionalDocument.$.id=Item {0} > Additional Document {1} > ID
+field.declaration.goodsShipment.governmentAgencyGoodsItem.$.additionalDocument.$.id=<ol><li>Goods Item {0}</li><li>Document {1}</li><li>ID</li></ol>
```
ed0906: Using ordered lists hopefully keeps this pattern accessible

PatrykRudnicki: I'm just thinking that it's a correct place to build the styles.

bambuchaAdm: I don't think that HTML in messages is good idea - it hide structure from template and spread it over project

PR #568, opened 2019-10-15
CEDS-1094 - Item Type page Add buttons add values from both fields

SteveSugden:
Looks worse than it is due to code refactoring.  Main changes
- ability for the 'Add' button to specify which fieldlist we are adding to
- fix related bug whereby removing an item caused edits to be lost
- separating the "form" data structure (ItemTypeForm) from the cache mode (ItemType)
- moved the 'ExportItem' class from 'services.cache' to 'models.declaration'

The major logic changes are in `ItemTypeController`


app/controllers/declaration/ItemTypeController.scala
```diff
@@ -81,121 +81,123 @@ class ItemTypeController @Inject()(
         val hasFiscalReferences = item.hasFiscalReferences
         val formAction = FormAction.bindFromRequest()
         formAction match {
-          case Add =>
-            handleAddition(mode, itemId, itemTypeInput, itemTypeCache, hasFiscalReferences)
+          case Add(field) =>
```
bambuchaAdm: I am not sure about clearance of this solution.

app/forms/declaration/ItemTypeForm.scala
```diff
@@ -0,0 +1,68 @@
...
+    statisticalValueKey -> text()
+  )(ItemTypeForm.apply)(ItemTypeForm.unapply)
+
+  val id = "ItemType"
+
+  def form(): Form[ItemTypeForm] = Form(mapping)
+
+  val empty: ItemTypeForm = ItemTypeForm("", None, None, "", None, None, "")
+
+  def convert(model: models.declaration.ItemType) =
```
PatrykRudnicki: I think this method should have different name, `convert` doesn't describe anything

PatrykRudnicki: Please impot `ItemType`

ed0906:
I'm open to the idea of extending `Add` like you have, perhaps we could do it in such a way that other controllers remain unaffected e.g
```
case class Add(name: Option[String])
case object Add extends Add(None)
```

PR #569, opened 2019-10-15
CEDS 1094 Item Type view Add button adds values from both code fields

SteveSugden:
Second attempt.    A combination of Patryk's and Ed's suggestions.


app/controllers/util/FormAction.scala
```diff
@@ -42,9 +44,10 @@ object FormAction {
...
 }
 
-case object Unknown extends FormAction
 case object Add extends FormAction
+case object Unknown extends FormAction
 case object SaveAndContinue extends FormAction
 case object SaveAndReturn extends FormAction
 case object Continue extends FormAction
 case class Remove(keys: Seq[String]) extends FormAction
+case class AddField(field: Option[String]) extends FormAction
```
ed0906: Do we ever do `AddField(None)`?

SteveSugden: No.  We have old Add for that.  I'll make it non-optional

SteveSugden:
ItemTypeSpec didn't do anything and wasn't referenced

PatrykRudnicki:
@SteveSugden But still we don't have tests for new form.

app/controllers/util/FormAction.scala
```diff
@@ -32,6 +33,7 @@ object FormAction {
   def bindFromRequest()(implicit request: Request[AnyContent]): FormAction =
     request.body.asFormUrlEncoded.flatMap { body =>
       body.flatMap {
+        case (`addFieldLabel`, values)   => Some(AddField(values.headOption))
```
bambuchaAdm: Are we interested in validating if there is more options here - just to make BadRequest ?

SteveSugden:
We don't seem to have many(any?) tests for "forms" in `forms.declaration` package

PatrykRudnicki:
Unit: https://github.com/hmrc/customs-declare-exports-frontend/tree/master/test/unit/forms/declaration

rest of the specs: https://github.com/hmrc/customs-declare-exports-frontend/tree/master/test/forms/declaration

We have tests for everything

PatrykRudnicki:
There wasn't enough time to move everything to unit

PR #570, opened 2019-10-15
CEDS-1726 - refactored remove button hint 

SteveSugden:
to enable sensible remove button messages for item type page


app/views/components/fields/LabelledValue.scala
```diff
@@ -19,5 +19,5 @@ package views.components.fields
 case class LabelledValue(label: String, value: String)
 
 object LabelledValue {
-  def apply(labelAndValue: String): LabelledValue = LabelledValue(labelAndValue, labelAndValue)
+  def apply(labelAndValue: Any): LabelledValue = LabelledValue(labelAndValue.toString, labelAndValue.toString)
```
bambuchaAdm: Why any ?

PatrykRudnicki: Why we need to change to `Any`? Is it a possibility to receive something different than `String`?

SteveSugden: Yes.  To save the client calling toString on the Json object

app/views/declaration/item_type.scala.html
```diff
@@ -71,7 +71,8 @@
             elements = taricAdditionalCodes,
             labelClass = Some("bold-small"),
             valuePrefix = Some(taricAdditionalCodesKey+"_"),
-            addHint = "declaration.itemType.taricAdditionalCodes.add.hint"
+            addHint = "declaration.itemType.taricAdditionalCodes.add.hint",
+            removeHint = "declaration.itemType.taricAdditionalCodes.remove.hint"
```
ed0906: 👍

app/views/declaration/item_type.scala.html
```diff
@@ -86,7 +87,8 @@
                 items = AutoCompleteItem.fromNationalAdditionalCode(NationalAdditionalCode.all)
             ),
             valuePrefix = Some(nationalAdditionalCodesKey+"_"),
-            addHint = "declaration.itemType.nationalAdditionalCode.add.hint"
+            addHint = "declaration.itemType.nationalAdditionalCode.add.hint",
+            removeHint = "declaration.itemType.nationalAdditionalCode.remove.hint"
```
ed0906: 👍

app/views/components/fields/LabelledValue.scala
PatrykRudnicki: Can we provide two apply methods? One for `String` and second one for `JsObject`?

Using `Any` is not a good practice, you can pass everything there.

app/views/components/fields/LabelledValue.scala
```diff
@@ -16,8 +16,11 @@
 
 package views.components.fields
 
+import play.api.libs.json.JsValue
+
 case class LabelledValue(label: String, value: String)
 
 object LabelledValue {
   def apply(labelAndValue: String): LabelledValue = LabelledValue(labelAndValue, labelAndValue)
+  def apply(labelAndValue: JsValue): LabelledValue = LabelledValue(labelAndValue.toString, labelAndValue.toString)
```
PatrykRudnicki: Yeah, it's better now. I'm just wondering why we need to have the same `label` and `value`?

Of course you can create now case class with two arguments, so it's okay, but I'm just curious what case is covered by it.

bambuchaAdm: Two apply is good enough - you have still strongly types and everything should work perfectly.

bambuchaAdm: @PatrykRudnicki this depends on how MultipleItemsHelper  works - it compare sent value with values in model

PR #571, opened 2019-10-15
CEDS-1635 Unify length of input field for autocomplete

bambuchaAdm:
Behaves same as other inputs plus have same width and height.

### DIFF
--- a/app/assets/stylesheets/partials/_autocomplete.scss
+++ b/app/assets/stylesheets/partials/_autocomplete.scss
@@ -1,3 +1,5 @@
+@import "./_utilities.scss";
+
 body {
   font-family: Arial, Helvetica, sans-serif;
 }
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/partials/_utilities.scss
```diff
@@ -6,7 +9,7 @@ Utility alignments
 .u-align--right {
   text-align: right;
 }
-@media screen and (min-width: 641px){
+@media screen and (min-width: #{$media-query-breakpoint}){
```
ed0906: Pretty sure you can just do `$media-query-breakpoint`

Not sure why we've done this though?

app/assets/stylesheets/partials/_autocomplete.scss
```diff
@@ -28,4 +31,9 @@ table {
   bottom: 0;
   margin-left: -1px;
   padding: 0;
-}
+  right: -#{$custom-combobox-input-border-width};
```
ed0906: Pretty sure you can just do `-1 * $custom-combobox-input-border-width`

PR #573, opened 2019-10-16
CEDS-1668 Add feature toogle to use improved error msg's

PauloPortugal:
For each CDS error code that we know about, we are now displaying the
CDS error code description defined by the BRAT team.

These CDS error code descriptions use domain language that may not make
sense to our users, or be consistent with the language we’ve used in
our service, based on user research.

We aim to replace some or all of these CDS error code descriptions with
our own descriptions that use language consistent with our service,
that our users will understand.


app/services/model/RejectionReason.scala
```diff
@@ -61,10 +54,17 @@ object RejectionReason {
       notification.errors.map { error =>
         RejectionReason(
           error.validationCode,
-          getErrorDescription(error.validationCode),
+          getCdsErrorDescription(error.validationCode),
```
ed0906: Couldn’t we keep the model as it was and do the if toggle here?

ed0906: Keeps the logic in Scala code

conf/code-lists/errors-dms-rej-list.csv
```diff
@@ -1,142 +1,148 @@
...
-CDS11005,Relation error: The requested procedure is not the same for each item in the declaration.
-CDS11006,Relation error: The NetMass per item or the sum of all in the declaration is larger than the gross mass.
-CDS11007,Relation error: The submitter in the request/supplementary declaration differs from the related declaration.
-CDS11012,Relation error: element is not allowed to be amended through this additional message type.
-CDS12003,Uniqueness error: Reuse of submitter reference is not (yet) allowed.
-CDS12005,Authorization error: Party ID unknown or invalid.
-CDS12006,Authorization warning: invalid import/export license number.
-CDS12007,Authorization error: authorization not valid for declaration.
-CDS12008,Authorization error: requested amount or quantity is not available for authorization reservation.
+CDS10001,"Obligation error: Mandatory Data Element has not been provided.","Obligation error: Mandatory Data Element has not been provided."
```
ed0906: We could always leave the last column blank if there’s no override, and then default to the DMS version if it’s empty

app/services/model/RejectionReason.scala
PauloPortugal: Very good point!!!!

conf/code-lists/errors-dms-rej-list.csv
PauloPortugal: Yes, I thought about that. The reason I did it is that we should always have a double quotes, because of the commas we might get in the text, this makes it safe to always assume there is content and it is always wrapped in double quotes

app/services/model/RejectionReason.scala
PauloPortugal: I have just discussed this with @PatrykRudnicki.
We would have to create a new `RejectionReasonViewModel` and possibly a new `RejectionReasonService`, since it would not be a good design to inject the AppConfig into the `RejectionReason`.
After discussing other approaches it became clear that all of them had advantages and disadvantages. 
I agree with the conditionality leaking into the view though, but I cannot find a compelling enough solution to do it in another way.

PR #575, opened 2019-10-18
CEDS-1772 - Remove `office of presentation`

SteveSugden:


app/forms/declaration/officeOfExit/OfficeOfExit.scala
```diff
@@ -17,14 +17,14 @@
 package forms.declaration.officeOfExit
 import play.api.libs.json.Json
 
-case class OfficeOfExit(officeId: String, presentationOfficeId: Option[String], circumstancesCode: Option[String])
+case class OfficeOfExit(officeId: String, circumstancesCode: Option[String])
 
 object OfficeOfExit {
   implicit val format = Json.format[OfficeOfExit]
 
   def from(officeOfExitSupplementary: OfficeOfExitSupplementary): OfficeOfExit =
```
ed0906: Maybe we could merge these Supplementary & Standard classes...

SteveSugden: There is still the `circumstancesCode` field that is different

ed0906: Could easily be handled with an `Option` :)

PR #576, opened 2019-10-21
CEDS-1718 Separate Choice and DeclarationType

ed0906:
Currently the `ExportsDeclaration` model has a choice field which is a `String`.
Given the many extra dec types we have to add over the next few months we should make this easy to expand. 

`choice` seems the wrong name for this & `String` seems the wrong `type`. 
Choice also represents irrelevant options such as "view submissions", "cancel dec". 

With this PR I propose we have a separate `DeclarationType`

What this PR changes:
- Replace `choice: String` on `ExportsDeclaration` with `type: DeclarationType`
- Use an Enum for DeclarationType which only allows `STANDARD`, `SUPPLEMENTARY` (for now)

### DIFF
--- a/app/connectors/exchange/ExportsDeclarationExchange.scala
+++ b/app/connectors/exchange/ExportsDeclarationExchange.scala
@@ -21,6 +21,7 @@ import java.time.Instant
 import forms.declaration._
 import forms.declaration.additionaldeclarationtype.AdditionalDeclarationType
 import models.DeclarationStatus.DeclarationStatus
+import models.DeclarationType.DeclarationType
 import models.ExportsDeclaration
 import models.declaration.{ExportItem, Locations, Parties, TransportInformationContainerData}
 import play.api.libs.json.{Json, OFormat}
@@ -31,7 +32,7 @@ case class ExportsDeclarationExchange(
   createdDateTime: Instant,
   updatedDateTime: Instant,
   sourceId: Option[String],
-  choice: String,
+  `type`: DeclarationType,
   dispatchLocation: Option[DispatchLocation] = None,
   additionalDeclarationType: Option[AdditionalDeclarationType] = None,
   consignmentReferences: Option[ConsignmentReferences] = None,
@@ -51,7 +52,7 @@ case class ExportsDeclarationExchange(
     createdDateTime = this.createdDateTime,
     updatedDateTime = this.updatedDateTime,
     sourceId = this.sourceId,
-    choice = this.choice,
+    `type` = this.`type`,
     dispatchLocation = this.dispatchLocation,
     additionalDeclarationType = this.additionalDeclarationType,
     consignmentReferences = this.consignmentReferences,
@@ -77,7 +78,7 @@ object ExportsDeclarationExchange {
       createdDateTime = declaration.createdDateTime,
       updatedDateTime = declaration.updatedDateTime,
       sourceId = declaration.sourceId,
-      choice = declaration.choice,
+      `type` = declaration.`type`,
       dispatchLocation = declaration.dispatchLocation,
       additionalDeclarationType = declaration.additionalDeclarationType,
       consignmentReferences = declaration.consignmentReferences,
--- a/app/controllers/ChoiceController.scala
+++ b/app/controllers/ChoiceController.scala
@@ -78,7 +78,7 @@ class ChoiceController @Inject()(
                 .getOrElse(throw new IllegalArgumentException(s"Cannot deduce Declaration Type from Choice [${choice.value}]"))
               request.declarationId match {
                 case Some(id) =>
-                  updateDeclarationTyoe(id, declarationType).map { _ =>
+                  updateDeclarationType(id, declarationType).map { _ =>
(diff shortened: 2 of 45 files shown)


app/controllers/ChoiceController.scala
```diff
@@ -111,7 +111,7 @@ class ChoiceController @Inject()(
           createdDateTime = Instant.now,
           updatedDateTime = Instant.now,
           sourceId = None,
-          choice.value
+          choice.toDeclarationType.get
```
PatrykRudnicki: Using just `get` will causes the exceptions for `None`.

It's for all other places too except tests.

app/models/DeclarationType.scala
```diff
@@ -0,0 +1,26 @@
...
+package models
+
+import play.api.libs.json._
+import utils.EnumJson
+
+object DeclarationType extends Enumeration {
+  type DeclarationType = Value
+  implicit val format: Format[DeclarationType.Value] = EnumJson.format(DeclarationType)
+  val STANDARD, SUPPLEMENTARY, SIMPLIFIED = Value
+}
```
ed0906: Highlighting a Key change

app/forms/Choice.scala
```diff
@@ -16,13 +16,21 @@
...
 import utils.validators.forms.FieldValidator.isContainedIn
 
-case class Choice(value: String)
+case class Choice(value: String) {
+  def toDeclarationType: Option[DeclarationType] = value match {
+    case Choice.AllowedChoiceValues.SupplementaryDec => Some(DeclarationType.SUPPLEMENTARY)
+    case Choice.AllowedChoiceValues.StandardDec      => Some(DeclarationType.STANDARD)
+    case _                                           => None
+  }
+}
```
ed0906: Highlighting a Key change

app/models/requests/JourneyRequest.scala
```diff
@@ -22,6 +22,6 @@ import play.api.mvc.WrappedRequest
 
 class JourneyRequest[A](val authenticatedRequest: AuthenticatedRequest[A], val cacheModel: ExportsDeclaration)
     extends AuthenticatedRequest[A](authenticatedRequest, authenticatedRequest.user) {
-  val choice: Choice = Choice(cacheModel.choice)
+  val choice: Choice = Choice(cacheModel.`type`)
```
ed0906: Arguably this could probably change too but I dont want this PR to grow any further

test/services/cache/ExportsDeclarationBuilder.scala
```diff
@@ -59,7 +60,7 @@ trait ExportsDeclarationBuilder {
 
   def withStatus(status: DeclarationStatus): ExportsDeclarationModifier = _.copy(status = status)
 
-  def withChoice(choice: String): ExportsDeclarationModifier = _.copy(choice = choice)
+  def withType(`type`: DeclarationType): ExportsDeclarationModifier = _.copy(`type` = `type`)
```
ed0906: Highlighting a Key change

app/connectors/exchange/ExportsDeclarationExchange.scala
```diff
@@ -31,7 +32,7 @@ case class ExportsDeclarationExchange(
   createdDateTime: Instant,
   updatedDateTime: Instant,
   sourceId: Option[String],
-  choice: String,
+  `type`: DeclarationType,
```
ed0906: Highlighting a Key change

app/controllers/ChoiceController.scala
ed0906: Yep, but it cant happen (see logic above)

ed0906:
> One small comment. LGTM, As I understand choice for know will work in the same way like before, it's just different type in exports declaration?

Yep, one minor change, instead of sending the "choice" to the backend as JSON
`"choice": "SMP"` it will be `"type":"SUPPLEMENTARY"`

A result of this is I'll need to clear the DB as the backend does the same

app/controllers/ChoiceController.scala
PatrykRudnicki: Yeah, I found it. What do you think about extracting this value before to pass it to the method without option?

PatrykRudnicki: I mean, instead of passing `Choice` to this method pass `DeclarationType`

app/controllers/ChoiceController.scala
```diff
@@ -94,15 +97,15 @@ class ChoiceController @Inject()(
       )
   }
 
-  private def updateChoice(id: String, choice: Choice)(implicit hc: HeaderCarrier) =
-    exportsCacheService.get(id).map(_.map(_.copy(`type` = choice.toDeclarationType.get))).flatMap {
+  private def updateDeclarationTyoe(id: String, `type`: DeclarationType)(implicit hc: HeaderCarrier) =
```
PatrykRudnicki: Typo in `updateDeclarationTyoe`

ed0906: You have a keen eye 👍

PR #577, opened 2019-10-22
CEDS-1839 Refactor Additional Declaration Type into an Enumeration

ed0906:


app/forms/declaration/additionaldeclarationtype/AdditionalDeclarationType.scala
```diff
@@ -16,12 +16,24 @@
...
+object AdditionalDeclarationType extends Enumeration {
+  type AdditionalDeclarationType = Value
+  implicit val format: Format[AdditionalDeclarationType.Value] =
+    Format(
+      Reads(
+        _.validate[String]
+          .filter(s => AdditionalDeclarationType.values.exists(_.toString == s))
+          .map(AdditionalDeclarationType.from(_).get)
+      ),
+      Writes(v => JsString(v.toString))
```
PatrykRudnicki: Proposal: Change `v` to `value` and `s` to `string` or `type` or `input`

app/utils/validators/forms/FieldValidator.scala
```diff
@@ -84,8 +84,8 @@ object FieldValidator {
 
   val startsWithCapitalLetter: String => Boolean = (input: String) => input.headOption.exists(_.isUpper)
 
-  val isContainedIn: Iterable[String] => String => Boolean =
-    (iterable: Iterable[String]) => (input: String) => iterable.exists(_ == input)
+  def isContainedIn[T](iterable: Iterable[T]): T => Boolean =
```
PatrykRudnicki: Can we add some test that will cover different type? Actually we have tests just for string. It'll be good to have something for some different this that obvious type.

ed0906: If it works for `String` surely it'd work for other types? it'd just be testing `iterable.exists(..)` works?

ed0906: Actually there already is in the AdditionalDeclarationType form tests

ed0906: As it first verifies what was entered was a valid enum value, then converts it to the enum, then verifies what was entered was a valid enum value for the options available.

Probably easier to look at the form mapping

PR #579, opened 2019-10-22
CEDS-1839 Change JourneyRequest.choice to DeclarationType

ed0906:


app/controllers/declaration/TransportContainerController.scala
```diff
@@ -175,6 +175,6 @@ class TransportContainerController @Inject()(
       navigator.continueTo(controllers.declaration.routes.TransportContainerController.displayContainerSummary(mode))
 
   private def allowSeals(implicit request: JourneyRequest[AnyContent]) =
-    request.choice.value == AllowedChoiceValues.StandardDec || request.choice.value == AllowedChoiceValues.SimplifiedDec
+    request.declarationType == DeclarationType.STANDARD || request.declarationType == DeclarationType.SIMPLIFIED
```
PatrykRudnicki: I'm not sure that it'll give us some value, but we can create methods inside `declarationType`, e.g. `isStandard` `isSimplified` etc, so in cases like here where we need boolean it'll be easier to use, I know that it can be helpful in the future because in a lot of place we will use it, but I know too that it's unnecessary for now. Up to you :)

ed0906: Yeh I did think about doing that as part of this but wasn't sure when it would be used (most usages of this method are inside `match` statements.

There's one or two I think a helper method will be useful for though, i'll change

PR #586, opened 2019-10-30
CEDS-1851 Skip 'Commodity measurements' for Simplified Journey

PauloPortugal:
A Simplified Declaration requires fewer data elements to be completed
than a Standard Declaration.
This change captures the requirement to skip the 'commodity-measure'
page, which is not relevant to Simplified Declaration.

### DIFF
--- a/app/controllers/declaration/PackageInformationController.scala
+++ b/app/controllers/declaration/PackageInformationController.scala
@@ -24,9 +24,8 @@ import forms.declaration.PackageInformation
 import forms.declaration.PackageInformation._
 import handlers.ErrorHandler
 import javax.inject.Inject
-import models.declaration.ExportItem
 import models.requests.JourneyRequest
-import models.{ExportsDeclaration, Mode}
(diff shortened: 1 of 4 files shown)


app/controllers/declaration/PackageInformationController.scala
```diff
@@ -84,19 +84,31 @@ class PackageInformationController @Inject()(
...
-            )
+              .successful(navigator.continueTo(nextPage(mode, itemId, request)))
       )
 
+  private def updateExportsCache(itemId: String, updatedCache: Seq[PackageInformation])(
+    implicit r: JourneyRequest[AnyContent]
+  ): Future[Option[ExportsDeclaration]] =
+    updateExportsDeclarationSyncDirect(model => {
+      val item: Option[ExportItem] = model.items
+        .find(item => item.id.equals(itemId))
```
bambuchaAdm: please use `updateItem` method from models - it would provide exactly this

PauloPortugal: Done.

PR #587, opened 2019-10-30
CEDS-1854 Skip Departure Transport on Simplified declaration

bambuchaAdm:


PauloPortugal:
Looks very interesting.
It will be useful as more branching occurs. It would also be interesting to send the user to the next page. 

**Pros:**
 - It will become more useful as more branching occurs
 - We could extend it to make decisions, based on user input as well

**Cons:**
 - There isn't _yet_ a need for it, as we simply the views but make other bits of the design more complex.
 - We do not have a business/technical need to look at this in a holistic manner

It would be good to listen to other people's viewpoints.

ed0906:
I approve of the general approach as I see this becoming more of a problem going forward

app/controllers/navigation/Navigator.scala
```diff
@@ -45,3 +49,28 @@ class Navigator @Inject()(appConfig: AppConfig, auditService: AuditService) {
...
+    case BorderTransport => controllers.declaration.routes.DepartureTransportController.displayPage
+    case _               => ???
+  }
+
+  val simplified: PartialFunction[DeclarationFieldCompanion, Mode => Call] = {
+    case BorderTransport => controllers.declaration.routes.WarehouseIdentificationController.displayPage
+    case _               => ???
+  }
+
+  def backLink(page: DeclarationFieldCompanion)(implicit request: JourneyRequest[_]): Mode => Call =
```
ed0906: I suspect we will need more than just knowledge of the page to go back, some pages will take into account previous answers etc

bambuchaAdm: I would like to do not make upfront work because we really don't know how to change it

PR #588, opened 2019-10-31
CEDS-1902 Disable auditing and graphite metrics

PatrykRudnicki:


conf/application.conf
```diff
@@ -100,13 +100,13 @@ metrics {
   name = ${appName}
   rateUnit = SECONDS
   durationUnit = SECONDS
-  showSamples = true
+  showSamples = false
   enabled = true
```
PatrykRudnicki: Those metrics are necessary for our unit tests. It's not graphite metrics

PR #591, opened 2019-11-04
CEDS-1675 Change format of Audit declation dates

PauloPortugal:
After the last TxM review, we were asked not to audit the `createDateTime`
and `updateDateTime` in the following MongoDB format:

```
"createdDateTime": {
   "$date": 1571135745223
 }
```

The approach is to have a serialisation strategy for TxM audit and
another one for MongoDB

### DIFF
--- a/test/services/audit/AuditServiceSpec.scala
+++ b/test/services/audit/AuditServiceSpec.scala
@@ -18,6 +18,7 @@ package services.audit
 
 import base.{Injector, TestHelper}
 import config.AppConfig
+import models.ExportsDeclaration.Audit.format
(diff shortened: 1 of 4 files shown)


test/services/audit/AuditServiceSpec.scala
```diff
@@ -18,6 +18,7 @@ package services.audit
 
 import base.{Injector, TestHelper}
 import config.AppConfig
+import models.ExportsDeclaration.Audit.format
```
ed0906: You sure the connector doesn't need this import too?

PauloPortugal: Yes. Without the explicit imports, we would have compilation failures

PR #592, opened 2019-11-04
CEDS-1720 Content for DMSREJ error messages

PatrykRudnicki:


app/services/model/RejectionReason.scala
```diff
@@ -62,6 +64,9 @@ object RejectionReason {
     }.getOrElse(Seq.empty)
   }
 
+  private def logMissingPointerMessageKey(pointer: Option[Pointer])(implicit messages: Messages): Unit =
```
ed0906:
```
error.pointer.filter { p =>
 val defined = messages.isDefinedAt(p.messageKey)
 if(!defined) {
    logger.warn("...")
 }
 defined
}
```

ed0906: Not much different but saves traversing the pointer list twice.

Up to you which you prefer

PatrykRudnicki: Good point

PatrykRudnicki: Fixed

PR #594, opened 2019-11-04
CEDS-1887 - update screens for selecting declaration type

SteveSugden:


app/config/AppConfig.scala
```diff
@@ -106,7 +107,14 @@ class AppConfig @Inject()(
     runModeConfiguration
       .getOptional[String]("list-of-available-journeys")
       .map(_.split(","))
-      .getOrElse(Array(Choice.AllowedChoiceValues.SupplementaryDec))
+      .getOrElse(Array(Choice.AllowedChoiceValues.Submissions))
```
PatrykRudnicki: I'm not sure that Submissions is a journey. This should be always on Choice screen

conf/application.conf
```diff
@@ -88,7 +88,8 @@ microservice {
   }
 }
 
-list-of-available-journeys = "SMP,STD,CAN,SUB,CON,SIM"
+list-of-available-journeys = "CRT,CAN,SUB,CON"
```
PatrykRudnicki: Similar here, I'm not sure that we should have Submission as available journey, journey and declaration is the same. Journey and choice is a different thing. Choice can have values that doesn't exists in journey, but it should have just a Choice that will allow it. I think in internal app @MaciejRewera added handling this. So the Submissions is choice that contains just redirection

app/controllers/declaration/DeclarationChoiceController.scala
```diff
@@ -0,0 +1,105 @@
...
+                Redirect(controllers.declaration.routes.DispatchLocationController.displayPage(Mode.Normal))
+                  .addingToSession(ExportsSessionKeys.declarationId -> created.id)
+              }
+          }
+        }
+      )
+  }
+
+  private def updateDeclarationType(id: String, `type`: DeclarationType)(implicit hc: HeaderCarrier) =
+    exportsCacheService.get(id).map(_.map(_.copy(`type` = `type`))).flatMap {
```
bambuchaAdm: it would be nice to move `_.copy(`type` = `type`)` as named method on `ExportsDeclaration`

conf/application.conf
SteveSugden: Depends what we mean by "journey" and "choice".  I took "list-of-available-journeys" to mean "options that are available on the Choice page" - so that we can turn them on or off.

Similar with "list-of-available-declarations" - we just want the ability to hide them via config.

I don't think we need to get hung-up on what is a journey

app/forms/declaration/DeclarationChoice.scala
```diff
@@ -0,0 +1,39 @@
...
+import play.api.data.Forms.text
+import play.api.data.{Form, Forms}
+import utils.validators.forms.FieldValidator.isContainedIn
+
+case class DeclarationChoice(value: DeclarationType)
+
+object DeclarationChoice {
+
+  val mapping = Forms.single(
+    "value" ->
```
bambuchaAdm: totaly trival - maybe `type` here would better show intention in form 
`type=Standard` looks little better then `value=Standard`

app/controllers/declaration/DeclarationChoiceController.scala
SteveSugden: Yes - it is a bit of a mouthful - I just moved it from the ChoiceController.

app/forms/declaration/DeclarationChoice.scala
SteveSugden: Probably right - another victim of copy the original form.  But I can change.

SteveSugden: Done

app/controllers/declaration/DeclarationChoiceController.scala
SteveSugden: Done - not sure it was worth it but it was low effort.

SteveSugden:
test this please

PR #597, opened 2019-11-05
CEDS-1851 Items summary page allows a Simplified journey item

PauloPortugal:
As part of the Simplified journey, we had to remove the commodity
measurements page from the item journey.
This change implies a change on the Items Summary validation rules for
the journey in question.

### DIFF
--- a/app/controllers/declaration/ItemsSummaryController.scala
+++ b/app/controllers/declaration/ItemsSummaryController.scala
@@ -22,6 +22,7 @@ import controllers.util.{Add, FormAction, SaveAndContinue, SaveAndReturn}
 import javax.inject.Inject
 import models.Mode
 import models.declaration.ExportItem
+import models.requests.JourneyRequest
 import play.api.data.FormError
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
@@ -49,7 +50,7 @@ class ItemsSummaryController @Inject()(
   //TODO Should we add validation for POST without items?
   def submit(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
     val action = FormAction.bindFromRequest()
-    val incorrectItems = buildIncorrectItemsErrors(request.cacheModel.items.toSeq)
+    val incorrectItems = buildIncorrectItemsErrors(request)
 
     action match {
       case Add =>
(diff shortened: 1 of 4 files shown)


app/controllers/declaration/ItemsSummaryController.scala
```diff
@@ -67,8 +68,8 @@ class ItemsSummaryController @Inject()(
     }
   }
 
-  private def buildIncorrectItemsErrors(items: Seq[ExportItem]): Seq[FormError] =
-    items.zipWithIndex.filterNot { case (item, _) => item.isCompleted }.map {
+  private def buildIncorrectItemsErrors(request: JourneyRequest[_]): Seq[FormError] =
```
PatrykRudnicki: I think it's better to use `JourneyRequest[AnyContent]` than `[_]`

app/models/declaration/ExportItem.scala
```diff
@@ -35,9 +37,15 @@ case class ExportItem(
...
-    procedureCodes.isDefined && isFiscalInformationCompleted && itemType.isDefined &&
-      packageInformation.nonEmpty && commodityMeasure.isDefined && additionalInformation.isDefined
+  val isCompleted: PartialFunction[DeclarationType, Boolean] = {
+    case DeclarationType.STANDARD | DeclarationType.SUPPLEMENTARY =>
+      procedureCodes.isDefined && isFiscalInformationCompleted && itemType.isDefined &&
+        packageInformation.nonEmpty && commodityMeasure.isDefined && additionalInformation.isDefined
+
+    case DeclarationType.SIMPLIFIED =>
+      procedureCodes.isDefined && isFiscalInformationCompleted && itemType.isDefined &&
+        packageInformation.nonEmpty && additionalInformation.isDefined
```
PatrykRudnicki: If we extract validation to method like `supplementaryDecValidation` or some kind `baseValidation` we can reuse it for Standard and Supplementary like `baseValidation && commodityMeasure.isDefined`.

Non blocking comment, I'm not sure about naming here and the value.

app/controllers/declaration/ItemsSummaryController.scala
PauloPortugal: Very good point! :)

bambuchaAdm: Here you just does not matter about `request.body` so `[_]` is more "any content" then `Request[AnyContent]` is.

app/models/declaration/ExportItem.scala
bambuchaAdm: Generally i would not share validation rules here between journeys  - even if they looks similar

PR #600, opened 2019-11-07
CEDS-1909 commodity details page split

SteveSugden:
Removes the commodity code and description fields from the 'Item Type' screen.
Also includes small refactoring of FiscalReference controller to make "back" navigation easier - caller does not need to know logic about fiscal references

### DIFF
--- a/app/controllers/declaration/CommodityDetailsController.scala
+++ b/app/controllers/declaration/CommodityDetailsController.scala
@@ -63,13 +63,9 @@ class CommodityDetailsController @Inject()(
   }
 
   private def updateExportsCache(itemId: String, updatedItem: CommodityDetails)(
-    implicit r: JourneyRequest[AnyContent]
+    implicit request: JourneyRequest[AnyContent]
   ): Future[Option[ExportsDeclaration]] =
-    updateExportsDeclarationSyncDirect(model => {
-      val itemList = model.items
-        .find(item => item.id.equals(itemId))
-        .map(_.copy(commodityDetails = Some(updatedItem)))
-        .fold(model.items)(model.items.filter(item => !item.id.equals(itemId)) + _)
-      model.copy(items = itemList)
-    })
+    updateExportsDeclarationSyncDirect { model =>
+      model.updatedItem(itemId, item => item.copy(commodityDetails = Some(updatedItem)))
+    }
 }
--- a/app/controllers/declaration/CommodityDetailsController.scala
+++ b/app/controllers/declaration/CommodityDetailsController.scala
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.declaration
+
+import controllers.actions.{AuthAction, JourneyAction}
+import controllers.navigation.Navigator
+import forms.declaration.CommodityDetails
+import forms.declaration.CommodityDetails.form
+import javax.inject.Inject
+import models.requests.JourneyRequest
+import models.{ExportsDeclaration, Mode}
+import play.api.data.Form
+import play.api.i18n.I18nSupport
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
+import services.cache.ExportsCacheService
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+import views.html.declaration.commodity_details
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class CommodityDetailsController @Inject()(
+  authenticate: AuthAction,
+  journeyType: JourneyAction,
+  override val exportsCacheService: ExportsCacheService,
+  navigator: Navigator,
+  mcc: MessagesControllerComponents,
+  commodityDetailsPage: commodity_details
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable {
+
+  def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    request.cacheModel.itemBy(itemId).flatMap(_.commodityDetails) match {
+      case Some(commodityDetails) => Ok(commodityDetailsPage(mode, itemId, form(request.declarationType).fill(commodityDetails)))
+      case _                      => Ok(commodityDetailsPage(mode, itemId, form(request.declarationType)))
+    }
+  }
+
+  def submitForm(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    form(request.declarationType)
+      .bindFromRequest()
+      .fold(
+        (formWithErrors: Form[CommodityDetails]) => Future.successful(BadRequest(commodityDetailsPage(mode, itemId, formWithErrors))),
+        validForm =>
+          updateExportsCache(itemId, validForm).map { _ =>
+            navigator
+              .continueTo(controllers.declaration.routes.ItemTypeController.displayPage(mode, itemId))
+        }
+      )
+  }
+
+  private def updateExportsCache(itemId: String, updatedItem: CommodityDetails)(
+    implicit r: JourneyRequest[AnyContent]
+  ): Future[Option[ExportsDeclaration]] =
+    updateExportsDeclarationSyncDirect(model => {
+      val itemList = model.items
+        .find(item => item.id.equals(itemId))
+        .map(_.copy(commodityDetails = Some(updatedItem)))
+        .fold(model.items)(model.items.filter(item => !item.id.equals(itemId)) + _)
+      model.copy(items = itemList)
+    })
+}
--- a/app/controllers/declaration/FiscalInformationController.scala
+++ b/app/controllers/declaration/FiscalInformationController.scala
@@ -22,7 +22,6 @@ import forms.declaration.FiscalInformation
 import forms.declaration.FiscalInformation.form
 import javax.inject.Inject
 import models.Mode
-import models.declaration.ExportItem
 import models.requests.JourneyRequest
 import play.api.data.Form
 import play.api.i18n.I18nSupport
@@ -43,11 +42,14 @@ class FiscalInformationController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
-  def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
-    request.cacheModel.itemBy(itemId) match {
-      case Some(ExportItem(_, _, _, Some(fiscalInformation), _, _, _, _, _, _)) =>
-        Ok(fiscalInformationPage(mode, itemId, form().fill(fiscalInformation)))
-      case response => Ok(fiscalInformationPage(mode, itemId, form()))
+  def displayPage(mode: Mode, itemId: String, fastForward: Boolean): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    if (fastForward && request.cacheModel.itemBy(itemId).exists(_.additionalFiscalReferencesData.exists(_.references.nonEmpty))) {
+      navigator.continueTo(routes.AdditionalFiscalReferencesController.displayPage(mode, itemId))
+    } else {
+      request.cacheModel.itemBy(itemId).flatMap(_.fiscalInformation) match {
+        case Some(fiscalInformation) => Ok(fiscalInformationPage(mode, itemId, form().fill(fiscalInformation)))
+        case _                       => Ok(fiscalInformationPage(mode, itemId, form()))
+      }
     }
   }
 
@@ -64,7 +66,7 @@ class FiscalInformationController @Inject()(
               }
             case FiscalInformation.AllowedFiscalInformationAnswers.no =>
               updateCacheForNo(itemId, formData) map { _ =>
-                navigator.continueTo(routes.ItemTypeController.displayPage(mode, itemId))
+                navigator.continueTo(routes.CommodityDetailsController.displayPage(mode, itemId))
               }
         }
       )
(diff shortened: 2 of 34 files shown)


test/views/declaration/spec/UnitViewSpec.scala
```diff
@@ -66,3 +67,26 @@ object MessagesKeyMatcher {
 object UnitViewSpec extends Injector {
   val realMessagesApi: MessagesApi = instanceOf[MessagesApi]
 }
+
+/*
+    Fails the test if a view is configured with a message key that doesnt exist in the messages file
+ */
+private class AllMessageKeysAreMandatoryMessages(msg: Messages) extends Messages {
```
PatrykRudnicki: Instead of comment I'll change the name to be better

app/controllers/declaration/CommodityDetailsController.scala
```diff
@@ -0,0 +1,75 @@
...
+
+  private def updateExportsCache(itemId: String, updatedItem: CommodityDetails)(
+    implicit r: JourneyRequest[AnyContent]
+  ): Future[Option[ExportsDeclaration]] =
+    updateExportsDeclarationSyncDirect(model => {
+      val itemList = model.items
+        .find(item => item.id.equals(itemId))
+        .map(_.copy(commodityDetails = Some(updatedItem)))
+        .fold(model.items)(model.items.filter(item => !item.id.equals(itemId)) + _)
+      model.copy(items = itemList)
```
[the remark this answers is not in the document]
SteveSugden: I just ported that method over from ItemTypeController.  I'll have a look to see if I can find a better way

app/controllers/declaration/CommodityDetailsController.scala
```diff
@@ -0,0 +1,75 @@
...
+  commodityDetailsPage: commodity_details
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable {
+
+  def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    request.cacheModel.itemBy(itemId).flatMap(_.commodityDetails) match {
+      case Some(commodityDetails) => Ok(commodityDetailsPage(mode, itemId, form(request.declarationType).fill(commodityDetails)))
+      case _                      => Ok(commodityDetailsPage(mode, itemId, form(request.declarationType)))
+    }
+  }
```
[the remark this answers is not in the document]
SteveSugden: Mine's only 3 lines though. :-)

app/forms/declaration/CommodityDetails.scala
```diff
@@ -0,0 +1,74 @@
...
+case class CommodityDetails(combinedNomenclatureCode: Option[String], descriptionOfGoods: String)
+
+object CommodityDetails extends DeclarationFieldCompanion {
+
+  implicit val format = Json.format[CommodityDetails]
+
+  val combinedNomenclatureCodeKey = "combinedNomenclatureCode"
+  val descriptionOfGoodsKey = "descriptionOfGoods"
+
+  val id = "CommodityDetails"
```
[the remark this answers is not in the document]
SteveSugden: Good question - just habit from copying old forms I guess

app/forms/declaration/CommodityDetails.scala
```diff
@@ -0,0 +1,74 @@
...
+  private val mappingOptionalCode: Mapping[CommodityDetails] =
+    Forms.mapping(combinedNomenclatureCodeKey -> mappingCombinedNomenclatureCodeOptional, descriptionOfGoodsKey -> mappingDescriptionOfGoods)(
+      CommodityDetails.apply
+    )(CommodityDetails.unapply)
+
+  def form(declarationType: DeclarationType): Form[CommodityDetails] = declarationType match {
+    case DeclarationType.SIMPLIFIED => Form(mappingOptionalCode)
+    case _                          => Form(mappingRequiredCode)
+  }
+}
```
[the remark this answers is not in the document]
SteveSugden: Why.  Reads ok to me

app/forms/declaration/CommodityDetails.scala
```diff
@@ -0,0 +1,74 @@
...
+    .verifying("declaration.commodityDetails.description.error.empty", nonEmpty)
+    .verifying("declaration.commodityDetails.description.error.length", isEmpty or noLongerThan(descriptionOfGoodsMaxLength))
+
+  private val mappingRequiredCode: Mapping[CommodityDetails] =
+    Forms.mapping(combinedNomenclatureCodeKey -> mappingCombinedNomenclatureCodeRequired, descriptionOfGoodsKey -> mappingDescriptionOfGoods)(
+      CommodityDetails.apply
+    )(CommodityDetails.unapply)
+
+  private val mappingOptionalCode: Mapping[CommodityDetails] =
+    Forms.mapping(combinedNomenclatureCodeKey -> mappingCombinedNomenclatureCodeOptional, descriptionOfGoodsKey -> mappingDescriptionOfGoods)(
```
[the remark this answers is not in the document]
SteveSugden: Formatter changes it back to one line.  Maybe if there were more fields?

app/views/declaration/commodity_details.scala.html
```diff
@@ -0,0 +1,93 @@
...
+
+        <details class="govuk-details" role="group">
+            <summary class="govuk-details__summary" role="button" aria-controls="details-content" aria-expanded="false">
+                <span class="govuk-details__summary-text">
+                @messages("declaration.commodityDetails.how.header")
+                </span>
+            </summary>
+            <div class="grey-border-left">
+                <ol class="list list-number">
+                @*<ol class="govuk-list ">*@
```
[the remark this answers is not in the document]
SteveSugden: Good spot - was experimenting with classes

conf/declaration.routes
```diff
@@ -130,7 +130,7 @@ POST        /items/:itemId/procedure-codes        controllers.declaration.Proced
 
 # Fiscal Information
 
-GET          /items/:itemId/fiscal-information    controllers.declaration.FiscalInformationController.displayPage(mode: models.Mode ?= models.Mode.Normal, itemId)
+GET          /items/:itemId/fiscal-information    controllers.declaration.FiscalInformationController.displayPage(mode: models.Mode ?= models.Mode.Normal, itemId, fastForward: Boolean ?= true)
```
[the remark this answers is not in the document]
SteveSugden: Lukasz can explain - was his idea.  Basically the idea is that we always go "back" to the question (yes/no) page.   If the cache contains data and FF is enabled (which it is by default) then the yes/no controller redirects to the next page (with the answers).  The only page that doesn't want this behaviour when going back is the answers page.

This means that the down-stream controller/view doesn't have to have conditional logic about which page to go back to.

test/views/declaration/spec/UnitViewSpec.scala
SteveSugden: Any suggestions.  Copied this from Internal UI.  I think the name is OK.  I'll get rid of the comment

app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -43,11 +42,14 @@ class FiscalInformationController @Inject()(
...
-        Ok(fiscalInformationPage(mode, itemId, form().fill(fiscalInformation)))
-      case response => Ok(fiscalInformationPage(mode, itemId, form()))
+  def displayPage(mode: Mode, itemId: String, fastForward: Boolean): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    if (fastForward && request.cacheModel.itemBy(itemId).exists(_.additionalFiscalReferencesData.exists(_.references.nonEmpty))) {
+      navigator.continueTo(routes.AdditionalFiscalReferencesController.displayPage(mode, itemId))
+    } else {
+      request.cacheModel.itemBy(itemId).flatMap(_.fiscalInformation) match {
+        case Some(fiscalInformation) => Ok(fiscalInformationPage(mode, itemId, form().fill(fiscalInformation)))
+        case _                       => Ok(fiscalInformationPage(mode, itemId, form()))
+      }
```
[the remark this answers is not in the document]
SteveSugden: Split the logic - does it help?

app/controllers/declaration/CommodityDetailsController.scala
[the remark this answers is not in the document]
SteveSugden: Found a more elegant way of doing it in AdditionalFiscalReferencesController

conf/messages.en
```diff
@@ -621,6 +612,24 @@ declaration.itemType.statisticalValue.error.empty = The statistical value cannot
 declaration.itemType.statisticalValue.error.length = Entered statistical value is too long
 declaration.itemType.statisticalValue.error.wrongFormat = Format of this value is incorrect
 
+declaration.commodityDetails.title = Commodity details
```
bambuchaAdm: @SteveSugden  what you think about start using `-` in names like `commodity-details` ? It looks very similar like in css and i think is more readable then camel case specially for longer names

conf/messages.en
```diff
@@ -621,6 +612,24 @@ declaration.itemType.statisticalValue.error.empty = The statistical value cannot
...
+declaration.commodityDetails.combinedNomenclatureCode.header.hint = A commodity code is an eight digit number specifically allocated to goods to classify them when moved outside the EU.  For example, 12345678.
+declaration.commodityDetails.combinedNomenclatureCode.error.empty = Commodity code cannot be empty
+declaration.commodityDetails.combinedNomenclatureCode.error.length = Enter a valid commodity code
+declaration.commodityDetails.combinedNomenclatureCode.error.specialCharacters = Commodity code cannot contain special characters
+declaration.commodityDetails.description.header = Describe the goods
+declaration.commodityDetails.description.header.hint = Try to be as specific as you can. For example, 'Copper pipe 15mm x 3m'
+declaration.commodityDetails.description.error.empty = Description cannot be empty
+declaration.commodityDetails.description.error.length = Description cannot be longer than 280 characters
+declaration.commodityDetails.how.header = How do I find a commodity code?
+declaration.commodityDetails.how.step1 = Go to the <a href='https://www.gov.uk/trade-tariff' target='_blank'>UK Trade Tariff</a>.
```
bambuchaAdm: Can we extract link for configuration ?

app/views/declaration/commodity_details.scala.html
```diff
@@ -0,0 +1,92 @@
...
+        <details class="govuk-details" role="group">
+            <summary class="govuk-details__summary" role="button" aria-controls="details-content" aria-expanded="false">
+                <span class="govuk-details__summary-text">
+                @messages("declaration.commodityDetails.how.header")
+                </span>
+            </summary>
+            <div class="grey-border-left">
+                <ol class="list list-number">
+                    <li>@Html(messages("declaration.commodityDetails.how.step1"))</li>
+                    <li>@Html(messages("declaration.commodityDetails.how.step2"))</li>
```
bambuchaAdm: only first one require to not be HTML escape - rest of those are text

conf/messages.en
SteveSugden: I've no preference - but this is the style used to-date in this file

app/views/declaration/commodity_details.scala.html
SteveSugden: I know.  Just wanted to be consistent.   I have a dream where one day a content designer can edit the message file in which case they could add links.   But I'll change if you don't like

SteveSugden:
Only difference in behaviour is for simplified decs (optional commodity code) and this is tested

conf/messages.en
SteveSugden: Can you inject config into messages file?

PatrykRudnicki:
Compilation failed

app/controllers/declaration/CommodityDetailsController.scala
```diff
@@ -0,0 +1,71 @@
...
+              .continueTo(controllers.declaration.routes.ItemTypeController.displayPage(mode, itemId))
+        }
+      )
+  }
+
+  private def updateExportsCache(itemId: String, updatedItem: CommodityDetails)(
+    implicit request: JourneyRequest[AnyContent]
+  ): Future[Option[ExportsDeclaration]] =
+    updateExportsDeclarationSyncDirect { model =>
+      model.updatedItem(itemId, item => item.copy(commodityDetails = Some(updatedItem)))
```
PatrykRudnicki: 👍

app/forms/declaration/CommodityDetails.scala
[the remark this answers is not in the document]
PatrykRudnicki: I mean very long lines with validation inside. Lines that contain more than one field mapping.
```
mapping(combinedNomenclatureCodeKey -> mappingCombinedNomenclatureCodeRequired, descriptionOfGoodsKey -> mappingDescriptionOfGoods)(
```
This above is one line. It's more readable, if you have every field in new line. One input form can be inlined, rest of them IMO are more readable in few lines.

👍 for extracting mappings to variable :)

test/views/declaration/CommodityDetailsViewSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+@ViewTest
+class CommodityDetailsViewSpec extends UnitViewSpec with ExportsTestData with Stubs with CommonMessages {
+
+  private val page = new commodity_details(mainTemplate, minimalAppConfig)
+  private val itemId = "item1"
+  private val realMessages = validatedMessages
+  private def createView(mode: Mode = Mode.Normal, itemId: String = itemId, declarationType: DeclarationType = DeclarationType.STANDARD): Document =
+    page(mode, itemId, CommodityDetails.form(declarationType))(journeyRequest(), realMessages)
+
+  def commodityDetailsView(declarationType: DeclarationType): Unit = {
```
PatrykRudnicki: Why this method contains whole specs inside? I can't find where is the closing bracket. But it looks like everything is in this method and this method is not invoked. View is created inside.

test/views/declaration/CommodityDetailsViewSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+import views.html.declaration.commodity_details
+import views.tags.ViewTest
+
+@ViewTest
+class CommodityDetailsViewSpec extends UnitViewSpec with ExportsTestData with Stubs with CommonMessages {
+
+  private val page = new commodity_details(mainTemplate, minimalAppConfig)
+  private val itemId = "item1"
+  private val realMessages = validatedMessages
+  private def createView(mode: Mode = Mode.Normal, itemId: String = itemId, declarationType: DeclarationType = DeclarationType.STANDARD): Document =
```
PatrykRudnicki: I don't see any value of creating `createView` method if we're using just one view and every argument is default. I'll just create here a view.

test/views/declaration/spec/UnitViewSpec.scala
PatrykRudnicki: I think it'll be better `AllMessageKeysAreMandatory`, but in practice we're not extending this very often so this name is not a problem.

test/views/declaration/CommodityDetailsViewSpec.scala
SteveSugden: This method is called by the three specs at the bottom of the file - once for each dec type.   Its a way of parameterising the tests that Lukasz showed me

test/views/declaration/CommodityDetailsViewSpec.scala
SteveSugden: The only param that was being used was dectype - I've removed the others

PR #602, opened 2019-11-08
CEDS-1906 Extract origination country to new page

PatrykRudnicki:


app/controllers/declaration/OriginationCountryController.scala
```diff
@@ -0,0 +1,71 @@
...
+    DestinationCountries
+      .form(OriginationCountryPage)
+      .bindFromRequest()
+      .fold(
+        formWithErrors => Future.successful(BadRequest(originationCountryPage(mode, formWithErrors))),
+        validCountry => {
+          val updatedCache =
+            request.cacheModel.locations.destinationCountries.getOrElse(DestinationCountries.empty()).copy(countryOfDispatch = validCountry)
+
+          updateExportsDeclarationSyncDirect(model => model.copy(locations = model.locations.copy(destinationCountries = Some(updatedCache)))).map {
```
bambuchaAdm: please move copy as method

app/views/declaration/destinationCountries/origination_country.scala.html
```diff
@@ -0,0 +1,60 @@
...
+
+    @components.heading(
+        title = messages("declaration.originationCountry.question"),
+        sectionHeader = messages("declaration.originationCountry.heading")
+    )
+
+    @helper.form(routes.OriginationCountryController.submit(), 'autoComplete -> "off") {
+        @helper.CSRF.formField
+
+        <div id="origination-country-field">
```
bambuchaAdm: why this id is on div not on `field_autocomplete`

test/views/declaration/destinationCountries/OriginationCountryViewSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+import unit.tools.Stubs
+import views.declaration.spec.UnitViewSpec
+import views.html.declaration.destinationCountries.origination_country
+
+class OriginationCountryViewSpec extends UnitViewSpec with Stubs {
+
+  val form: Form[String] = DestinationCountries.form(OriginationCountryPage)
+
+  val originationCountryPage = new origination_country(mainTemplate)
+  val view = originationCountryPage(Mode.Normal, form)
```
bambuchaAdm: Please test this for each declaration type

app/controllers/declaration/OriginationCountryController.scala
PatrykRudnicki: Changed

app/views/declaration/destinationCountries/origination_country.scala.html
PatrykRudnicki: It's the same like other usage of autocomplete input. I don't want to do it with this PR but I'll keep in mind to move id to autocomplete

test/views/declaration/destinationCountries/OriginationCountryViewSpec.scala
PatrykRudnicki: This view doesn't change and it's not related with any specific declaration type. I'll change behaviour of all of this, to align with specific journeys, this is just starting point for everything. It'll be added in the future

PR #605, opened 2019-11-08
CEDS-1950 New 'What is the warehouse identification number' screen

PauloPortugal:
* **Renamed WarehouseIdentificationController to WarehouseDetailsController**
* **New endpoint and controller for warehouse-identification**
* **Make the label optional for the `field_text` component**
As part of the new prototypes the field does not have a label, intead
the label is an `h1`, something we currently not support on field_text
Instead we use the heading and than an input text field underneath.
* **Add new warehouse-identification template**
  * The customs-declare-exports/declaration/warehouse screen will be split
   into multiple new screens.
  * All of these new screens are displayed for Standard and Supplementary
   declarations.
  * A subset of these is displayed for Simplified declarations.


app/controllers/declaration/WarehouseDetailsController.scala
```diff
@@ -0,0 +1,90 @@
...
+              inlandModeOfTransportCode = formData.inlandModeOfTransportCode
+          )
+        )
+        .getOrElse(
+          WarehouseDetails(
+            supervisingCustomsOffice = formData.supervisingCustomsOffice,
+            inlandModeOfTransportCode = formData.inlandModeOfTransportCode
+          )
+        )
+      model.copy(locations = model.locations.copy(warehouseIdentification = Some(warehouseDetails)))
```
bambuchaAdm: please move update/lens to model as method

app/views/components/fields/field_text.scala.html
```diff
@@ -18,7 +18,7 @@
 
 @(
     field: Field,
-    label: String,
+    label: Option[String] = None,
```
bambuchaAdm: @PauloPortugal what do you think abut use here empty string as null object pattern ?

conf/messages.en
```diff
@@ -424,6 +424,12 @@ standard.officeOfExit.circumstancesCode.hint = This is where the goods have been
 standard.officeOfExit.circumstancesCode.empty = Please provide an answer on this question
 standard.officeOfExit.circumstancesCode.error = Circumstance code is incorrect
 
+declaration.warehouseIdentification.sectionHeader = Locations
```
bambuchaAdm: Maybe `warehouse.identification` ?

test/views/declaration/WarehouseDetailsViewSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+      viewWithMessage.title() must include(viewWithMessage.getElementsByTag("h1").text())
+    }
+
+    "display 'Back' button that links to 'Warehouse Identification Number' page" in {
+      val backButton = view.getElementById("link-back")
+
+      backButton.text() mustBe "site.back"
+      backButton.getElementById("link-back") must haveHref(controllers.declaration.routes.WarehouseIdentificationController.displayPage(Mode.Normal))
+    }
+
```
bambuchaAdm: Please add test for having inputs

app/views/components/fields/field_text.scala.html
PatrykRudnicki: I agree with @bambuchaAdm . In 90% we will have some value here and this component handle empty string correctly.

test/views/declaration/WarehouseDetailsViewSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+  private def createView(mode: Mode = Mode.Normal, form: Form[WarehouseDetails] = form, messages: Messages = stubMessages()): Document =
+    page(mode, form)(journeyRequest(), messages)
+
+  "Warehouse Details View" should {
+    val view = createView()
+
+    "have proper messages for labels" in {
+      val messages = instanceOf[MessagesApi].preferred(journeyRequest())
+      messages must haveTranslationFor("supplementary.warehouse.title")
```
PatrykRudnicki: Hmm, you added few messages, but you're testing just one key. Can you add tests to check translations for all keys?

PauloPortugal: ooops!!!! My fault! Well spotted

PauloPortugal: We just renamed the view spec. These were all the messages we were testing. We will update those.

PauloPortugal: This has been addressed.

test/views/declaration/WarehouseDetailsViewSpec.scala
PauloPortugal: This page was simply renamed. It should be removed by the end of the sprint, so I am going to avoid spending time here.

conf/messages.en
PauloPortugal: This has been addressed.

app/views/components/fields/field_text.scala.html
PauloPortugal: We have decided to go with `Option` because the component signature has all the other text elements as optional as well.
We probably need to review this with the design/product team and understand if this will be the new way of creating text fields, i.e., having a section header, `h1` and a header page hint. 
If so we will have to create a new component and probably then we would avoid optional fields and use just Strings.

app/controllers/declaration/WarehouseDetailsController.scala
PauloPortugal: Good point. We are reusing the model and the next story we will have to assess our approach as we might find the limitations of our use. If so we will then address this concern there. THank you

PR #609, opened 2019-11-11
CEDS-1911 CUS code page extraction

SteveSugden:


app/views/declaration/un_dangerous_goods_code.scala.html
```diff
@@ -34,7 +34,7 @@
 
 @hintHtml = {
     @messages("declaration.unDangerousGoodsCode.header.hint")
-    <a target="_blank" href=@appConfig.tradeTariffUrl>@messages("declaration.unDangerousGoodsCode.header.hint.link")</a>
+    <a target="_blank" href=@appConfig.tradeTariffUrl>@messages("declaration.unDangerousGoodsCode.header.hint.link")</a>.
```
PatrykRudnicki: Is this `dot` at the end necessary? Can we move it to the messages?

SteveSugden: The prototype has the full-stop outside of the link.   I thought of adding it to the message but I'm sure someone would complain

PatrykRudnicki: Yeah, in that case I'll pass link to the messages as argument and put this `dot` in the messages. Probably in movement frontend you'll find examples. It's better to keep it in messages. I know that this will looks weird, but it will be safe, alone dot can looks like mistake

app/forms/declaration/CUSCode.scala
```diff
@@ -0,0 +1,72 @@
...
+      hasCusCodeKey -> requiredRadio("error.yesNo.required"),
+      cusCodeKey -> mandatoryIfEqual(
+        hasCusCodeKey,
+        yes,
+        text()
+          .verifying("declaration.cusCode.error.empty", nonEmpty)
+          .verifying("declaration.cusCode.error.length", isEmpty or hasSpecificLength(cusCodeLength))
+          .verifying("declaration.cusCode.error.specialCharacters", isEmpty or isAlphanumeric)
+      )
+    )(form2Model)(model2Form)
```
ed0906: 👍 for not storing the "yes/no" boolean in the model

app/models/declaration/ExportItem.scala
```diff
@@ -31,6 +31,7 @@ case class ExportItem(
   itemType: Option[ItemType] = None,
   commodityDetails: Option[CommodityDetails] = None,
   dangerousGoodsCode: Option[UNDangerousGoodsCode] = None,
+  cusCode: Option[CUSCode] = None,
```
ed0906: Could cusCode not just be an `Option[String]`?

SteveSugden: It could.  So could UNDangerousCode.   CommodityDetails could be a Tuple.   I thought we wanted a case-class to encapsulate the values on each page/form?

app/views/declaration/un_dangerous_goods_code.scala.html
SteveSugden: Good idea.   I've re-factored all of the links I've done over the last three pages.

app/models/declaration/ExportItem.scala
ed0906: Only if it helps, otherwise it's a little unneccessary

ed0906: If CommodityDetails has two fields then thats fine

SteveSugden: Isn't it better to be consistent?

PatrykRudnicki: Two fields should be in one class, for easier managing this. One field IMO can be in class, but in practice you can build form without model. I'm using this for DestinationCountries. So I stayed with backend model, but I'm using just `Form[String]` to manage form. Here CUSCode can be used like String, so it's okay, but `UNDangerousCode. CommodityDetails could be a Tuple` I'll leave it as it is. Or if `UNDangerousCode` can just one String or `CommodityDetails` can one String we can change it to not using model. But if you have place in model where you need to put `Tuple` I'll prefer to stay with class.

SteveSugden: Tuple wasn't a serious suggestion!  :-) .  I'm going to stick with what I've got.   As mentioned, it's more consistent if a page has a Form[Model] even if that Model could be a String.

test/forms/declaration/CUSCodeSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      "provided with valid input when user said Yes" in {
+        val form = CUSCode.form.bind(formData("Yes", Some("1234ABCD")))
+
+        form.hasErrors must be(false)
+      }
+
+      "provided with no input when user said No" in {
+        val form = CUSCode.form.bind(formData("No", None))
+
+        form.hasErrors must be(false)
```
bambuchaAdm: I think there is exist somehting like `FormMatchers` and there is matcher for `haveErrors` or something similar

test/views/declaration/CUSCodeViewSpec.scala
```diff
@@ -0,0 +1,98 @@
...
+    "we are on Simplified journey" should {
+      behave like cusCodeView(DeclarationType.SIMPLIFIED)
+    }
+  }
+
+  "UN Dangerous Goods Code View on populated page" when {
+    val dangerousGoodsCode = Some(CUSCode(Some("12345678")))
+
+    "we are on Standard journey" should {
+      behave like cusCodeView(DeclarationType.STANDARD, dangerousGoodsCode)
```
bambuchaAdm: :+1:

test/unit/controllers/declaration/CUSControllerSpec.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  val mockPage = mock[cus_code]
+
+  val controller =
+    new CUSCodeController(mockAuthAction, mockJourneyAction, mockExportsCacheService, navigator, stubMessagesControllerComponents(), mockPage)(ec)
+
+  override protected def beforeEach(): Unit = {
+    super.beforeEach()
+    authorizedUser()
+    withNewCaching(aDeclaration(withType(DeclarationType.STANDARD)))
```
bambuchaAdm: Is this page is only for standard ?

PR #612, opened 2019-11-12
CEDS-1953 New 'Where is the supervising customs office?' page

janeturner:
Created new Supervising Customs Office controller and view.
Added endpoint to declaration.routes.
Removed the Supervising Customs Office field from Warehouse Details page.
Changed Back button on Warehouse Details page and 'Save and Continue' button on Warehouse Identification page to direct to Supervising Customs Office page.
Added View Spec and Controller Spec for Supervising Customs Office, and updated Warehouse Identification Controller spec and Warehouse Details View Spec.

### DIFF
--- a/app/controllers/declaration/SupervisingCustomsOfficeController.scala
+++ b/app/controllers/declaration/SupervisingCustomsOfficeController.scala
@@ -19,26 +19,25 @@ package controllers.declaration
 import controllers.actions.{AuthAction, JourneyAction}
 import controllers.navigation.Navigator
 import forms.declaration.WarehouseDetails
-import forms.declaration.WarehouseDetails.form
 import javax.inject.Inject
 import models.requests.JourneyRequest
-import models.{DeclarationType, ExportsDeclaration, Mode}
+import models.{ExportsDeclaration, Mode}
 import play.api.data.Form
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import services.cache.ExportsCacheService
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
-import views.html.declaration.{supervising_customs_office, warehouse_identification}
+import views.html.declaration.supervising_customs_office
 
 import scala.concurrent.{ExecutionContext, Future}
 
 class SupervisingCustomsOfficeController @Inject()(
-                                                    authenticate: AuthAction,
-                                                    journeyType: JourneyAction,
-                                                    navigator: Navigator,
-                                                    override val exportsCacheService: ExportsCacheService,
-                                                    mcc: MessagesControllerComponents,
-                                                    supervisingCustomsOfficePage: supervising_customs_office
+  authenticate: AuthAction,
+  journeyType: JourneyAction,
+  navigator: Navigator,
+  override val exportsCacheService: ExportsCacheService,
+  mcc: MessagesControllerComponents,
+  supervisingCustomsOfficePage: supervising_customs_office
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
@@ -72,13 +71,9 @@ class SupervisingCustomsOfficeController @Inject()(
               identificationNumber = dbWarehouseDetails.identificationNumber,
               supervisingCustomsOffice = formData.supervisingCustomsOffice,
               inlandModeOfTransportCode = dbWarehouseDetails.inlandModeOfTransportCode
-            )
-        )
-        .getOrElse(
-          WarehouseDetails(
-            supervisingCustomsOffice = formData.supervisingCustomsOffice
           )
         )
+        .getOrElse(WarehouseDetails(supervisingCustomsOffice = formData.supervisingCustomsOffice))
       model.copy(locations = model.locations.copy(warehouseIdentification = Some(warehouseDetails)))
     }
 }
--- a/app/controllers/declaration/SupervisingCustomsOfficeController.scala
+++ b/app/controllers/declaration/SupervisingCustomsOfficeController.scala
@@ -57,14 +57,8 @@ class SupervisingCustomsOfficeController @Inject()(
       .fold(
         (formWithErrors: Form[WarehouseDetails]) => Future.successful(BadRequest(supervisingCustomsOfficePage(mode, formWithErrors))),
         form => {
-          val nextStep = request.cacheModel.`type` match {
-            case DeclarationType.STANDARD | DeclarationType.SUPPLEMENTARY =>
-              controllers.declaration.routes.DepartureTransportController.displayPage(mode)
-            case DeclarationType.SIMPLIFIED =>
-              controllers.declaration.routes.BorderTransportController.displayPage(mode)
-          }
           updateCache(form)
-            .map(_ => navigator.continueTo(nextStep))
+            .map(_ => navigator.continueTo(controllers.declaration.routes.WarehouseDetailsController.displayPage(mode)))
         }
       )
   }
--- a/app/controllers/declaration/SupervisingCustomsOfficeController.scala
+++ b/app/controllers/declaration/SupervisingCustomsOfficeController.scala
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.declaration
+
+import controllers.actions.{AuthAction, JourneyAction}
+import controllers.navigation.Navigator
+import forms.declaration.WarehouseDetails
+import forms.declaration.WarehouseDetails.form
+import javax.inject.Inject
+import models.requests.JourneyRequest
+import models.{DeclarationType, ExportsDeclaration, Mode}
+import play.api.data.Form
+import play.api.i18n.I18nSupport
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
+import services.cache.ExportsCacheService
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+import views.html.declaration.{supervising_customs_office, warehouse_identification}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class SupervisingCustomsOfficeController @Inject()(
+                                                    authenticate: AuthAction,
+                                                    journeyType: JourneyAction,
+                                                    navigator: Navigator,
+                                                    override val exportsCacheService: ExportsCacheService,
+                                                    mcc: MessagesControllerComponents,
+                                                    supervisingCustomsOfficePage: supervising_customs_office
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable {
+
+  import forms.declaration.WarehouseDetails._
+
+  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    request.cacheModel.locations.warehouseIdentification match {
+      case Some(data) => Ok(supervisingCustomsOfficePage(mode, form().fill(data)))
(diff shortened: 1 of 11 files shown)


app/controllers/declaration/SupervisingCustomsOfficeController.scala
```diff
@@ -0,0 +1,79 @@
...
+        .map(
+          dbWarehouseDetails =>
+            WarehouseDetails(
+              identificationNumber = dbWarehouseDetails.identificationNumber,
+              supervisingCustomsOffice = formData.supervisingCustomsOffice,
+              inlandModeOfTransportCode = dbWarehouseDetails.inlandModeOfTransportCode
+          )
+        )
+        .getOrElse(WarehouseDetails(supervisingCustomsOffice = formData.supervisingCustomsOffice))
+      model.copy(locations = model.locations.copy(warehouseIdentification = Some(warehouseDetails)))
```
PatrykRudnicki: Can we extract logic about copying to exports declaration model?

app/views/declaration/supervising_customs_office.scala.html
```diff
@@ -0,0 +1,58 @@
...
+@(mode: Mode, form: Form[WarehouseDetails])(implicit request: Request[_], messages: Messages)
+
+    @main_template(
+        title = Title("declaration.warehouse.supervisingCustomsOffice.title")
+    ) {
+
+        @helper.form(routes.SupervisingCustomsOfficeController.submit(mode), 'autoComplete -> "off") {
+            @helper.CSRF.formField
+
+            @components.back_link(controllers.declaration.routes.WarehouseIdentificationController.displayPage(mode), mode)
```
PatrykRudnicki: Should we use Navigator for back link?

app/controllers/declaration/SupervisingCustomsOfficeController.scala
```diff
@@ -0,0 +1,79 @@
...
+      case Some(data) => Ok(supervisingCustomsOfficePage(mode, form().fill(data)))
+      case _          => Ok(supervisingCustomsOfficePage(mode, form()))
+    }
+  }
+
+  def submit(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    form()
+      .bindFromRequest()
+      .fold(
+        (formWithErrors: Form[WarehouseDetails]) => Future.successful(BadRequest(supervisingCustomsOfficePage(mode, formWithErrors))),
```
bambuchaAdm: do we have to get here type annotation ?

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+import views.html.declaration.supervising_customs_office
+
+class SupervisingCustomsOfficeControllerSpec extends ControllerSpec with BeforeAndAfterEach with WarehouseIdentificationMessages with OptionValues {
+
+  val supervisingCustomsOfficeTemplate: supervising_customs_office = mock[supervising_customs_office]
+
+  val controller = new SupervisingCustomsOfficeController(
+    authenticate = mockAuthAction,
+    journeyType = mockJourneyAction,
+    navigator,
```
PatrykRudnicki: If for every argument you're using structure `field = something` you should follow this rule here too.

app/controllers/declaration/SupervisingCustomsOfficeController.scala
```diff
@@ -0,0 +1,79 @@
...
+        form => {
+          updateCache(form)
+            .map(_ => navigator.continueTo(controllers.declaration.routes.WarehouseDetailsController.displayPage(mode)))
+        }
+      )
+  }
+
+  private def updateCache(formData: WarehouseDetails)(implicit request: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] =
+    updateExportsDeclarationSyncDirect { model =>
+      val warehouseDetails = model.locations.warehouseIdentification
```
bambuchaAdm: what do you think about moving this updater to `Locations`

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+
+  "Supervising Customs Office Controller on GET request" should {
+    "return 200 OK" in {
+      withNewCaching(supplementaryCacheModel)
+      val response = controller.displayPage(Mode.Normal).apply(getRequest())
+      status(response) must be(OK)
+    }
+
+    "read item from cache and display it" in {
+      withNewCaching(supplementaryCacheModel)
```
PatrykRudnicki: Separating logic in tests using empty line make code more readable.

```
setup()

runningWhatYouWantToTest

checkingConditions
```

test/views/declaration/SupervisingCustomsOfficeViewSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+      val messages = instanceOf[MessagesApi].preferred(journeyRequest())
+      messages must haveTranslationFor("declaration.warehouse.supervisingCustomsOffice.sectionHeader")
+      messages must haveTranslationFor("declaration.warehouse.supervisingCustomsOffice.title")
+      messages must haveTranslationFor("declaration.warehouse.supervisingCustomsOffice.hint")
+      messages must haveTranslationFor("declaration.warehouse.supervisingCustomsOffice.error")
+
+    }
+
+    "display same page title as header" in {
+      val viewWithMessage = createView(messages = realMessagesApi.preferred(request))
```
PatrykRudnicki: Here you're using `realMessagesApi` above you're injecting a different one. To be consistent we can use one approach. In practice more of those tests are using the real one, so I think we can put real in creating view method.

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
```diff
@@ -0,0 +1,192 @@
...
+import org.scalatest.{BeforeAndAfterEach, OptionValues}
+import play.api.libs.json.Json
+import play.api.test.Helpers._
+import play.twirl.api.HtmlFormat
+import unit.base.ControllerSpec
+import views.html.declaration.supervising_customs_office
+
+class SupervisingCustomsOfficeControllerSpec extends ControllerSpec with BeforeAndAfterEach with WarehouseIdentificationMessages with OptionValues {
+
+  val supervisingCustomsOfficeTemplate: supervising_customs_office = mock[supervising_customs_office]
```
PatrykRudnicki: Inside tests specially for small scope like this there is no value of putting return type for `vals` (for `def` it's always good to add return type). Moreover I don't see value of using word `Template` in this name. It's just make things longer

PatrykRudnicki:
Tests passed, build failed on scala formatter. Please run formatter `sbt scalafmt test:scalafmt`. This command will reformat whole codebase :)

app/controllers/declaration/SupervisingCustomsOfficeController.scala
janeturner: I've changed this

app/controllers/declaration/SupervisingCustomsOfficeController.scala
janeturner: I've spoken to Paulo about the model which is going to be reviewed. At the moment 3 pages are using the same model, but there will be another ticket to have new models for each page.

app/controllers/declaration/SupervisingCustomsOfficeController.scala
janeturner: I've spoken to Paulo about this and the model is going to be reviewed in another ticket, at the moment we are using the same model for three pages.

app/views/declaration/supervising_customs_office.scala.html
janeturner: this is another issue with the model which will be part of another story

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
janeturner: When i remove the return type I get a Type Mismatch error `Required: String, Found: supervising_customs_office` so i've kept it for now unless there's another way around this.
I've removed `Template` from the name

janeturner: When I changed the name and removed the return type I got a Type Mismatch `Required: String, Found: supervising_customs_office`.
I've removed the return type but kept the name the same otherwise I get errors, unless there's another way around this.

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
janeturner: This has been addressed

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
janeturner: This has been changed

test/views/declaration/SupervisingCustomsOfficeViewSpec.scala
janeturner: This has been changed

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
```diff
@@ -0,0 +1,199 @@
...
+import org.mockito.Mockito
+import org.mockito.Mockito.{verify, when}
+import org.scalatest.{BeforeAndAfterEach, OptionValues}
+import play.api.libs.json.Json
+import play.api.test.Helpers._
+import play.twirl.api.HtmlFormat
+import unit.base.ControllerSpec
+import views.html.declaration.supervising_customs_office
+
+class SupervisingCustomsOfficeControllerSpec extends ControllerSpec with BeforeAndAfterEach with WarehouseIdentificationMessages with OptionValues {
```
ed0906: What does `OptionValues` give us? I've not seen that one before

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
```diff
@@ -0,0 +1,199 @@
...
+
+  val supervisingCustomsOfficeTemplate: supervising_customs_office = mock[supervising_customs_office]
+
+  val controller = new SupervisingCustomsOfficeController(
```
ed0906: We have some fields marked as `private` and some not. Lets be consistent :) 
They probably all want to be `private`

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
```diff
@@ -0,0 +1,199 @@
...
+      val response = controller.displayPage(Mode.Normal).apply(getRequest())
+
+      status(response) must be(OK)
+    }
+
+    "read item from cache and display it" in {
+      withNewCaching(supplementaryCacheModel)
+
+      val result = controller.displayPage(Mode.Normal).apply(getRequest())
+      await(result)
```
ed0906: no need for these to be on separate lines, await returns the content of the finished future so it's better to do

`val result = await(controller.whatever)`

also, no need to explicitly write `.apply` just do `controller.displayPage(Mode.Normal)(getRequest())`

ed0906: This applies to all tests in this class

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
PatrykRudnicki: `OptionValues` gives method `value` which extract the value from Option or throw failed test exception.

`val result: String = Option("aaaa").value`

We use it in places where we're dealing with Options.

```
/**
     * Returns the value contained in the wrapped <code>Option</code>, if defined, else throws <code>TestFailedException</code> with
     * a detail message indicating the option was not defined.
     */
    def value: T = {
      try {
        opt.get
      }
      catch {
        case cause: NoSuchElementException => 
          throw new TestFailedException((_: StackDepthException) => Some(Resources.optionValueNotDefined), Some(cause), pos)
      }
    }
```

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
PatrykRudnicki: Yeah, the problem is because you extended `WarehouseIdentificationMessages` and you have a variable `supervisingCustomsOffice` in scope. So return type can be removed, but it's minor comment. Actual it's not a mistake

test/unit/controllers/declaration/SupervisingCustomsOfficeControllerSpec.scala
janeturner: This has been changed

PR #614, opened 2019-11-13
CEDS-1907 Align model with backend

PatrykRudnicki:


app/controllers/declaration/DestinationCountriesController.scala
```diff
@@ -50,139 +47,89 @@ class DestinationCountriesController @Inject()(
...
-              destinationCountriesStandardPage(
-                mode,
-                Standard.form(data.countryOfDispatch, data.countryOfDestination).fill(data),
-                data.countriesOfRouting
-              )
-            )
-        }
-      case None => Redirect(controllers.declaration.routes.OriginationCountryController.displayPage(mode))
+    request.declarationType match {
+      case DeclarationType.SUPPLEMENTARY => Redirect(routes.DestinationCountryController.displayPage())
```
ed0906: Is this an infinite loop?

ed0906: We probably need a `ReturnToStartException` for this kind of thing

PatrykRudnicki: `DestinationCountryController` is different controller. I know stupid naming, but this Controller will disappear in next PR, so I decided to don't refactor this. Whole this controller now support just one input `countriesOfRouting` and will be changed and renamed

app/models/declaration/Locations.scala
```diff
@@ -23,17 +23,19 @@ import models.ExportsDeclaration
 import play.api.libs.json.Json
 
 case class Locations(
-  destinationCountries: Option[DestinationCountries] = None,
+  originationCountry: Option[String] = None,
```
ed0906: `origination` is not a word.... :p

PR #616, opened 2019-11-14
CEDS-1914 - NACT code extraction

SteveSugden:
This is very similar to the recent extraction of the TARIC codes page.

The main difference is that removing NACT codes from the ItemType page now leaves that page with only one text field and I was able to simplify the ItemTypeController to remove all of the add/remove item functionality.

The next and final ticket in this series will be to rename ItemType to StatisticalValue and to skip that page when on a simplified dec journey.


app/forms/declaration/ItemType.scala
```diff
@@ -0,0 +1,51 @@
...
+import play.api.libs.json.Json
+import utils.validators.forms.FieldValidator._
+
+case class ItemType(statisticalValue: String)
+
+object ItemType {
+
+  implicit val format = Json.format[ItemType]
+
+  val statisticalValueKey = "statisticalValue"
```
PatrykRudnicki: You don't need this key

app/forms/declaration/NactCode.scala
```diff
@@ -0,0 +1,43 @@
...
+import services.NationalAdditionalCode
+import utils.validators.forms.FieldValidator._
+
+case class NactCode(nactCode: String)
+
+object NactCode extends DeclarationPage {
+
+  implicit val format = Json.format[NactCode]
+
+  val nactCodeKey = "nactCode"
```
PatrykRudnicki: You don't need this key

conf/messages.en
```diff
@@ -640,14 +633,25 @@ declaration.cusCode.error.empty = Enter a CUS code
...
+declaration.taricAdditionalCodes.header = Is there an additional TARIC code?
 declaration.taricAdditionalCodes.header.hint = This will be the four digits that come after the commodity code in the {0}. If no additional codes are required, leave blank.
+declaration.taricAdditionalCodes.header.tradeTariff.link = UK Trade Tariff
 declaration.taricAdditionalCodes.error.empty = Enter a TARIC code
 declaration.taricAdditionalCodes.error.length = TARIC code must be exactly 4 characters long
 declaration.taricAdditionalCodes.error.specialCharacters = TARIC code cannot contain specialCharacters
 declaration.taricAdditionalCodes.add.hint = Add TARIC code
 declaration.taricAdditionalCodes.remove.hint = Remove TARIC code {0}
 
+declaration.nationalAdditionalCode.header = Do these goods require a national additional (NACT) code?
```
PatrykRudnicki: But in the UI we're using `National additional code` so we can use this as variable. but it's okay

conf/messages.en
```diff
@@ -640,14 +633,25 @@ declaration.cusCode.error.empty = Enter a CUS code
...
+declaration.nationalAdditionalCode.header.hint = You can find this in ‘Export measures and restrictions’ tab in the {0}.  If no additional code is required, leave blank.
```
PatrykRudnicki: Double space in this message

test/unit/controllers/declaration/ItemTypeControllerSpec.scala
```diff
@@ -97,112 +95,42 @@ class ItemTypeControllerSpec extends ControllerSpec with ErrorHandlerMocks with
...
-        verify(mockItemTypePage, times(0)).apply(any(), any(), any(), any())(any(), any())
-      }
-
-      "correct item type is added during saving" in {
+      "valid data is posted" in {
 
         withNewCaching(aDeclaration(withItem(anItem(withItemId(itemId)))))
 
-        val correctForm = Seq(("nationalAdditionalCode", "VATR"), ("statisticalValue", "1234"), saveAndContinueActionUrlEncoded)
+        val badData = Json.toJson(ItemType("7"))
```
PatrykRudnicki: `wrongData` sounds better?

conf/messages.en
PatrykRudnicki: After {0}. and If

SteveSugden: You not part of the double-space-after-full-stop debate?   I think the browser ignores it anyway but I'll remove.

app/forms/declaration/NactCode.scala
SteveSugden: These keys are used as constants for field name in the view and tests

conf/messages.en
PatrykRudnicki: No, I hate double spaces everywhere :)

SteveSugden: That's because you are young and have never used a real typewriter (I suspect)

bambuchaAdm:
Please add temper monkey scripts

PR #620, opened 2019-11-14
CEDS-1968 - rename ItemType to StatisticalValue

SteveSugden:
And skip page for Simplified Decs


test/unit/controllers/declaration/NactCodeControllerSpec.scala
```diff
@@ -171,6 +171,12 @@ class NactCodeControllerSpec extends ControllerSpec with ErrorHandlerMocks with
 
       def controllerRedirectsToNextPage(decType: DeclarationType): Unit =
         "user submits valid code" in {
+
+          def nextPage = decType match {
+            case DeclarationType.SIMPLIFIED => controllers.declaration.routes.PackageInformationController.displayPage(Mode.Normal, "itemId")
```
bambuchaAdm: maybe paramter for test page

app/controllers/navigation/Navigator.scala
```diff
@@ -60,25 +60,45 @@ object Navigator {
     case DestinationCountryPage => controllers.declaration.routes.OriginationCountryController.displayPage
     case _                      => throw new IllegalArgumentException("Navigator back-link route not implemented")
   }
+  val standardItemPage: PartialFunction[DeclarationPage, (Mode, String) => Call] = {
```
bambuchaAdm: Maybe it time to make little wrapper over string for `ItemId`

app/views/declaration/package_information.scala.html
```diff
@@ -42,7 +43,7 @@
     title = Title("supplementary.packageInformation.title", "supplementary.items")
 ) {
 
-    @components.back_link(Navigator.backLinkItemPage(PackageInformation)(request)(mode, itemId), mode)
+    @components.back_link(Navigator.backLink(PackageInformation, mode, ItemId(itemId)))
```
ed0906: You could pass ItemId into the view

And maybe add it to the model so the routes with an item-id are type safe :D

PR #624, opened 2019-11-19
CEDS-1955 Clean up the Warehouse model

PauloPortugal:
As part of splitting the Warehouse Details page into three different
pages, we have tried to reuse the model. This became a problem for
several reasons:
 - Convoluted code on the update cache methods
 - Only one attribute belonged to the Warehouse type
 - Different approach from the rest of the controllers/views

With that in mind we decided to create one single model per page/view as
per our previous approach and keep the model clean.
This change also forced a code refactor on the BE service and on the
Acceptance tests suite.

Relates to : CEDS-1950, CEDS-1953, CEDS-1954


app/controllers/declaration/InlandTransportDetailsController.scala
```diff
@@ -68,18 +68,6 @@ class InlandTransportDetailsController @Inject()(
...
-              identificationNumber = dbWarehouseDetails.identificationNumber,
-              supervisingCustomsOffice = dbWarehouseDetails.supervisingCustomsOffice,
-              inlandModeOfTransportCode = formData.inlandModeOfTransportCode
-          )
-        )
-        .getOrElse(WarehouseDetails(inlandModeOfTransportCode = formData.inlandModeOfTransportCode))
-      model.copy(locations = model.locations.copy(warehouseIdentification = Some(warehouseDetails)))
-    }
+  private def updateCache(formData: InlandModeOfTransportCode)(implicit request: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] =
+    updateExportsDeclarationSyncDirect(model => model.copy(locations = model.locations.copy(inlandModeOfTransportCode = Some(formData))))
```
bambuchaAdm: please move update logic to method on model

test/forms/declaration/WarehouseIdentificationSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+        JsObject(Map("identificationNumber" -> JsString(TestHelper.createRandomAlphanumericString(36))))
+
+      form().bind(incorrectWarehouseDetails).errors.map(_.message) must contain(identificationNumberError)
+    }
+
+    "validate correct identification number" in {
+      val incorrectWarehouseDetails: JsValue =
+        JsObject(Map("identificationNumber" -> JsString(TestHelper.createRandomAlphanumericString(34))))
+
+      form().bind(incorrectWarehouseDetails).errors.size mustBe 0
```
bambuchaAdm: there is matcher for `From[_]` in `FormMatchers` - `withoutErrors`

PauloPortugal: Good shout! 
Done.

app/controllers/declaration/InlandTransportDetailsController.scala
PauloPortugal: Do we have examples of this being done in our project?

app/controllers/declaration/InlandTransportDetailsController.scala
```diff
@@ -41,20 +40,21 @@ class InlandTransportDetailsController @Inject()(
...
-    request.cacheModel.locations.warehouseIdentification match {
+    request.cacheModel.locations.inlandModeOfTransportCode match {
       case Some(data) => Ok(inlandTransportDetailsPage(mode, form().fill(data)))
       case _          => Ok(inlandTransportDetailsPage(mode, form()))
     }
   }
 
   def submit(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
-    form()
+    InlandModeOfTransportCode
```
PatrykRudnicki: I think you're importing everything and using form without object above. This is unnecessary and can make wrong thinking about form in 47 line. Is it the same?

app/controllers/declaration/InlandTransportDetailsController.scala
PatrykRudnicki: Inside `ExportsDeclaration` model you can find it

PR #625, opened 2019-11-21
CEDS-1919 update final screen

SteveSugden:


app/models/responses/FlashKeys.scala
```diff
@@ -18,5 +18,6 @@ package models.responses
 
 object FlashKeys {
   val lrn: String = "LRN"
+  val decType: String = "DECTYPE"
```
PatrykRudnicki: I'll use as value `DecType` the same approach like `expiryDate` not `lrn`

SteveSugden: Good point.  I should be able to change them all to camelCase.   Will (hopefully) flush out anywhere that is not using these constants (like confirmation page wasn't)

SteveSugden: 2nd thought - initial case.  And "LRN" is probably OK as it's an abbreviation.

PatrykRudnicki: Yeah, `lrn` is correct

test/views/declaration/SubmissionConfirmationPageViewSpec.scala
```diff
@@ -30,65 +31,51 @@ import views.tags.ViewTest
...
-      messages must haveTranslationFor("supplementary.confirmation.explanation.linkText")
-      messages must haveTranslationFor("supplementary.confirmation.submitAnotherDeclaration")
-      messages must haveTranslationFor("supplementary.confirmation.rejection.header")
-    }
-
-    "display header" in {
-      view.select("article>div.govuk-box-highlight>h1").text() mustBe "supplementary.confirmation.header"
-      view.select("article>div.govuk-box-highlight>p").text() mustBe "-"
+    "display header with default" in {
+      val highlightBox = view.selectFirst("article>div.govuk-box-highlight")
```
bambuchaAdm: quite agresive matcher

PR #631, opened 2019-11-21
CEDS-1907 Implementaiton for change button in countries summary

PatrykRudnicki:


app/views/declaration/destinationCountries/country_of_routing.scala.html
```diff
@@ -42,17 +42,14 @@
...
-            @components.fields.field_autocomplete(
-                form("country"),
-                "",
-                None,
-                None,
-                messages(s"declaration.${page.id}.empty"),
-                AutoCompleteItem.fromCountry(Countries.allCountries, _.countryCode)
-            )
-        </div>
+        @components.fields.field_autocomplete(
```
SteveSugden: Was the div with id not useful for tests?

PatrykRudnicki: I'm checking this now, if yes, I'll fix tests. Unit tests don't use it

PatrykRudnicki: `field_autocomplete` has his own id based on the field name, so this should be enough

PR #632, opened 2019-11-22
CEDS-2018 Make additional information optional

PatrykRudnicki:


app/models/declaration/ExportItem.scala
```diff
@@ -45,11 +45,10 @@ case class ExportItem(
   val isCompleted: PartialFunction[DeclarationType, Boolean] = {
     case DeclarationType.STANDARD | DeclarationType.SUPPLEMENTARY =>
       procedureCodes.isDefined && isFiscalInformationCompleted && statisticalValue.isDefined &&
-        packageInformation.nonEmpty && commodityMeasure.isDefined && additionalInformation.isDefined
+        packageInformation.nonEmpty && commodityMeasure.isDefined
```
SteveSugden: So additionalInformation is not required for ANY dec type.  Is that correct?

PatrykRudnicki: Yes, confirmed with @bethany-kish

PR #634, opened 2019-11-22
CEDS-1908 Fix for changing countries

PatrykRudnicki:


app/controllers/declaration/RoutingCountriesSummaryController.scala
```diff
@@ -122,7 +122,7 @@ class RoutingCountriesSummaryController @Inject()(
 
   def submitChangeCountry(mode: Mode, countryToChange: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
     val cachedCountries = request.cacheModel.locations.routingCountries
-    val countriesForValidation = cachedCountries.filter(_ == countryToChange)
+    val countriesForValidation = cachedCountries.filterNot(_ == countryToChange)
```
SteveSugden: So countriesForValidation are the existing countries in the cache except the one you want to change.  Makes sense.

PatrykRudnicki: Yes, to check duplication in easier way

test/unit/controllers/declaration/RoutingCountriesSummaryControllerSpec.scala
```diff
@@ -281,4 +281,18 @@ class RoutingCountriesSummaryControllerSpec extends ControllerSpec {
...
+
+  "Routing countries controller" should {
+
+    "handle duplication during changing the country" in {
+
+      withNewCaching(aDeclaration(withRoutingCountries(Seq("PL", "DZ", "AD"))))
+
+      val duplicatedCountry = JsObject(Seq("country" -> JsString("PL")))
+
+      val result = controller.submitChangeCountry(Mode.Normal, "DZ")(postRequest(duplicatedCountry))
```
SteveSugden: So in this request, DZ is the country I want to change, and I want to change it to PL (which is already in the list - hence error).  Is that right?

PatrykRudnicki: Yes, exactly. I can add some local variables to describe those countries if you want

PR #635, opened 2019-11-25
CEDS-1957 - extract payment and has containers from border transport

SteveSugden:
This PR extracts both the "transport payment" and "has containers" question from the "border transport". In addition, the contain data is re-located into the new "TransportData" top level of the cache model.
Also fixed an in-correct use of the radio choice component which was accepting message keys rather than messages.


app/controllers/declaration/TransportContainerController.scala
```diff
@@ -44,33 +45,52 @@ class TransportContainerController @Inject()(
...
+    } else {
+
+      ContainerYesNo.form
+        .bindFromRequest()
+        .fold(
+          (formWithErrors: Form[ContainerYesNo]) => Future.successful(BadRequest(addFirstPage(mode, formWithErrors))),
+          validForm =>
+            validForm.id match {
+              case Some(containerId) => saveFirstContainer(mode, containerId)
+              case _                 => successful(navigator.continueTo(routes.SummaryController.displayPage(Mode.Normal)))
```
PatrykRudnicki: I prefer to have whole `Future.successful` not just `successful`.

Comment to whole file

app/models/declaration/TransportData.scala
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.declaration
+import forms.declaration.TransportPayment
```
PatrykRudnicki: Style comment, if you want you can fix it.

Empty line that separate package and import makes code more readable. You don't need to think where imports start.

app/controllers/declaration/TransportContainerController.scala
```diff
@@ -165,13 +184,15 @@ class TransportContainerController @Inject()(
   private def containerId(values: Seq[String]): String = values.headOption.getOrElse("")
 
   private def updateCache(updatedContainers: Seq[Container])(implicit r: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] =
-    updateExportsDeclarationSyncDirect(model => model.copy(containerData = Some(TransportInformationContainerData(updatedContainers))))
+    updateExportsDeclarationSyncDirect(
+      model => model.copy(transportData = Some(model.transportData.getOrElse(new TransportData).copy(containers = updatedContainers)))
```
bambuchaAdm: can we move this `copy` to model ?

app/models/declaration/TransportData.scala
```diff
@@ -0,0 +1,38 @@
...
+      Seq(updatedContainer)
+    } else if (!containers.exists(_.id == updatedContainer.id)) {
+      containers :+ updatedContainer
+    } else {
+      containers.map {
+        case container if updatedContainer.id == container.id => updatedContainer
+        case otherContainer                                   => otherContainer
+      }
+    }
+}
```
PatrykRudnicki: I think it'll be better and easier to understand when second if `!containers.exists(_.id == updatedContainer.id)` will be without negation and contains the last logic, and last `else` will have logic for new containers.

You'll be able to not use another nested `if` in last case

app/controllers/declaration/TransportPaymentController.scala
```diff
@@ -0,0 +1,73 @@
...
+      case Some(data) => Ok(transportPayment(mode, form().fill(data)))
+      case _          => Ok(transportPayment(mode, form()))
+    }
+  }
+
+  def submitForm(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    form()
+      .bindFromRequest()
+      .fold(
+        (formWithErrors: Form[TransportPayment]) => Future.successful(BadRequest(transportPayment(mode, formWithErrors))),
```
bambuchaAdm: why this type annotation ?

app/controllers/declaration/TransportPaymentController.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  private def nextPage(mode: Mode, borderTransport: Option[BorderTransport])(implicit request: JourneyRequest[AnyContent]): Result =
+    navigator.continueTo(controllers.declaration.routes.TransportContainerController.displayContainerSummary(mode))
+
+  private def updateCache(formData: TransportPayment)(implicit r: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] = {
+    def update(model: ExportsDeclaration) = {
+      val data = model.transportData.getOrElse(TransportData())
+      model.copy(transportData = Some(data.copy(transportPayment = Some(formData))))
+    }
+    updateExportsDeclarationSyncDirect(model => update(model))
```
bambuchaAdm: can we put just `update` or even inline this local function ?

docs/Customs Declare Exports AutoComplete.js
```diff
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         Customs Declare Exports AutoComplete
 // @namespace    http://tampermonkey.net/
-// @version      1.9
+// @version      2.0
```
PatrykRudnicki: I think it'll be better to stay with `1.10`. We changed this with `major` change last time, but there is no point to upgrade `major` version. Of course it's not a mistake

app/forms/declaration/ContainerYesNo.scala
```diff
@@ -0,0 +1,70 @@
...
+import forms.DeclarationPage
+import forms.Mapping.requiredRadio
+import forms.declaration.ContainerAdd.maxContainerIdLength
+import play.api.data.Forms.text
+import play.api.data.{Form, Forms}
+import play.api.libs.json.Json
+import uk.gov.voa.play.form.ConditionalMappings.mandatoryIfEqual
+import utils.validators.forms.FieldValidator.{isAlphanumeric, noLongerThan, nonEmpty}
+
+case class ContainerYesNo(id: Option[String])
```
bambuchaAdm: I am not sure about this name. Name suggest `Boolean` but it contains id.

SteveSugden: Renamed to 'ContainerFirst'.  It's now better aligned with the view that uses it.

app/controllers/declaration/TransportContainerController.scala
SteveSugden: Moved.  Also created similar method for updating TransportPayment

app/controllers/declaration/TransportPaymentController.scala
SteveSugden: No reason - removed

app/controllers/declaration/TransportPaymentController.scala
SteveSugden: Done

app/controllers/declaration/TransportContainerController.scala
SteveSugden: Done

app/models/declaration/TransportData.scala
SteveSugden: scalafmt didn't fix that for me.  Shame!

app/models/declaration/TransportData.scala
SteveSugden: Method was copied over from a different class. but I've re-structured it

docs/Customs Declare Exports AutoComplete.js
SteveSugden: Done

app/models/declaration/TransportData.scala
```diff
@@ -15,21 +15,24 @@
...
     if (containers.isEmpty) {
       Seq(updatedContainer)
-    } else if (!containers.exists(_.id == updatedContainer.id)) {
-      containers :+ updatedContainer
-    } else {
+    } else if (hasContainer(updatedContainer.id)) {
       containers.map {
         case container if updatedContainer.id == container.id => updatedContainer
         case otherContainer                                   => otherContainer
       }
```
PatrykRudnicki: Another imprevements, if you retrieve index of the container to update there is a method `updated` which will take a index and new element and update it for you. There should be less iterations and less cases to check.

Method `indexWhere` with condition `_.id == updatedContainer.id`. This will give us `index` and `containers.updated(index, updatedContainer`.

Just a propose

PR #638, opened 2019-11-26
CEDS-1960 Skip Border Transport for Simplified declaration

bambuchaAdm:


PatrykRudnicki:
@SteveSugden There is removed BorderTransport from Navigator

test/views/declaration/BorderTransportViewSpec.scala
```diff
@@ -139,15 +139,15 @@ class BorderTransportViewSpec extends UnitViewSpec with ExportsTestData with Stu
...
   private def createView(mode: Mode = Mode.Normal, form: Form[BorderTransport] = form, request: JourneyRequest[_] = journeyRequest()): Document =
     page(mode, form)(request, realMessages)
 
-  "TransportDetails View" when {
-    "we are on Standard journey" should {
-      val requrestOnStandard = journeyRequest(DeclarationType.STANDARD)
+  "TransportDetails View" must {
 
-      val view = createView(request = requrestOnStandard)
+    onStandard { request =>
```
ed0906: I'm not a fan of this style

I'd prefer if we just added a method to request e.g. `request.asStandard` or something

test/views/declaration/BorderTransportViewSpec.scala
```diff
@@ -159,9 +159,8 @@ class BorderTransportViewSpec extends UnitViewSpec with ExportsTestData with Stu
       behave like havingMeansOfTransport(view)
     }
 
-    "we are on Supplementary journey" should {
-      val requrestOnSupplementary = journeyRequest(DeclarationType.SUPPLEMENTARY)
-      val view = createView(request = requrestOnSupplementary)
+    onSupplementary { request =>
```
ed0906: Similar, I'd prefer if this was less invasive e.g. `request.asSupplementary`

test/views/declaration/BorderTransportViewSpec.scala
ed0906: I'm open to seeing what the team thinks though

SteveSugden: I'm not sure what's going on here.  What's "onStandard" etc?

PauloPortugal: IT would be good to explain to us about this approach.

test/views/declaration/spec/UnitViewSpec.scala
```diff
@@ -43,6 +47,36 @@ class UnitViewSpec extends UnitSpec with ViewMatchers {
   }
 
   def messagesKey(key: String): BeMatcher[String] = new MessagesKeyMatcher(key)
```
PauloPortugal: Same here. It is a different pattern of what we are used to.
It would be good to understand the benefits.

test/views/declaration/BorderTransportViewSpec.scala
bambuchaAdm: This is support method on ViewSpec that would add prefix for registration test case part of `onAllJourneys` and `onJourney`

bambuchaAdm:
We have lot of conditional logic on view but that was untested. Testing for each case generated lot of boilerplate for same behavior tests. methods `onAllJourneys`, `onJourney` and others are shortcut for defining for each declaration type case and preserve consistency in test names.

test/views/declaration/spec/UnitViewSpec.scala
```diff
@@ -43,6 +47,31 @@ class UnitViewSpec extends UnitSpec with ViewMatchers {
...
+
+  def onEveryDeclarationJourney(f: JourneyRequest[_] => Unit): Unit =
+    onJourney(DeclarationType.STANDARD, DeclarationType.SUPPLEMENTARY, DeclarationType.SIMPLIFIED)(f)
+
+  def onJourney(types: DeclarationType*)(f: JourneyRequest[_] => Unit): Unit =
```
SteveSugden: I get it now.  This would have been good to have a while ago.   I'm wondering how many of the existing tests I need to replicate for the new OCCASIONAL dec type.

PR #639, opened 2019-11-26
CEDS-1956 update border transport page

SteveSugden:
Took the opportunity to rename the rest of the transport message keys as discussed at last techtro.


app/forms/declaration/BorderTransport.scala
```diff
@@ -34,21 +35,118 @@ object BorderTransport extends DeclarationPage {
...
+        model.meansOfTransportCrossingTheBorderType,
+        model2Ref(IMOShipIDNumber),
+        model2Ref(NameOfVessel),
+        model2Ref(WagonNumber),
+        model2Ref(VehicleRegistrationNumber),
+        model2Ref(IATAFlightNumber),
+        model2Ref(AircraftRegistrationNumber),
+        model2Ref(EuropeanVesselIDNumber),
+        model2Ref(NameOfInlandWaterwayVessel)
+    )
```
PatrykRudnicki: Why logic here is so complicated?

app/views/declaration/summary/transport_containers_section.scala.html
```diff
@@ -19,12 +19,12 @@
 
 @(containers: Seq[Container])(implicit messages: Messages)
 
-@components.summary_list(Some(messages("declaration.transportInfo.containers.title"))) {
+@components.summary_list(Some(messages("declaration.transportInformation.containers.title"))) {
```
bambuchaAdm: @PatrykRudnicki here could be conflict with your ticket

PatrykRudnicki: I removed this section, so it's not a big deal

app/forms/declaration/BorderTransport.scala
SteveSugden: Yes it is a bit fiddly.  It's because we have 8 different input fields now - all optional except for the one that is associated with the selected radio option.

I started by having all 8 bound to the same form field.  That works but when you change radio option the previously typed value is visible.  Plus the validation for each type field might be different (it isn't at the moment but that could change)

PR #642, opened 2019-11-27
CEDS-1976 Fix Quick Submit button

PauloPortugal:
As part of the GDS Design system the styling has changed and broke the
quick submit button on the choice page.


PauloPortugal:
Version bumped! :) Thanks for spotting it!

PR #643, opened 2019-11-27
CEDS-1886 New summary page

PatrykRudnicki:


app/forms/Ducr.scala
```diff
@@ -21,6 +21,7 @@ import play.api.data.Forms.{mapping, text}
 import play.api.data.validation.Constraints.pattern
 import play.api.libs.json.Json
 
+// TODO change argument name from 'ducr' to 'value', this change is incompatible with backend
```
ed0906: Should this be done now?

app/forms/common/Address.scala
```diff
@@ -31,6 +31,8 @@ case class Address(
 ) {
   def isDefined(): Boolean =
     fullName.nonEmpty || addressLine.nonEmpty || townOrCity.nonEmpty || postCode.nonEmpty || country.nonEmpty
+
+  def extractValues: Seq[String] = Seq(fullName, addressLine, townOrCity, postCode, country)
```
ed0906: Is this an `unapply` in disguise?

PatrykRudnicki: Some kind, I don't have better idea how to handle some things. In this case I need to have a sequence to pass to the view. Our components handle how to display `String` or `Seq[String]`, it makes everything in next row. Probably naming for the method here and in all different models are wrong. I'm open for any idea how to do it in different way

app/forms/declaration/BorderTransport.scala
```diff
@@ -29,7 +30,12 @@ case class BorderTransport(
   meansOfTransportCrossingTheBorderNationality: Option[String],
   meansOfTransportCrossingTheBorderType: String,
   meansOfTransportCrossingTheBorderIDNumber: String
-)
+) {
+
+  def extractModeValue: String = TransportCodes.extractBorderTransportValue(meansOfTransportCrossingTheBorderType)
```
ed0906: cant we just  call this method  `mode`?

app/forms/declaration/BorderTransport.scala
```diff
@@ -29,7 +30,12 @@ case class BorderTransport(
...
+
+  def transportLeavingBorder: Seq[String] = Seq(extractModeValue, meansOfTransportCrossingTheBorderIDNumber)
```
ed0906: Seems a little strange to return a `Seq`

app/forms/declaration/BorderTransport.scala
PatrykRudnicki: 👍

app/forms/declaration/BorderTransport.scala
PatrykRudnicki: Any proposal how to change name or implement it different?

app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -28,6 +29,15 @@ case class DeclarationAdditionalActors(eori: Option[String], partyType: Option[S
 
   def toJson: JsValue = Json.toJson(this)(DeclarationAdditionalActors.format)
 
+  def extractPartyType: String =
```
ed0906: Sounds like a method that should belong on  `PartyType`

e.g. `PartyType.toString`

ed0906: Then you can  do `additionalActors.partyType.map(_.toString).getOrElse("")`

PatrykRudnicki: Good point, will move it

app/forms/declaration/DepartureTransport.scala
```diff
@@ -23,10 +23,19 @@ import play.api.data.Forms.{mapping, optional, text}
...
 import utils.validators.forms.FieldValidator.{isContainedIn, noLongerThan, _}
 
-case class DepartureTransport(borderModeOfTransportCode: String, meansOfTransportOnDepartureType: String, meansOfTransportOnDepartureIDNumber: String)
+case class DepartureTransport(
+  borderModeOfTransportCode: String,
+  meansOfTransportOnDepartureType: String,
+  meansOfTransportOnDepartureIDNumber: String
+) {
+
+  val extractModeValue: String = TransportCodes.extractModeOfTransportValue(borderModeOfTransportCode)
```
ed0906: Can we just call this `mode`?

app/forms/declaration/DepartureTransport.scala
```diff
@@ -23,10 +23,19 @@ import play.api.data.Forms.{mapping, optional, text}
...
+  val extractTypeMode: String = TransportCodes.extractBorderTransportValue(meansOfTransportOnDepartureType)
```
ed0906: Can we just call this `typeMode` or similar

app/forms/declaration/DepartureTransport.scala
```diff
@@ -23,10 +23,19 @@ import play.api.data.Forms.{mapping, optional, text}
...
+
+  def transportDetails: Seq[String] = Seq(extractTypeMode, meansOfTransportOnDepartureIDNumber)
```
ed0906: Once again, seems a little strange to return a `Seq[String]`

app/forms/declaration/DepartureTransport.scala
```diff
@@ -56,6 +65,18 @@ object TransportCodes {
   val InlandWaterway = "8"
   val Unknown = "9"
 
+  val extractModeOfTransportValue: String => String = _ match {
```
ed0906: I'd prefer if we made this an enum with a `code` and a `name`

app/forms/declaration/DepartureTransport.scala
```diff
@@ -65,6 +86,18 @@ object TransportCodes {
   val EuropeanVesselIDNumber = "80"
   val NameOfInlandWaterwayVessel = "81"
 
+  val extractBorderTransportValue: String => String = _ match {
```
ed0906: Similar, I'd prefer if we had an enum for this

app/forms/declaration/InlandModeOfTransportCode.scala
```diff
@@ -23,7 +23,21 @@ import play.api.data.{Form, Forms}
 import play.api.libs.json.Json
 import utils.validators.forms.FieldValidator.isContainedIn
 
-case class InlandModeOfTransportCode(inlandModeOfTransportCode: Option[String] = None)
+case class InlandModeOfTransportCode(inlandModeOfTransportCode: Option[String] = None) {
+  import TransportCodes._
+
+  def extractModeValue: Option[String] = inlandModeOfTransportCode.map {
```
ed0906: Enummmmm

app/forms/declaration/NatureOfTransaction.scala
```diff
@@ -21,7 +21,22 @@ import play.api.data.{Form, Forms, Mapping}
 import play.api.libs.json.{Json, OFormat}
 import utils.validators.forms.FieldValidator._
 
-case class NatureOfTransaction(natureType: String)
+case class NatureOfTransaction(natureType: String) {
+  import NatureOfTransaction._
+
+  def extractNatureType: String = natureType match {
```
ed0906: Enummmmm

app/forms/declaration/RepresentativeDetails.scala
```diff
@@ -21,7 +21,15 @@ import play.api.data.{Form, Forms}
 import play.api.libs.json.Json
 import utils.validators.forms.FieldValidator.isContainedIn
 
-case class RepresentativeDetails(details: Option[EntityDetails], statusCode: Option[String])
+case class RepresentativeDetails(details: Option[EntityDetails], statusCode: Option[String]) {
+
+  def extractStatusCodeValue: Option[String] = statusCode.map {
```
ed0906: Enummmm

app/forms/declaration/TransportPayment.scala
```diff
@@ -21,7 +21,19 @@ import play.api.data.{Form, Mapping}
 import play.api.libs.json.{Json, OFormat}
 import utils.validators.forms.FieldValidator.isContainedIn
 
-case class TransportPayment(paymentMethod: Option[String] = None)
+case class TransportPayment(paymentMethod: Option[String] = None) {
+
+  def extractPaymentType: Option[String] = paymentMethod.map {
```
ed0906: Enummmmm

app/forms/declaration/additionaldeclarationtype/AdditionalDeclarationType.scala
```diff
@@ -38,4 +38,11 @@ object AdditionalDeclarationType extends Enumeration {
   val SIMPLIFIED_PRE_LODGED = Value("F")
 
   def from(string: String): Option[AdditionalDeclarationType] = AdditionalDeclarationType.values.find(_.toString == string)
+
+  def asText(additionalDeclarationType: AdditionalDeclarationType): String = additionalDeclarationType match {
```
ed0906: You know what I'm going to say here :p

app/models/ExportsDeclaration.scala
```diff
@@ -50,6 +50,8 @@ case class ExportsDeclaration(
 
   def isComplete: Boolean = status == DeclarationStatus.COMPLETE
 
+  def capitalizedType: String = `type`.toString().toLowerCase().capitalize
```
ed0906: Seems a bit of a strange method to have on the model, Is it used in multiple places? if not lets just do this logic in the view/wherever

ed0906: I think we should avoid cluttering the models with helper methods that are only used in one place. It doesn't feel very Object Oriented.

app/forms/declaration/InlandModeOfTransportCode.scala
PatrykRudnicki: Yeah, it was old implementation, I just added possibility to retrieve text based on the code.

app/models/declaration/ProcedureCodesData.scala
```diff
@@ -24,6 +24,8 @@ case class ProcedureCodesData(procedureCode: Option[String], additionalProcedure
   def toProcedureCode(): ProcedureCodes = ProcedureCodes(procedureCode, None)
 
   def containsAdditionalCode(code: String): Boolean = additionalProcedureCodes.contains(code)
+
+  def summaryValue: String = procedureCode.fold("")(code => (Seq(code) ++ additionalProcedureCodes).mkString(" "))
```
ed0906: As above, I'd prefer we didn't clutter the model with helper methods that don't naturally belong there. If the summary page is the only place that needs it, lets just stick the code there.

*Unless* every model will have a `summaryValue` method, in which case lets create a `trait` for it

app/models/declaration/TransportInformation.scala
```diff
@@ -23,6 +23,8 @@ case class TransportInformation(transportPayment: Option[TransportPayment] = Non
 
   private def hasContainer(id: String) = containers.exists(_.id == id)
 
+  val hasContainers: String = if (containers.nonEmpty) "Yes" else "No"
```
ed0906: You would naturally expect a method called `hasContainers` to return a `Boolean`. The summary page can change that to a Yes/No if required

app/services/Countries.scala
```diff
@@ -45,7 +45,13 @@ object Countries {
 
   val countryCodeMap: Map[String, Country] = countries.map(country => (country.countryCode, country)).toMap
 
-  def retrieveCountryNameFromCode(code: String): Country = countryCodeMap(code)
+  val countryNameMap: Map[String, Country] = countries.map(country => (country.countryName, country)).toMap
+
+  def retrieveCountryFromCode(code: String): Country = countryCodeMap(code)
```
ed0906: `findByCode`

Should probably also return an `Option[Country]` but that's up to you

app/services/Countries.scala
```diff
@@ -45,7 +45,13 @@ object Countries {
...
+
+  def retrieveCountryFromName(name: String): Country = countryNameMap(name)
```
ed0906: Similar

app/services/Countries.scala
```diff
@@ -45,7 +45,13 @@ object Countries {
...
+
+  def retrieveCountryFromCode(code: Option[String]): Option[Country] = code.map(countryCodeMap(_))
```
ed0906: Why do we need this?
Surely the methods above cover it

app/services/DocumentType.scala
```diff
@@ -18,7 +18,10 @@ package services
 
 import utils.JsonFile
 
-case class DocumentType(description: String, code: String)
+case class DocumentType(description: String, code: String) {
+
+  def asText: String = description + " - " + code
```
ed0906: `toString`

app/services/DocumentType.scala
```diff
@@ -27,4 +30,11 @@ object DocumentType {
   val allDocuments: List[DocumentType] = JsonFile
     .readFromJsonFile("/code-lists/document-type-autocomplete-list.json", deserialiser)
     .sortBy(_.description.toLowerCase)
+
+  val documentCodesMap: Map[String, DocumentType] =
+    allDocuments.map(documentType => (documentType.code, documentType)).toMap
+
+  def retrieveDocumentTypeFromCode(code: String): DocumentType = documentCodesMap(code)
```
ed0906: `findByCode`

app/services/DocumentType.scala
```diff
@@ -27,4 +30,11 @@ object DocumentType {
...
+
+  def retrieveDocumentTypesFromCodes(codes: Seq[String]): Seq[DocumentType] = codes.map(documentCodesMap(_))
```
ed0906: `findByCodes`

app/services/PackageTypes.scala
```diff
@@ -30,4 +30,8 @@ object PackageTypes {
       case regex(code: String, description: String) =>
         PackageType(code, description)
     }.sortBy(_.description)
+
+  val packageTypeCodeMap: Map[String, PackageType] = all.map(packageType => (packageType.code, packageType)).toMap
+
+  def getPackageType(code: String): PackageType = packageTypeCodeMap(code)
```
ed0906: `findByCode`

ed0906:
My proposal, to avoid all the summary helper methods that return `Seq` etc cluttering up the model is:

Lets try and represent the Summary Components using some models
```
case class SummarySection(heading: String, subSections: Seq[SummarySubSection])
case class SummarySubSection(heading: Option[String], items: Seq[SummarySectionItem])
case class SummarySectionItem(label: String, value: String, changeLink: Call)
```

And we could build the content in a `SummaryBuilder` class e.g.
```
class SummaryBuilder {
    def build(declaration: ExportsDeclaration):  Seq[SummarySection] =Seq(
        thisDeclarationSection
    )
    
    private def thisDeclarationSection: SummarySection = SummarySection(
        heading = "This Declaration",
        subSections = Seq(
            SummarySubSection(
                heading = None, 
                items = Seq(
                    SummarySectionItem("The declaration type", "Standard", routes.SomeController.display(Mode.Review))
                )
        )
    )
}
```

Benefits:
- The builder can be responsible for getting the data in the right format (so no `summaryValue` helper methods)
- The mapping can be tested in isolation (and not in a view test)
- The view can just become a matter of rendering some html based on a `Seq[SummarySection]` (much easier to view test)

app/forms/declaration/DeclarationAdditionalActors.scala
bambuchaAdm: Generaly `toString` in domain code is not good thing unless is stable and fully defined - for enum 'https://www.scala-lang.org/api/current/scala/Enumeration.html' we don't have this case

app/models/ExportsDeclaration.scala
bambuchaAdm: But on the other hand it hide internal object representation to "function" Contract allow to change `type` field without breaking external clients

app/models/declaration/TransportInformation.scala
bambuchaAdm: Name is good but aggree it should return boolean and mapping to Yes/No should be on view.

bambuchaAdm:
> My proposal, to avoid all the summary helper methods that return `Seq` etc cluttering up the model is:
> 
> Lets try and represent the Summary Components using some models
> 
> ```
> case class SummarySection(heading: String, subSections: Seq[SummarySubSection])
> case class SummarySubSection(heading: Option[String], items: Seq[SummarySectionItem])
> case class SummarySectionItem(label: String, value: String, changeLink: Call)
> ```
> 
> And we could build the content in a `SummaryBuilder` class e.g.
> 
> ```
> class SummaryBuilder @Inject()(messages: Messages) {
>     def build(declaration: ExportsDeclaration):  Seq[SummarySection] =Seq(
>         thisDeclarationSection
>     )
>     
>     private def thisDeclarationSection: SummarySection = SummarySection(
>         heading = "This Declaration", // use messages to get this though
>         subSections = Seq(
>             SummarySubSection(
>                 heading = None, // Additional parties involved would use this, but not all sub-sections would
>                 items = Seq(
>                     SummarySectionItem("The declaration type", "Standard", routes.SomeController.display(Mode.Review))
>                 )
>         )
>     )
> }
> ```
> 
> Benefits:
> 
> * The builder can be responsible for getting the data in the right format (so no `summaryValue` helper methods)
> * The mapping can be tested in isolation (and not in a view test)
> * The view can just become a matter of rendering some html based on a `Seq[SummarySection]` (much easier to view test)
> * If the builder gets too big we can break it down into `ThisDeclarationSummaryBuilder` etc to keep it manageable
> 
> Cons:
> 
> * A little bit of an upfront cost to refactoring it but it's an investment of time I think is worth it

bambuchaAdm:
Sorry - missclick

app/forms/Ducr.scala
SteveSugden: Why rename at all?  There are a number of "form" classes with single elements with the same name as the class.  If we "fix" this we should fix them all.

app/forms/declaration/DepartureTransport.scala
```diff
@@ -56,6 +65,18 @@ object TransportCodes {
   val InlandWaterway = "8"
   val Unknown = "9"
 
+  val extractModeOfTransportValue: PartialFunction[String, String] = {
+    case Maritime                    => "Sea transport"
```
SteveSugden: I don't think we should be converting code -> display value like this.  That's what the messages are for.

app/forms/declaration/DispatchLocation.scala
```diff
@@ -17,13 +17,17 @@
...
 import play.api.data.{Form, Forms, Mapping}
 import play.api.libs.json.{Json, OFormat}
 import utils.validators.forms.FieldValidator.isContainedIn
 
 case class DispatchLocation(
   dispatchLocation: String // 2 upper case alphabetic characters
-)
+) {
+
+  def summaryValue: String = if (dispatchLocation == AllowedDispatchLocations.OutsideEU) "Yes" else "No"
```
SteveSugden: Similar - its not translatable.

app/forms/declaration/GoodsLocation.scala
```diff
@@ -31,7 +31,17 @@ case class GoodsLocation(
   addressLine: Option[String],
   postCode: Option[String],
   city: Option[String]
-)
+) {
+
+  def extractLocationCode: String =
```
SteveSugden: Just 'locationCode' ?

app/forms/declaration/GoodsLocation.scala
```diff
@@ -31,7 +31,17 @@ case class GoodsLocation(
...
+    retrieveCountryFromName(country).countryCode + typeOfLocation + qualifierOfIdentification + identificationOfLocation.getOrElse("")
+
+  def extractAddressValues: Seq[String] = {
+    val address = Seq(addressLine, city, postCode, Some(country)).flatten
+
+    if (address.length > 3) address else Seq.empty
```
SteveSugden: Why "3"?   Does this mean that if we don't have all four address elements then we return (show) nothing?

app/forms/declaration/InlandModeOfTransportCode.scala
SteveSugden: Again - not the right way to get display text

app/forms/declaration/NatureOfTransaction.scala
SteveSugden: Ditto

app/models/ExportsDeclaration.scala
SteveSugden: Is that really the best way to create a display text?  What if design aren't happy with "Simplified" or "Occasional" - then we'd have to change our enum definitions?

app/models/declaration/TransportInformation.scala
SteveSugden: ExportDeclaration has this method already (if that helps)

app/services/Countries.scala
SteveSugden: Just interested.  Why do we need to look up by name?

app/forms/declaration/DepartureTransport.scala
bambuchaAdm: Done

app/views/declaration/summary/related_documents.scala.html
```diff
@@ -0,0 +1,43 @@
...
+@(documents: Seq[Document])(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
+
+@if(documents.isEmpty) {
+    @components.table_row("previous-documents", HtmlTableRow(
+        label = messages("declaration.summary.transaction.previousDocuments"),
+        value = messages("site.no")
+    ))
+}
+
+@if(documents.nonEmpty) {
```
bambuchaAdm: else ?

PatrykRudnicki: I stayed with speparated `if` to have more readable code here. I mean to see the condition when this part will appear, but probably it'll be better to use `else`

app/forms/Ducr.scala
PatrykRudnicki: I'm removing this TODO

app/forms/declaration/BorderTransport.scala
PatrykRudnicki: This is moved to the view

app/forms/declaration/BorderTransport.scala
PatrykRudnicki: This is moved to the view

ed0906:
> My proposal, to avoid all the summary helper methods that return `Seq` etc cluttering up the model is:
> 
> Lets try and represent the Summary Components using some models
> 
> ```
> case class SummarySection(heading: String, subSections: Seq[SummarySubSection])
> case class SummarySubSection(heading: Option[String], items: Seq[SummarySectionItem])
> case class SummarySectionItem(label: String, value: String, changeLink: Call)
> ```
> 
> And we could build the content in a `SummaryBuilder` class e.g.
> 
> ```
> class SummaryBuilder @Inject()(messages: Messages) {
>     def build(declaration: ExportsDeclaration):  Seq[SummarySection] =Seq(
>         thisDeclarationSection
>     )
>     
>     private def thisDeclarationSection: SummarySection = SummarySection(
>         heading = "This Declaration", // use messages to get this though
>         subSections = Seq(
>             SummarySubSection(
>                 heading = None, // Additional parties involved would use this, but not all sub-sections would
>                 items = Seq(
>                     SummarySectionItem("The declaration type", "Standard", routes.SomeController.display(Mode.Review))
>                 )
>         )
>     )
> }
> ```
> 
> Benefits:
> 
> * The builder can be responsible for getting the data in the right format (so no `summaryValue` helper methods)
> * The mapping can be tested in isolation (and not in a view test)
> * The view can just become a matter of rendering some html based on a `Seq[SummarySection]` (much easier to view test)
> * If the builder gets too big we can break it down into `ThisDeclarationSummaryBuilder` etc to keep it manageable
> 
> Cons:
> 
> * A little bit of an upfront cost to refactoring it but it's an investment of time I think is worth it

Are we going to implement something along the lines of this? I think it will solve 90% of the review comments above

PatrykRudnicki:
@ed0906 After Friday discussion with the team, we decided to go with the messages. Additional class to inject or reuse for everything is not necessary. Idea was good, but we can solve everything with just messages (we still need to use it for translations).

app/forms/declaration/DeclarationAdditionalActors.scala
PatrykRudnicki: Moved to the messages, which are based on the code

ed0906:
> @ed0906 After Friday discussion with the team, we decided to go with the messages. Additional class to inject or reuse for everything is not necessary. Idea was good, but we can solve everything with just messages (we still need to use it for translations).

Does this mean we can get rid of all the helper methods in the model then?

PatrykRudnicki:
@ed0906 Yes, I'm doing it now, I didn't realise that it's not done on Friday

app/forms/common/Address.scala
PatrykRudnicki: Moved to the view and handled there

app/forms/declaration/GoodsLocation.scala
PatrykRudnicki: Moved to the view

app/forms/declaration/GoodsLocation.scala
PatrykRudnicki: Yes, frontend logic, moved to the view, but still there is the same if. I can wrap it in some value to describe this code. What do you think? Something like

`val isAddressContainsAllFields = if(address.length > 3)`

`if(isAddressContainsAllFields) address else Seq.empty`

Is it better? Do you have another idea

app/forms/declaration/InlandModeOfTransportCode.scala
PatrykRudnicki: Similar, moved to use messages

app/forms/declaration/NatureOfTransaction.scala
PatrykRudnicki: Moved to be handled by messages

app/forms/declaration/RepresentativeDetails.scala
PatrykRudnicki: Moved to be handled by messages

app/forms/declaration/TransportPayment.scala
PatrykRudnicki: Moved to be handled by messages

app/forms/declaration/additionaldeclarationtype/AdditionalDeclarationType.scala
PatrykRudnicki: Moved to be handled by messages

app/models/ExportsDeclaration.scala
PatrykRudnicki: Moved to the view

app/models/declaration/ProcedureCodesData.scala
PatrykRudnicki: Moved to the view

app/models/declaration/TransportInformation.scala
PatrykRudnicki: Removed. This logic is handled in the view

app/services/Countries.scala
PatrykRudnicki: Changed

app/services/Countries.scala
PatrykRudnicki: because our models are not align, somewhere we're storing code, somewhere whole name. Of course in places where we storing whole name we need to have code...

Changed

app/services/Countries.scala
PatrykRudnicki: Removed

app/services/DocumentType.scala
PatrykRudnicki: I'll stay with this, everyone has different opinion about it. Something that will be better to discuss

app/services/DocumentType.scala
PatrykRudnicki: Changed

app/services/DocumentType.scala
PatrykRudnicki: Changed

app/services/PackageTypes.scala
PatrykRudnicki: Changed

app/forms/declaration/DepartureTransport.scala
PatrykRudnicki: Removed

app/forms/declaration/DepartureTransport.scala
PatrykRudnicki: Removed

app/forms/declaration/DepartureTransport.scala
PatrykRudnicki: Removed

app/forms/declaration/DepartureTransport.scala
PatrykRudnicki: Removed

app/forms/declaration/DepartureTransport.scala
PatrykRudnicki: Removed

app/forms/declaration/DispatchLocation.scala
PatrykRudnicki: Removed

PatrykRudnicki:
@ed0906 @SteveSugden Please take a look once again. All comments fixed

app/forms/declaration/GoodsLocation.scala
SteveSugden: Not that concerned how the logic is implemented, but if that logic is for the view, I'd move it there.

PatrykRudnicki: So this is moved to the view

app/models/viewmodels/HtmlTableRow.scala
```diff
@@ -22,9 +22,14 @@ class HtmlTableRow(val label: String, val values: Seq[Option[String]])
...
 
-  def apply(label: String, value: Iterable[_]): HtmlTableRow = new HtmlTableRow(label, adjust(value))
+  def apply[T](label: String, value: Iterable[T]): HtmlTableRow = new HtmlTableRow(label, adjust(value))
 
-  private def adjust(iterable: Iterable[_]): Seq[Option[String]] = iterable match {
+  def apply[T](label: String, value: Seq[String]): HtmlTableRow = new HtmlTableRow(label, value.map(Option.apply))
+
+  def apply(label: String, value: String): HtmlTableRow = new HtmlTableRow(label, Seq(Some(value)))
+
+  // TODO review adjust method, working wrong
```
SteveSugden: Out of interest, how is it "wrong"?   Maybe add some detail to the TODO for next person to fix?

app/views/components/button_link.scala.html
```diff
@@ -17,5 +17,5 @@
 @(messageKey: String = "", href: String)(implicit messages: Messages)
 
 <div class="section">
-    <a href="@href" role="button" class="button">@messages(messageKey)</a>
+    <a href="@href" role="button" class="button" id="link-back">@messages(messageKey)</a>
```
SteveSugden: Why have you added an id here?  Is this not a general "button as a link" component?  The back_link component uses the id "back-link".

app/models/viewmodels/HtmlTableRow.scala
ed0906: We should explain what is wrong at least in the comment :)

app/forms/declaration/GoodsLocation.scala
ed0906: Looks like it's still there?

app/models/declaration/ProcedureCodesData.scala
ed0906: Looks like it needs deleting in the model

app/views/components/table_row.scala.html
```diff
@@ -16,16 +16,23 @@
...
-    <td class="previous-question-body">
-    @for(value <- tableRow.values){
-        @value.getOrElse("None")<br/>
-    }
-    </td>
-    <td class="link-right">
-        <a href="">Change</a>
+    <td class="bold"><span id="@id-label">@tableRow.label</span></td>
+    <td>
+        @if(tableRow.values.drop(1).isEmpty) {
```
SteveSugden: Is this "if tableRow has one value"?  If so, why the extra logic?  Is the only difference the lack of an index number in the id?  Why not be consistent?

app/models/viewmodels/HtmlTableRow.scala
PatrykRudnicki: Yes, thank you, it's fixed right now. The method handles Option[Option[String]] but struggle with Option[String] or something like this. it's unnecessary. I removed adjust method

app/views/components/button_link.scala.html
PatrykRudnicki: This is some kind of `button-link` it's different than other back links we have on pages. I added ID to have a possibility to fetch this element on the page
what is your proposal about it?

ed0906: I agree with Steve, this is a generic `button` component

Pass in an ID as an argument if you need it

app/views/components/table_row.scala.html
PatrykRudnicki: yes, it's mean if tableRow has one element. `drop` is safer here than tail, because will return empty seq is the sequence will be empty. Tail will throw an exception. If I have one element or don't have any value here I don't want to have index in id

app/views/components/button_link.scala.html
SteveSugden: My point was that the id seems to indicate this component has a specific use (to render a back link).   But I think the component is more general than that.  Maybe pass the id in as a parameter (defaulted to "button-link"? for views that don't care)?

app/views/components/table_row.scala.html
ed0906: why `drop(1).isEmpty` and not `size == whatever`

app/views/declaration/summary/items_section.scala.html
```diff
@@ -14,33 +14,19 @@
  * limitations under the License.
  *@
 
-@import models.declaration.Items
+@import models.requests.JourneyRequest
 @import models.viewmodels.HtmlTableRow
-@import models.DeclarationType.DeclarationType
 
-@(items: Option[Items], declarationType: DeclarationType)(implicit messages: Messages)
+@(declarationData: ExportsDeclaration)(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
```
ed0906: You can get `declarationData`  from the request cant you?

app/views/declaration/summary/locations_section.scala.html
```diff
@@ -14,67 +14,44 @@
...
  *@
 
-@import models.declaration.Locations
+@import forms.declaration.GoodsLocation
+@import models.requests.JourneyRequest
 @import models.viewmodels.HtmlTableRow
+@import services.Countries._
 
-@(locations: Option[Locations])(implicit messages: Messages)
+@(declarationData: ExportsDeclaration)(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
```
ed0906: Similar

app/views/declaration/summary/parties_section.scala.html
```diff
@@ -14,101 +14,57 @@
  * limitations under the License.
  *@
 
-@import forms.common.Address
-@import models.declaration.Parties
+@import models.requests.JourneyRequest
 @import models.viewmodels.HtmlTableRow
 
-@(parties: Option[Parties])(implicit messages: Messages)
+@(declarationData: ExportsDeclaration)(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
```
ed0906: Similar

app/views/declaration/summary/references_section.scala.html
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import forms.declaration.additionaldeclarationtype.AdditionalDeclarationType
+@import models.requests.JourneyRequest
+@import models.viewmodels.HtmlTableRow
+@import views.ViewDates
+
+@(declarationData: ExportsDeclaration)(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
```
ed0906: Similar

app/views/declaration/summary/summary_page.scala.html
```diff
@@ -17,21 +17,20 @@
...
+@import models.{ExportsDeclaration, Mode}
 @import models.requests.JourneyRequest
 @import models.viewmodels.HtmlTableRow
 @import views.html.declaration.summary._
 @import views.{Title, ViewDates}
 
 @this(main_template: views.html.main_template)
 
-@(mode: Mode, suppDecData: SupplementaryDeclarationData, legalDeclarationForm: Form[LegalDeclaration])(implicit request: JourneyRequest[_], messages: Messages, appConfig: AppConfig)
+@(mode: Mode, declarationData: ExportsDeclaration, legalDeclarationForm: Form[LegalDeclaration])(implicit request: JourneyRequest[AnyContent], messages: Messages, appConfig: AppConfig)
```
ed0906: similar

app/views/declaration/summary/transaction_section.scala.html
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import models.requests.JourneyRequest
+@import models.viewmodels.HtmlTableRow
+
+@(declarationData: ExportsDeclaration)(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
```
ed0906: similar

app/views/declaration/summary/transport_section.scala.html
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ *@
+
+@import models.requests.JourneyRequest
+@import models.viewmodels.HtmlTableRow
+
+@import forms.declaration.DepartureTransport
+@import forms.declaration.BorderTransport
+
+@(declarationData: ExportsDeclaration)(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
```
ed0906: similar

app/views/declaration/summary/warehouse_section.scala.html
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import models.requests.JourneyRequest
+@import models.viewmodels.HtmlTableRow
+
+@(declarationData: ExportsDeclaration)(implicit messages: Messages, journeyRequest: JourneyRequest[AnyContent])
```
ed0906: similar

app/views/components/table_row.scala.html
PatrykRudnicki: Size is not recommended method to check isEmpty or nonEmpty without checking specific size, because it must iterate through whole sequence to find how many elements you have. Moreover in this case I'll need to check isEmpty and size == 1, so two cases

app/views/declaration/summary/items_section.scala.html
PatrykRudnicki: Good point, it was previous implementation. I'll remove it right now

app/views/components/table_row.scala.html
SteveSugden: All this interesting discussion doesn't explain what's wrong with "id=item-0" if there happens to be 1 item in the list?  That just means that whoever is writing the test(s) has to know/care how many items were there.

app/models/viewmodels/HtmlTableRow.scala
```diff
@@ -16,22 +16,11 @@
...
-  private def adjust(iterable: Iterable[_]): Seq[Option[String]] = iterable match {
-    case Nil => Seq.empty
-    case _ =>
-      iterable.map {
-        case str: String       => Some(str)
-        case Some(str: String) => Some(str)
-        case _                 => None
-      }.filter(_.isDefined).toSeq
-  }
+  def apply[T](label: String, value: Seq[String]): HtmlTableRow = new HtmlTableRow(label, value.map(Option.apply))
```
bambuchaAdm: You could remove type parameter

PatrykRudnicki: Removed

app/views/declaration/summary/items_section.scala.html
PatrykRudnicki: For every section we will stay with declarationData as argument. We will use implicit one on the `summaryPage`

app/views/declaration/summary/locations_section.scala.html
PatrykRudnicki: For every section we will stay with declarationData as argument. We will use implicit one on the `summaryPage`

app/views/declaration/summary/parties_section.scala.html
PatrykRudnicki: For every section we will stay with declarationData as argument. We will use implicit one on the `summaryPage`

app/views/declaration/summary/references_section.scala.html
PatrykRudnicki: For every section we will stay with declarationData as argument. We will use implicit one on the `summaryPage`

app/views/declaration/summary/summary_page.scala.html
PatrykRudnicki: Removed

app/views/declaration/summary/transport_section.scala.html
PatrykRudnicki: For every section we will stay with declarationData as argument. We will use implicit one on the `summaryPage`

app/views/declaration/summary/transaction_section.scala.html
PatrykRudnicki: For every section we will stay with declarationData as argument. We will use implicit one on the `summaryPage`

app/views/declaration/summary/warehouse_section.scala.html
PatrykRudnicki: For every section we will stay with declarationData as argument. We will use implicit one on the `summaryPage`

app/views/declaration/summary/additional_and_authorised_parties_section.scala.html
```diff
@@ -0,0 +1,62 @@
...
+    ))
+}
+@if(holders.isEmpty) {
+    @components.table_row("holders", HtmlTableRow(
+        label = messages("declaration.summary.parties.holders"),
+        value = messages("site.no")
+    ))
+}
+
+@if(additionalActors.nonEmpty) {
```
bambuchaAdm: Please move this next to inverted condition - would be easier to consider about this

app/views/components/table_row.scala.html
PatrykRudnicki: It's table row not item. So this can be applied for additional document, package information and every place where we use it. Previous implementation was similar probably.

Doesn't matter for me how we want to deal with it. I can remove this part. For me it's better to have `additional-document` instead of `additional-document-0` if I have just one element

app/views/declaration/summary/additional_and_authorised_parties_section.scala.html
PatrykRudnicki: The logic on this page is weird. We need to display section in specific order. You're not ably to do it with if else, because sometimes holders are at the top of the page, sometimes below additional actors

app/views/components/table_row.scala.html
SteveSugden: Up to you.  I think `additional-document-0` is better and it makes the logic simpler.   (Both for view and test)

app/views/components/button_link.scala.html
bambuchaAdm: Yea - this is generic component should have id parameter or does not have it at all.

PatrykRudnicki: Changed

app/views/components/table_row.scala.html
PatrykRudnicki: As we discussed, if we change it here we will have `-0` in every id for every field. I'll try to find some more elegant solution

PatrykRudnicki: We're using `HtmlTableRow` for it. It's difficult to achieve it without passing additional arguments or changing whole approach around this class

app/views/components/button_link.scala.html
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  *@
 
-@(messageKey: String = "", href: String)(implicit messages: Messages)
+@(messageKey: String = "", href: String, id: String = "back-link")(implicit messages: Messages)
```
ed0906: I still dont think `back-link` is appropriate

this is a `button` component, it should know the context it is used in is "going back"

PatrykRudnicki: Done

PR #647, opened 2019-11-29
CEDS-1945 fix back navigation and container seals for occasional dec

SteveSugden:


app/controllers/declaration/TransportContainerController.scala
```diff
@@ -187,12 +186,6 @@ class TransportContainerController @Inject()(
     updateExportsDeclarationSyncDirect(_.updateContainers(updatedContainers))
 
   private def redirectAfterAdd(mode: Mode, containerId: String)(implicit request: JourneyRequest[AnyContent]) =
-    if (allowSeals)
-      navigator.continueTo(routes.SealController.displaySealSummary(mode, containerId))
-    else
-      navigator.continueTo(routes.TransportContainerController.displayContainerSummary(mode))
-
-  private def allowSeals(implicit request: JourneyRequest[AnyContent]) =
```
SteveSugden: All dec types allow seals to be added.  So I decided to get rid of this method to avoid future bugs/confusion.

app/controllers/navigation/Navigator.scala
```diff
@@ -122,7 +122,7 @@ object Navigator {
 
   val occasional: PartialFunction[DeclarationPage, Mode => Call] = {
     case BorderTransport             => controllers.declaration.routes.SupervisingCustomsOfficeController.displayPage
-    case TransportPayment            => controllers.declaration.routes.BorderTransportController.displayPage
+    case TransportPayment            => controllers.declaration.routes.SupervisingCustomsOfficeController.displayPage
```
SteveSugden: This should be the same as simplified decs

app/controllers/navigation/Navigator.scala
```diff
@@ -122,7 +122,7 @@ object Navigator {
 
   val occasional: PartialFunction[DeclarationPage, Mode => Call] = {
     case BorderTransport             => controllers.declaration.routes.SupervisingCustomsOfficeController.displayPage
```
PatrykRudnicki: Can you remove in that case BorderTransport? user shouldn't be there

SteveSugden: Good spot!.   Removing it for simp and occ simp

PR #648, opened 2019-11-29
CEDS-1439 - preserve item list order

SteveSugden:


app/controllers/declaration/AdditionalFiscalReferencesController.scala
```diff
@@ -111,6 +111,6 @@ class AdditionalFiscalReferencesController @Inject()(
     implicit request: JourneyRequest[AnyContent]
   ): Future[Option[ExportsDeclaration]] =
     updateExportsDeclarationSyncDirect { model =>
-      model.updatedItem(itemId, item => item.copy(additionalFiscalReferencesData = Some(updatedAdditionalFiscalReferencesData)))
+      model.updateItem(itemId, item => item.copy(additionalFiscalReferencesData = Some(updatedAdditionalFiscalReferencesData)))
```
PatrykRudnicki: What is the difference between those methods?

app/models/ExportsDeclaration.scala
```diff
@@ -50,10 +50,9 @@ case class ExportsDeclaration(
 
   def isComplete: Boolean = status == DeclarationStatus.COMPLETE
 
-  def updatedItem(itemId: String, update: ExportItem => ExportItem): ExportsDeclaration =
-    itemBy(itemId).fold(this) { item =>
-      val updated = update(item)
-      copy(items = items.filterNot(_.id.equalsIgnoreCase(itemId)) + updated)
+  def updateItem(itemId: String, update: ExportItem => ExportItem): ExportsDeclaration =
+    itemBy(itemId).fold(this) { itemToReplace =>
+      copy(items = items.map(item => if (item.equals(itemToReplace)) update(itemToReplace) else item))
```
bambuchaAdm: item eq itemToReplace - nice chat for reference equality

app/controllers/declaration/AdditionalFiscalReferencesController.scala
SteveSugden: The letter 'd'

app/models/ExportsDeclaration.scala
SteveSugden: You're right.  Reference equality is good enough here.  I didn't even realise that scala had a third way of defining "equal" (i.e. 'eq') - I'll stick with '==' as an old Java hack.

app/controllers/declaration/AdditionalFiscalReferencesController.scala
PatrykRudnicki: I thought that there are two different method for update item, didn't notice that was rename

SteveSugden: It's been renamed back.  I thought it was a typo.  Apparently its named after scala convention for collection libraries.

PR #653, opened 2019-12-03
CEDS-2015 Change buttons for summary page

PatrykRudnicki:


PatrykRudnicki:
@SteveSugden Yes, for now we decided to just redirect to the specific page, in the future it will be improved, for now there is no decision.

Probably from technical perspective we will pass `Mode.Amend` or different mode that will define what mode is on and we will be handling this somehow :)

PR #654, opened 2019-12-03
CEDS-1126 Add scope attributes to the table headings and rows

PauloPortugal:
These are required for accessibility purposes.


app/views/components/fields/removable_elements_table.scala.html
```diff
@@ -33,17 +33,17 @@
...
+                    <th id="@{id}-heading" colspan="2" scope="col">@heading</th>
                   </tr>
               </thead>
           }
           <tbody>
             @for((elem, index) <- elements.zipWithIndex) {
                 <tr id="@{id}__row@index">
-                    <th id="@{id}-row@index-label">@{elem.label}</th>
-                    <th id="@{id}-row@index-remove_button">
+                    <td scope="row" id="@{id}-row@index-label">@{elem.label}</td>
```
bambuchaAdm: td ?

app/views/declaration/additional_information.scala.html
```diff
@@ -49,19 +49,19 @@
                 <table id="additional_information">
                     <thead>
                         <tr>
-                            <th class="code-header" colspan="1">@messages("supplementary.additionalInformation.table.headers.code")</th>
-                            <th class="description-header" colspan="1">@messages("supplementary.additionalInformation.table.headers.description")</th>
-                            <th colspan="1"></th>
+                            <th class="code-header" scope="col">@messages("supplementary.additionalInformation.table.headers.code")</th>
+                            <th class="description-header" scope="col">@messages("supplementary.additionalInformation.table.headers.description")</th>
+                            <td></td>
```
PatrykRudnicki: If it's in `thead` is should be an empty `th`?

app/views/declaration/additional_information.scala.html
```diff
@@ -49,19 +49,19 @@
...
                         </tr>
                     </thead>
                     <tbody>
                     @for((elem, index) <- items.zipWithIndex) {
                         <tr id="removable_elements__row@index">
-                            <th class="code-row" id="removable_elements-row@index-code">@{elem.code}</th>
-                            <th class="description-row" id="removable_elements-row@index-info">@{elem.description}</th>
-                            <th id="removable_elements-row@index-remove_button">
+                            <td class="code-row" id="removable_elements-row@index-code" scope="row">@{elem.code}</td>
```
PatrykRudnicki: If it was a `th` before should we add `bold` class to this?

app/views/components/fields/removable_elements_table.scala.html
PauloPortugal: **TH** is used for table header cells while **TD** is used for table data cells. This distinction gives user agents a means to render such cells distinctly, for instance by using a larger or heavier font for header cells.

app/views/declaration/additional_information.scala.html
PauloPortugal: It can be `TD`.
It was an empty `TH`, which caused a warning/alert. We have two options:
1. remove that and add `colspan="2"`
2. add `td` which can be empty

Two shifts the button to the right, which we prefer.

app/views/declaration/additional_information.scala.html
PauloPortugal: **TH** is used for table header cells while **TD** is used for table data cells. This distinction gives user agents a means to render such cells distinctly, for instance by using a larger or heavier font for header cells.

PR #661, opened 2019-12-05
CEDS-2013 - skip country of routing for clearance request

SteveSugden:


app/controllers/declaration/DestinationCountryController.scala
```diff
@@ -67,9 +67,9 @@ class DestinationCountryController @Inject()(
 
   private def redirectToNextPage(mode: Mode)(implicit request: JourneyRequest[AnyContent]): Result =
     request.declarationType match {
-      case DeclarationType.SUPPLEMENTARY =>
+      case DeclarationType.SUPPLEMENTARY | DeclarationType.CLEARANCE =>
         navigator.continueTo(controllers.declaration.routes.LocationController.displayPage(mode))
-      case DeclarationType.STANDARD | DeclarationType.SIMPLIFIED | DeclarationType.OCCASIONAL | DeclarationType.CLEARANCE =>
+      case DeclarationType.STANDARD | DeclarationType.SIMPLIFIED | DeclarationType.OCCASIONAL =>
```
ed0906: Do we need to skip the back end mapping for this too?

test/views/declaration/LocationViewSpec.scala
```diff
@@ -401,4 +401,39 @@ class LocationViewSpec extends UnitViewSpec with ExportsTestData with Stubs with
...
+        "display 'Back' button that links to correct page" in {
+          val backButton = view.getElementById("back-link")
+
+          backButton must containText("site.back")
+          backButton.getElementById("back-link") must haveHref(redirect)
+        }
+      }
+
+    "for Standard declaration" should {
+      behave like viewWithCorrectBackButton(DeclarationType.STANDARD, controllers.declaration.routes.RoutingCountriesSummaryController.displayPage())
```
ed0906: `behave like` thats new

What's that do?

app/controllers/declaration/DestinationCountryController.scala
SteveSugden: What do you mean @ed0906 ?

ed0906: The back end is set up to map clearance requests like  standards is it not?
If we are skipping pages, does that mean we can remove that part of the mapping from the back end for this journey type?

test/views/declaration/LocationViewSpec.scala
SteveSugden: It's not that new - it's used in a couple of places.   Basically used to create parameterised blocks of code that can be re-used

app/controllers/declaration/DestinationCountryController.scala
PatrykRudnicki: I think that those fields are optional in the model, so it's not a problem that backend model has everything

app/controllers/declaration/RoutingCountriesSummaryController.scala
```diff
@@ -46,18 +46,13 @@ class RoutingCountriesSummaryController @Inject()(
...
-      case _ =>
-        val countryCodes = request.cacheModel.locations.routingCountries
-        val countries = findByCodes(countryCodes)
-
-        if (countries.nonEmpty) {
-          Ok(routingCountriesSummaryPage(mode, RoutingQuestionYesNo.form(), countries))
-        } else {
-          navigator.continueTo(controllers.declaration.routes.RoutingCountriesController.displayRoutingQuestion(mode))
-        }
+    val countryCodes = request.cacheModel.locations.routingCountries
```
PatrykRudnicki: What about case that you removed? This screen is not a part of the `SUPPLEMENTARY` journey

PR #663, opened 2019-12-05
CEDS-2034 Skip UN Dangerous Goods screen

PatrykRudnicki:


app/controllers/declaration/CommodityDetailsController.scala
```diff
@@ -56,8 +56,11 @@ class CommodityDetailsController @Inject()(
...
         validForm =>
           updateExportsCache(itemId, validForm).map { _ =>
-            navigator
-              .continueTo(controllers.declaration.routes.UNDangerousGoodsCodeController.displayPage(mode, itemId))
+            if (request.isType(DeclarationType.CLEARANCE)) {
+              navigator.continueTo(controllers.declaration.routes.CUSCodeController.displayPage(mode, itemId))
+            } else {
+              navigator.continueTo(controllers.declaration.routes.UNDangerousGoodsCodeController.displayPage(mode, itemId))
+            }
         }
```
SteveSugden: Not critical but we normally extract this logic into a sep. "nextPage" method.

test/views/declaration/spec/UnitViewSpec.scala
```diff
@@ -49,13 +50,15 @@ class UnitViewSpec extends UnitSpec with ViewMatchers {
   def messagesKey(key: String): BeMatcher[String] = new MessagesKeyMatcher(key)
 
   def onEveryDeclarationJourney(f: JourneyRequest[_] => Unit): Unit =
```
SteveSugden: Nice.  I forgot about this way of testing different journeys

test/unit/controllers/declaration/CommodityDetailsControllerSpec.scala
```diff
@@ -69,6 +72,8 @@ class CommodityDetailsControllerSpec extends ControllerSpec with OptionValues {
 
       "display page method is invoked and cache is empty" in {
 
+        withNewCaching(aDeclaration())
```
bambuchaAdm: If we would merge earlier #662 there are helper methods for testing controllers against multiple declaration type.

app/controllers/declaration/CommodityDetailsController.scala
PatrykRudnicki: Good point, done

test/unit/controllers/declaration/CommodityDetailsControllerSpec.scala
PatrykRudnicki: We can merge it before, but I'll prefer to stay with this solution for now to finish ticket, we need to refactor a lot of controller specs, so it can be done in separate PR

PR #664, opened 2019-12-06
CEDS-2034 Improved test and cleanup

PatrykRudnicki:


PatrykRudnicki:
There is just `CommonMessages` trait that make slightly sense for information like `site.change` etc, but it'll be good to review it and remove non general messages

PR #666, opened 2019-12-06
Cleanup for rest not common messages

PatrykRudnicki:


test/helpers/views/components/DateMessages.scala
```diff
@@ -18,11 +18,9 @@ package helpers.views.components
...
-  val date: String = "dateTime.date"
-
-  val dateFormatError: String = date + ".error.format"
-  val dateOutOfRangeError: String = date + ".error.outOfRange"
-  val dayEmptyFieldError: String = date + ".day.error.empty"
-  val monthEmptyFieldError: String = date + ".month.error.empty"
-  val yearEmptyFieldError: String = date + ".year.error.empty"
+  val dateFormatError: String = "dateTime.date.error.format"
+  val dateOutOfRangeError: String = "dateTime.date.error.outOfRange"
+  val dayEmptyFieldError: String = "dateTime.date.day.error.empty"
```
bambuchaAdm: Is this are still in use ?

PatrykRudnicki: Yes, as I mentioned in description on slack, I should put it here. Mostly it's used for DataSpec, but there is a few places in tests where we're using it. I don't know that there is some value in this. We can inline those too, because there is small amount of places to change

PatrykRudnicki: Removed and cleaned

PR #674, opened 2019-12-10
CEDS-2037 skip border transport and payment

SteveSugden:


test/unit/controllers/declaration/BorderTransportControllerSpec.scala
```diff
@@ -101,69 +100,70 @@ class BorderTransportControllerSpec extends ControllerSpec {
...
-      }
+    "we are on occasional declaration journey" should {
+      def declarationFactory() = aDeclaration(withType(DeclarationType.OCCASIONAL))
+      behave like borderTransportControllerForInvalidJourney(declarationFactory)
     }
 
+    "we are on clearance declaration journey" should {
+      def declarationFactory() = aDeclaration(withType(DeclarationType.CLEARANCE))
+      behave like borderTransportControllerForInvalidJourney(declarationFactory)
+    }
```
PatrykRudnicki: Should we use our helper `onJourney(...)`?

SteveSugden: I just went with fixing / extending whatever style was there already.  I could have a go at refactoring I guess as we are not too rushed.

app/controllers/declaration/BorderTransportController.scala
```diff
@@ -42,25 +42,28 @@ class BorderTransportController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
-  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+  private val validTypes = Seq(DeclarationType.STANDARD, DeclarationType.SUPPLEMENTARY)
+
+  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType(validTypes: _*)) { implicit request =>
```
ed0906: Perhaps its worth adding an extra method to JourneyAction  that accepts an `Iterable` so you dont have to do the `:_*` everywhere

SteveSugden: Yes, I was going to ask about doing something like that.   Consider it done.

SteveSugden:
Tests updated if you want to take another look @bambuchaAdm @PatrykRudnicki

PR #675, opened 2019-12-10
CEDS-2047 Simplified declaration Summary page

PatrykRudnicki:


test/views/declaration/spec/UnitViewSpec.scala
```diff
@@ -41,17 +42,17 @@ class UnitViewSpec extends UnitSpec with ViewMatchers {
...
-  def checkErrorsSummary(view: Document) = {
+  def checkErrorsSummary(view: Document): Assertion = {
     view.getElementById("error-summary-heading").text() must be("error.summary.title")
     view.getElementsByClass("error-summary error-summary--show").get(0).getElementsByTag("p").text() must be("error.summary.text")
   }
 
   def messagesKey(key: String): BeMatcher[String] = new MessagesKeyMatcher(key)
 
-  def onEveryDeclarationJourney(f: JourneyRequest[_] => Unit): Unit =
+  def onEveryDeclarationJourney(f: JourneyRequest[AnyContent] => Unit): Unit =
```
bambuchaAdm: why go from wildcard to concrete implementation? View should not depend on value how controller would parse it

PatrykRudnicki: View required AnyContent. I'll change the views to have underscore

PatrykRudnicki: Done, please take another look

PR #678, opened 2019-12-10
BAU - fix TamperMonkey bug with remembering dec choice

SteveSugden:


docs/Customs Declare Exports AutoComplete.js
```diff
@@ -289,6 +290,7 @@ function completePage() {
         document.getElementsByClassName('button')[0].click()
     }
     if (currentPageIs('/customs-declare-exports/declaration/items/.*/add-document')) {
+        console.log("using choice " + getChoice());
```
PatrykRudnicki: Do we want to keep `console.log`?

SteveSugden: Thought I deleted that

PR #681, opened 2019-12-12
CEDS-2036 Fix test compilation

bambuchaAdm:


bambuchaAdm:
It didn't - dunno why but it didn't also triggered and looks all green without it.

PR #683, opened 2019-12-13
CEDS-2036 Correct format for declaration exchange

bambuchaAdm:


app/connectors/exchange/ExportsDeclarationExchange.scala
```diff
@@ -234,8 +234,8 @@ object ExportsDeclarationExchange {
     )
     JsObject(values.flatten)
   }
-
-  implicit val format: OFormat[ExportsDeclarationExchange] = Json.format[ExportsDeclarationExchange]
```
PatrykRudnicki: What about some tests for reads and writes? We should cover it in cases when we're making custom reads and writes

PR #684, opened 2019-12-13
CEDS-2015 Add accessibility messages to change buttons

PatrykRudnicki:


app/views/components/table_row.scala.html
```diff
@@ -19,9 +19,16 @@
 
 @(id: String, tableRow: HtmlTableRow)(implicit messages: Messages)
 
-@changeLink(callOpt: Option[Call]) = {
+@changeLink(callOpt: Option[Call], changeLabel: Option[String]) = {
```
SteveSugden: Can the changeLabel be non-optional?    Currently if it's None then the "Change" text will be "aria-hidden" with nothing to replace it.

conf/messages.en
```diff
@@ -891,38 +922,59 @@ declaration.summary.transaction.natureOfTransaction.7 = Military
...
 declaration.summary.transaction.previousDocuments = Documents related to this declaration
+declaration.summary.transaction.previousDocuments.change = Change documents related to this declaration
 declaration.summary.transaction.previousDocuments.type = Document type
 declaration.summary.transaction.previousDocuments.reference = Document reference
+declaration.summary.transaction.previousDocuments.document.change = Change document {0}
 declaration.summary.items = About the items
 declaration.summary.items.empty = You have not added any items yet
 declaration.summary.items.item.sequenceId = Declaration item {0}
 declaration.summary.items.item.procedureCode = Procedure code
+declaration.summary.items.item.procedureCode.change = Change procedure code from item {0}
```
SteveSugden: I would personally have "Change x for item y" (rather than 'from') but I guess that wasn't your call.

app/views/components/table_row.scala.html
PatrykRudnicki: Yes

conf/messages.en
PatrykRudnicki: There wasn't anything about it in the ticket. I asked designers about it during implementation and they approved it. I think it requires more discussion. Let's leave it for now

SteveSugden: If it wasn't on the ticket then I'd either ask design or do a quick replace text.

PatrykRudnicki: Yup, you're right, confirmed and changing

PR #687, opened 2019-12-17
Ceds 2036 Split departure transport + skip departure transport on clearance request

bambuchaAdm:


app/connectors/exchange/ExportsDeclarationExchange.scala
```diff
@@ -94,7 +94,7 @@ object ExportsDeclarationExchange {
     dispatchLocation: Option[DispatchLocation] = None,
     additionalDeclarationType: Option[AdditionalDeclarationType] = None,
     consignmentReferences: Option[ConsignmentReferences] = None,
-    departureTransport: Option[DepartureTransport] = None,
+    departureTransportOld: Option[(Option[String], Option[DepartureTransport])] = None,
```
ed0906: Why?

We haven't had to do this for any other of these "split" stories?

app/connectors/exchange/ExportsDeclarationExchange.scala
```diff
@@ -183,7 +186,9 @@ object ExportsDeclarationExchange {
       (__ \ "dispatchLocation").readNullable[DispatchLocation] and
       (__ \ "additionalDeclarationType").readNullable[AdditionalDeclarationType] and
       (__ \ "consignmentReferences").readNullable[ConsignmentReferences] and
-      (__ \ "departureTransport").readNullable[DepartureTransport] and
+      (__ \ "departureTransport").readNullable({
```
ed0906: Why do we still have 2 reads?

I thought we agreed this wasn't necessary?

ed0906:
I think this task has taken a lot longer than it should have because of scope creep :(
In my opinion, the problem of versioning should been a technical ticket (and solved as one)

app/connectors/exchange/ExportsDeclarationExchange.scala
bambuchaAdm: No backend speak with version1 but understood version1 and version2. Now frontend could undestood version1 and version2 and writes to backend with version2.

bambuchaAdm:
@PatrykRudnicki After this PR we are able to remove second reads from backend and bump writes to version2 - after that PR we would be able to remove second reads end with only one.

@ed0906 It was little widen though work made by @PatrykRudnicki on summary page - required more place to update `ModeOfTransportCode`

PR #688, opened 2019-12-18
CEDS-2048 Summary page for Occasional Declaration

PatrykRudnicki:
Additionally Improvements around Fiscal Information Controller in separate commit

### DIFF
--- a/app/controllers/declaration/FiscalInformationController.scala
+++ b/app/controllers/declaration/FiscalInformationController.scala
@@ -19,7 +19,7 @@ package controllers.declaration
 import controllers.actions.{AuthAction, JourneyAction}
 import controllers.navigation.Navigator
 import forms.declaration.FiscalInformation
-import forms.declaration.FiscalInformation.form
+import forms.declaration.FiscalInformation._
 import javax.inject.Inject
 import models.{ExportsDeclaration, Mode}
 import models.requests.JourneyRequest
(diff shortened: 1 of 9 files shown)


app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -60,29 +60,32 @@ class FiscalInformationController @Inject()(
...
-      model.updatedItem(itemId, item => item.copy(fiscalInformation = Some(updatedFiscalInformation)))
-    }).map(_ => ())
+  private def updateCache(itemId: String, updatedFiscalInformation: FiscalInformation)(
+    implicit request: JourneyRequest[AnyContent]
+  ): Future[Option[ExportsDeclaration]] = {
+    val updateModel: ExportsDeclaration => ExportsDeclaration =
+      (model: ExportsDeclaration) =>
+        updatedFiscalInformation.onwardSupplyRelief match {
+          case FiscalInformation.AllowedFiscalInformationAnswers.yes =>
+            model.updatedItem(itemId, item => item.copy(fiscalInformation = Some(updatedFiscalInformation)))
```
MaciejRewera: Nit: Maybe the `item` could be extracted into a variable just before passing into the update?

PatrykRudnicki: Item is different for Yes and for No, so I'm not sure that building this outside make sense

bambuchaAdm: What do you think about move item copy part as dedicated method on `ExportItem` ?

app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -60,29 +60,32 @@ class FiscalInformationController @Inject()(
...
 
-  //TODO Use one method instead of updateCacheForYes and updateCacheForNo
-  private def updateCacheForYes(itemId: String, updatedFiscalInformation: FiscalInformation)(implicit req: JourneyRequest[AnyContent]): Future[Unit] =
-    updateExportsDeclarationSyncDirect(model => {
-      model.updatedItem(itemId, item => item.copy(fiscalInformation = Some(updatedFiscalInformation)))
-    }).map(_ => ())
+  private def updateCache(itemId: String, updatedFiscalInformation: FiscalInformation)(
+    implicit request: JourneyRequest[AnyContent]
+  ): Future[Option[ExportsDeclaration]] = {
+    val updateModel: ExportsDeclaration => ExportsDeclaration =
```
bambuchaAdm: why function val instead of local def ?

app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -60,29 +60,32 @@ class FiscalInformationController @Inject()(
...
-      model.updatedItem(itemId, item => item.copy(fiscalInformation = Some(updatedFiscalInformation), additionalFiscalReferencesData = None))
-    }).map(_ => ())
+    updateExportsDeclarationSyncDirect(updateModel)
+  }
 
+  private def redirectToNextPage(mode: Mode, itemId: String, fiscalInformation: FiscalInformation)(
+    implicit request: JourneyRequest[AnyContent]
+  ): Result =
+    fiscalInformation.onwardSupplyRelief match {
+      case FiscalInformation.AllowedFiscalInformationAnswers.yes =>
```
bambuchaAdm: What do you  think about import all things from `FiscalInformation` ?

app/controllers/declaration/FiscalInformationController.scala
PatrykRudnicki: Logic here is slightly connected with the user choice, so we're updating differently for Yes and for No. Not sure that there is value to moving this to the item, but I can do it

app/controllers/declaration/FiscalInformationController.scala
PatrykRudnicki: Good point.

app/controllers/declaration/FiscalInformationController.scala
PatrykRudnicki: Don't know. I'll change it, local def is probably easier to read

PatrykRudnicki: Done

app/controllers/declaration/FiscalInformationController.scala
PatrykRudnicki: Done

PR #689, opened 2019-12-18
CEDS-2036 Update quick submit script after page split

bambuchaAdm:


docs/Customs Declare Exports AutoComplete.js
```diff
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         Customs Declare Exports AutoComplete
 // @namespace    http://tampermonkey.net/
-// @version      1.18
+// @version      1.20
```
PatrykRudnicki: Why jump to 1.20? Not 1.19

bambuchaAdm: Really dunno - it should just have to be higher the previews

PR #704, opened 2020-01-09
CEDS-2097 Align tables and change links

PauloPortugal:
* All lists and tables of questions/answers should be the same width,
  and the Change link should be right-aligned
* All Questions should be vertically-aligned top
* Headings and sub-heading should not be wrapped.
* The vertical spacing between elements of the submission form at the
  bottom of the screen should be consistent with the prototype.

This is to align the page to match the prototype.
This page is not accessible, this will be addressed once we move to the
GDS design system

with @apodi @alsahid1993

### DIFF
--- a/app/views/declaration/summary/package_information.scala.html
+++ b/app/views/declaration/summary/package_information.scala.html
@@ -22,24 +22,26 @@
 @(itemId: String, itemNo: Int, packageInformation: Seq[PackageInformation])(implicit messages: Messages, journeyRequest: JourneyRequest[_])
 
 @if(packageInformation.nonEmpty){
+  <table>
     <caption class="heading-small"><span id="package-information-@itemNo">@messages("declaration.summary.items.item.packageInformation")</span></caption>
-    <tr class="section">
-        <td class="bold"><span id="package-information-type-@itemNo">@messages("declaration.summary.items.item.packageInformation.type")</span></td>
-        <td class="bold"><span id="package-information-number-@itemNo">@messages("declaration.summary.items.item.packageInformation.number")</span></td>
-        <td class="bold"><span id="package-information-markings-@itemNo">@messages("declaration.summary.items.item.packageInformation.markings")</span></td>
-        <td></td>
-    </tr>
-    @packageInformation.zipWithIndex.map { case (information, index) =>
         <tr class="section">
-            <td><span id="package-information-@itemNo-code-@index">@information.typesOfPackagesText</span></td>
-            <td><span id="package-information-@itemNo-number-@index">@information.numberOfPackages</span></td>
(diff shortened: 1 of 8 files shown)


app/views/declaration/summary/package_information.scala.html
```diff
@@ -22,24 +22,26 @@
 @(itemId: String, itemNo: Int, packageInformation: Seq[PackageInformation])(implicit messages: Messages, journeyRequest: JourneyRequest[_])
 
 @if(packageInformation.nonEmpty){
+  <table>
```
PatrykRudnicki: I know that this `table` will solve the problem with `caption`, but now we will have a nested tables. Is it okay? I mean this component is used inside another table

app/views/declaration/summary/supporting_documents.scala.html
```diff
@@ -22,22 +22,24 @@
...
 
 @if(supportingDocuments.nonEmpty){
-    <caption class="heading-small"><span id="supporting-documents-@itemNo">@messages("declaration.summary.items.item.supportingDocuments")</span></caption>
-    <tr class="section">
-        <td class="bold"><span id="supporting-documents-code-@itemNo">@messages("declaration.summary.items.item.supportingDocuments.code")</span></td>
-        <td class="bold"><span id="supporting-documents-information-@itemNo">@messages("declaration.summary.items.item.supportingDocuments.information")</span></td>
-        <td></td>
-    </tr>
-    @supportingDocuments.zipWithIndex.map { case (document, index) =>
+  <table>
```
PatrykRudnicki: The same question here

app/views/declaration/summary/union_and_national_codes.scala.html
```diff
@@ -22,22 +22,24 @@
...
 
 @if(additionalInformation.nonEmpty){
-    <caption class="heading-small"><span id="additional-information-@itemNo">@messages("declaration.summary.items.item.additionalInformation")</span></caption>
-    <tr class="section">
-        <td class="bold"><span id="additional-information-code-@itemNo">@messages("declaration.summary.items.item.additionalInformation.code")</span></td>
-        <td class="bold"><span id="additional-information-information-@itemNo">@messages("declaration.summary.items.item.additionalInformation.information")</span></td>
-        <td></td>
-    </tr>
-    @additionalInformation.zipWithIndex.map { case (information, index) =>
+  <table>
```
PatrykRudnicki: And here

app/views/declaration/summary/package_information.scala.html
PauloPortugal: It actually solves the problem with the Change link alignment. 
The summary page was not accessible, with this change we introduce three new alerts

PR #711, opened 2020-01-13
CEDS-1098: Non GB EORI number allowed

suryend4u:
* Removed validation for only GB. 
* First 2 entries must be letters.
* Max limit of 17 & minimum limit of 3.

### DIFF
--- a/app/utils/validators/forms/FieldValidator.scala
+++ b/app/utils/validators/forms/FieldValidator.scala
@@ -137,7 +137,7 @@ object FieldValidator {
   private val emailPattern = Pattern.compile("""^\S+@\S+$""")
   val isValidEmail: String => Boolean = (name: String) => emailPattern.matcher(name).matches()
 
-  private val eoriPattern = Pattern.compile("^(GB|gb|Gb|gB)[\\d]+$")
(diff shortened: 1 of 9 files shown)


app/utils/validators/forms/FieldValidator.scala
```diff
@@ -137,7 +137,7 @@ object FieldValidator {
   private val emailPattern = Pattern.compile("""^\S+@\S+$""")
   val isValidEmail: String => Boolean = (name: String) => emailPattern.matcher(name).matches()
 
-  private val eoriPattern = Pattern.compile("^(GB|gb|Gb|gB)[\\d]+$")
+  private val eoriPattern = Pattern.compile("^[a-zA-z]{2}[\\da-zA-z]+$")
```
PatrykRudnicki: Is `a-zA-z` correct? I think it should be `a-zA-Z`. Second `Z` capital

suryend4u: Fixed

PatrykRudnicki: There is no fix for it

PR #712, opened 2020-01-14
CEDS-1788 - further fixes for error messages

SteveSugden:


app/forms/declaration/ModeOfTransportCodes.scala
```diff
@@ -56,9 +58,12 @@ object ModeOfTransportCodes extends DeclarationPage {
 
   implicit val writes: Writes[ModeOfTransportCodes] = Writes(code => JsString(code.value))
 
-  import Forms._
-
-  private val mapping = Forms.single("code" -> of(formatter("error.unknown")))
+  private val mapping = Forms.single(
+    "code" ->
+      requiredRadio("declaration.transportInformation.borderTransportMode.error.empty")
+        .verifying("declaration.transportInformation.borderTransportMode.error.empty", isContainedIn(reverseLookup.keySet))
```
bambuchaAdm: empyt could not be the best error key here.

app/controllers/declaration/SealController.scala
```diff
@@ -55,13 +55,18 @@ class SealController @Inject()(
...
-    request.cacheModel.containerBy(containerId) match {
-      case Some(container) =>
-        saveSeal(mode, boundForm, container)
-      case _ => errorHandler.displayErrorPage()
-    }
+    Seal
+      .form()
+      .bindFromRequest()
+      .fold(
+        (formWithErrors: Form[Seal]) => Future.successful(BadRequest(addPage(mode, formWithErrors, containerId))),
```
PatrykRudnicki: We have this part in `saveSeal` method. It probably means that we don't need to check once again errors in `saveSeal` method

SteveSugden: I think the form errors in saveSeal will be a result of calling saveAndContinue - i.e. list max length validation errors.    So I think both are needed

PR #713, opened 2020-01-14
Ceds 1707

apodi:
Do not allow user to enter 0 values for mass etc.

.validation on input field for no zero value.
.unit test cases fix
.smoke test passed

Location url : http://localhost:6791/customs-declare-exports/declaration/items/9aecb69g/commodity-measure


app/forms/declaration/CommodityMeasure.scala
```diff
@@ -28,13 +28,15 @@ object CommodityMeasure {
...
+      text().verifying("supplementary.commodityMeasure.supplementaryUnits.error", validateDecimal(16)(6) and containsNotOnlyZeros)
+    ),
     "netMass" -> text
       .verifying("supplementary.commodityMeasure.netMass.empty", nonEmpty)
-      .verifying("supplementary.commodityMeasure.netMass.error", isEmpty or validateDecimal(11)(3)),
+      .verifying("supplementary.commodityMeasure.netMass.error", isEmpty or validateDecimal(16)(6) and containsNotOnlyZeros),
     "grossMass" -> text
       .verifying("supplementary.commodityMeasure.grossMass.empty", nonEmpty)
-      .verifying("supplementary.commodityMeasure.grossMass.error", isEmpty or validateDecimal(16)(2))
+      .verifying("supplementary.commodityMeasure.grossMass.error", isEmpty or validateDecimal(16)(6) and containsNotOnlyZeros)
```
PatrykRudnicki: Why you changed `validateDecimal` values?

apodi: we validated with BA and found that error message and accordingly implementation was not correct . we should allow user to enter 6 digits after decimal instead of 2. 

is it okay?

PatrykRudnicki: Yes, but if you can please ask BA to update a ticket with those information. It can be necessary for testing purposes.

apodi: Done

PR #716, opened 2020-01-15
CEDS-2117 - remove duplicate error validation messages

SteveSugden:
For this ticket we identified fields that had more than one validation message key which translated to the same statement.  Once found, the keys were de-duplicated and the validation changed to show the message if any of the validation rules were breached.


app/forms/declaration/ContainerFirst.scala
```diff
@@ -61,8 +61,7 @@ object ContainerFirst extends DeclarationPage {
       yes,
       text()
         .verifying("declaration.transportInformation.containerId.empty", nonEmpty)
-        .verifying("declaration.transportInformation.containerId.error.alphanumeric", isAlphanumeric)
-        .verifying("declaration.transportInformation.containerId.error.length", noLongerThan(maxContainerIdLength))
+        .verifying("declaration.transportInformation.containerId.error.invalid", isAlphanumeric and noLongerThan(maxContainerIdLength))
```
PatrykRudnicki: Here probably `isEmpty` is missing. In case that the input will be empty we can have two errors (Not sure that `isAlphanumeric` return `true` or `false` for empty input)

app/forms/declaration/DepartureTransport.scala
```diff
@@ -35,8 +35,10 @@ object DepartureTransport extends DeclarationPage {
       .verifying("declaration.transportInformation.meansOfTransport.departure.error.incorrect", isContainedIn(allowedMeansOfTransportTypeCodes)),
     "meansOfTransportOnDepartureIDNumber" -> text()
       .verifying("declaration.transportInformation.meansOfTransport.reference.error.empty", nonEmpty)
-      .verifying("declaration.transportInformation.meansOfTransport.reference.error.length", noLongerThan(27))
-      .verifying("declaration.transportInformation.meansOfTransport.reference.error.invalid", isAlphanumericWithAllowedSpecialCharacters)
+      .verifying(
+        "declaration.transportInformation.meansOfTransport.reference.error.invalid",
+        noLongerThan(27) and isAlphanumericWithAllowedSpecialCharacters
```
PatrykRudnicki: Similar like above

app/forms/declaration/ContainerFirst.scala
SteveSugden: Seems to work without it.  But I'll add it to be consistent.  (For both comments)

PR #717, opened 2020-01-15
CEDS-2052 Back button navigation on change from summary page 

bambuchaAdm:
This is first PR. Works till goods location. It is enough big now and it still work to do.


bambuchaAdm:
please rebuild

bambuchaAdm:
retest this please

bambuchaAdm:
rebuild this please

app/controllers/navigation/Navigator.scala
```diff
@@ -41,6 +43,17 @@ class Navigator @Inject()(appConfig: AppConfig, auditService: AuditService) {
       case _ => Results.Redirect(call)
     }
 
+  def continueTo(mode: Mode, factory: Mode => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
+    redirectTo(mode.next, factory)
```
PatrykRudnicki: Why we need `continueTo` if it's just a wrapper for `redirectTo`?

app/controllers/navigation/Navigator.scala
```diff
@@ -182,21 +195,52 @@ object Navigator {
     case page                  => throw new IllegalArgumentException(s"Navigator back-link route not implemented for $page on occasional")
   }
 
+  val universal: PartialFunction[DeclarationPage, Mode => Call] = {
```
PatrykRudnicki: Can we use the better name for it? If I'm understanding this correctly what do you think about `common`?

app/controllers/navigation/Navigator.scala
bambuchaAdm: `mode.next` - redirect preserve mode - continue change it

app/controllers/navigation/Navigator.scala
bambuchaAdm: Would be fixed in part2

PR #719, opened 2020-01-16
Regression issue fixed

apodi:
* Allow user to enter zero after decimal e.g 31.0 or 31.00
 * Unit test cases

### DIFF
--- a/app/forms/declaration/CommodityMeasure.scala
+++ b/app/forms/declaration/CommodityMeasure.scala
@@ -29,14 +29,14 @@ object CommodityMeasure {
 
   val mapping = Forms.mapping(
     "supplementaryUnits" -> optional(
-      text().verifying("supplementary.commodityMeasure.supplementaryUnits.error", validateDecimalGreaterThanZero(17)(7) and containsNotOnlyZeros)
+      text().verifying("supplementary.commodityMeasure.supplementaryUnits.error", validateDecimalGreaterThanZero(16)(6) and containsNotOnlyZeros)
     ),
     "grossMass" -> text
       .verifying("supplementary.commodityMeasure.grossMass.empty", nonEmpty)
-      .verifying("supplementary.commodityMeasure.grossMass.error", isEmpty or validateDecimalGreaterThanZero(17)(7) and containsNotOnlyZeros),
+      .verifying("supplementary.commodityMeasure.grossMass.error", isEmpty or validateDecimalGreaterThanZero(16)(6) and containsNotOnlyZeros),
     "netMass" -> text
       .verifying("supplementary.commodityMeasure.netMass.empty", nonEmpty)
-      .verifying("supplementary.commodityMeasure.netMass.error", isEmpty or validateDecimalGreaterThanZero(17)(7) and containsNotOnlyZeros)
+      .verifying("supplementary.commodityMeasure.netMass.error", isEmpty or validateDecimalGreaterThanZero(16)(6) and containsNotOnlyZeros)
   )(CommodityMeasure.apply)(CommodityMeasure.unapply)
 
   def form(): Form[CommodityMeasure] = Form(mapping)
--- a/app/forms/declaration/CommodityMeasure.scala
+++ b/app/forms/declaration/CommodityMeasure.scala
@@ -29,14 +29,14 @@ object CommodityMeasure {
 
   val mapping = Forms.mapping(
     "supplementaryUnits" -> optional(
-      text().verifying("supplementary.commodityMeasure.supplementaryUnits.error", validateDecimal(16)(6) and containsNotOnlyZeros)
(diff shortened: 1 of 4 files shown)


app/utils/validators/forms/FieldValidator.scala
```diff
@@ -115,10 +115,15 @@ object FieldValidator {
...
-          case _                                           => false
+        input.trim().endsWith(".") match {
+          case false =>
+            input.trim().split('.') match {
+              case Array(a, b) if isNumeric(a) && isNumeric(b) =>
+                b.length <= decimalPlaces && (a + b).length <= totalLength && (b.toDouble > 0 || a.toDouble > 0)
+              case Array(a) if isNumeric(a) && nonEmpty(a) => a.length <= totalLength && a.toDouble > 0
+              case _                                       => false
+            }
+          case true => false
```
PatrykRudnicki: Using `true` `false` in pattern matching is not a good approach. In that case `if` is better.

Additionally I think we shouldn't trim value for validating. We're trimming this just for the validation actually, but the original value can have the spaces and fail on the xml validation, because in practice we can have more characters than allowed.

Moreover this validation is really complex and cases like `case true => false` are not enough described to understand it.

I'll come back to totally first implementation for `validateDecimal` or go with @PauloPortugal idea to validate it using `BigDecimal`

PatrykRudnicki: Additionally `0` is a correct `BigDecimal`. This file contains general validation helper. We should add validation for zero just in specific case where we want to use it.

apodi: Done please review

app/utils/validators/forms/FieldValidator.scala
```diff
@@ -116,11 +116,18 @@ object FieldValidator {
...
+          case Array(a, b) if isNumeric(a) && isNumeric(b) => b.length <= decimalPlaces && (a + b).length <= totalLength
+          case Array(a) if isNumeric(a)                    => a.length <= totalLength
           case _                                           => false
   }
 
+  val validateDecimalGreaterThanZero: Int => Int => String => Boolean = (totalLength: Int) =>
+    (decimalPlaces: Int) =>
+      (input: String) =>
+        try { BigDecimal(input.trim()) > 0 && BigDecimal(input.trim()).scale < decimalPlaces && input.trim().replace(".", "").length < totalLength } catch {
+          case e: java.lang.NumberFormatException => false
```
PatrykRudnicki: Still IMO we shouldn't trim value. Please 
```
try { 
       BigDecimal(input) > 0 && 
       BigDecimal(input).scale <= decimalPlaces && 
       input.length <= totalLength 
} catch {
          case e: java.lang.NumberFormatException => false
}
```

app/forms/declaration/CommodityMeasure.scala
```diff
@@ -29,14 +29,14 @@ object CommodityMeasure {
 
   val mapping = Forms.mapping(
     "supplementaryUnits" -> optional(
-      text().verifying("supplementary.commodityMeasure.supplementaryUnits.error", validateDecimal(16)(6) and containsNotOnlyZeros)
+      text().verifying("supplementary.commodityMeasure.supplementaryUnits.error", validateDecimalGreaterThanZero(17)(7) and containsNotOnlyZeros)
     ),
     "grossMass" -> text
       .verifying("supplementary.commodityMeasure.grossMass.empty", nonEmpty)
-      .verifying("supplementary.commodityMeasure.grossMass.error", isEmpty or validateDecimal(16)(6) and containsNotOnlyZeros),
+      .verifying("supplementary.commodityMeasure.grossMass.error", isEmpty or validateDecimalGreaterThanZero(17)(7) and containsNotOnlyZeros),
```
PatrykRudnicki: In all cases it should be `(16)(6)`. Because We should use `<=` in the validation method.

app/utils/validators/forms/FieldValidator.scala
PauloPortugal: I agree with what Patryk said. 
The `trim` was my suggestion. My fault guys! :)

PauloPortugal: I would also format the code in the way Patryk did because it is easy for us to read the code and reason about it, than putting everything in one line.

app/forms/declaration/CommodityMeasure.scala
PauloPortugal: Patryk's suggestion is a very good one. Let's use the same data/numbers as the XML Schema, to avoid confusion in the future, between business analysts, developers and QAs

PauloPortugal: Using the `17` was also my suggestion :stuck_out_tongue:

app/utils/validators/forms/FieldValidator.scala
apodi: done

app/forms/declaration/CommodityMeasure.scala
apodi: done

PR #720, opened 2020-01-17
* Updated *

leszeklorek:
- changes to quick submit script (added declarations based on data from Khurram)
- except both supplementary and clearance prelodged (K), everything is working and produce either recevied or accepted/cleared decs on QA
- the script is a bit hackey, there are possible ways to improve it, but it is additional thing
- also fixed the small typo in one of our routes (possibly it could affect performance tests)

### DIFF
--- a/docs/Customs Declare Exports AutoComplete.js
+++ b/docs/Customs Declare Exports AutoComplete.js
@@ -1,8 +1,8 @@
 // ==UserScript==
 // @name         Customs Declare Exports AutoComplete
 // @namespace    http://tampermonkey.net/
-// @version      1.20
-// @description  try to take over the world!
+// @version      1.21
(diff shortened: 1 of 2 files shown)


docs/Customs Declare Exports AutoComplete.js
```diff
@@ -42,8 +43,6 @@ function dropDown() {
     var selectList = document.createElement("select");
     selectList.style.position = "absolute"
     selectList.style.top = "100px"
-    selectList.style.right = "5px"
-    selectList.style.zIndex = "999"
```
PatrykRudnicki: This has been added by our designers. Are you sure that it's not necessary?

leszeklorek: it works fine, for me, but we will see :)

PR #730, opened 2020-01-22
CEDS-1757: Fix for right order of error messages for city and post code

suryend4u:
CEDS-1757: Fix for right order of error messages for city and post code

### DIFF
--- a/app/forms/declaration/GoodsLocation.scala
+++ b/app/forms/declaration/GoodsLocation.scala
@@ -29,8 +29,8 @@ case class GoodsLocation(
   identificationOfLocation: Option[String],
   additionalIdentifier: Option[String],
   addressLine: Option[String],
-  postCode: Option[String],
   city: Option[String],
+  postCode: Option[String],
   country: String
 )
 
--- a/app/forms/declaration/GoodsLocation.scala
+++ b/app/forms/declaration/GoodsLocation.scala
@@ -57,12 +57,12 @@ object GoodsLocation extends DeclarationPage {
     "addressLine" -> optional(
       text().verifying("supplementary.goodsLocation.addressLine.error", isAlphanumericWithAllowedSpecialCharacters and noLongerThan(70))
     ),
-    "postCode" -> optional(
-      text().verifying("supplementary.goodsLocation.postCode.error", isAlphanumericWithAllowedSpecialCharacters and noLongerThan(9))
-    ),
     "city" -> optional(
       text()
         .verifying("supplementary.goodsLocation.city.error", isAlphanumericWithAllowedSpecialCharacters and noLongerThan(35))
+    ),
+    "postCode" -> optional(
+      text().verifying("supplementary.goodsLocation.postCode.error", isAlphanumericWithAllowedSpecialCharacters and noLongerThan(9))
     ),
     "country" ->
       text()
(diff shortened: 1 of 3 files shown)


test/views/declaration/LocationViewSpec.scala
```diff
@@ -302,8 +329,8 @@ class LocationViewSpec extends UnitViewSpec with ExportsTestData with Stubs with
             Some(TestHelper.createRandomAlphanumericString(36)),
             Some(TestHelper.createRandomAlphanumericString(4)),
             Some(TestHelper.createRandomAlphanumericString(71)),
-            Some(TestHelper.createRandomAlphanumericString(10)),
             Some(TestHelper.createRandomAlphanumericString(36)),
+            Some(TestHelper.createRandomAlphanumericString(10)),
```
MaciejRewera: Here you also have the same pattern as in the cases above. In my opinion it would be better to use `bind` method and field names.
Actually, here you should not switch the lines, as you have not changed the order in the constructor.

test/views/declaration/LocationViewSpec.scala
```diff
@@ -347,8 +374,8 @@ class LocationViewSpec extends UnitViewSpec with ExportsTestData with Stubs with
             None,
             Some(TestHelper.createRandomAlphanumericString(33)),
             Some(TestHelper.createRandomAlphanumericString(71)),
-            Some(TestHelper.createRandomAlphanumericString(10)),
             Some(TestHelper.createRandomAlphanumericString(36)),
+            Some(TestHelper.createRandomAlphanumericString(10)),
```
MaciejRewera: Same as above

test/views/declaration/LocationViewSpec.scala
suryend4u: Fixed

test/views/declaration/LocationViewSpec.scala
suryend4u: Fixed

app/forms/declaration/GoodsLocation.scala
```diff
@@ -57,13 +57,13 @@ object GoodsLocation extends DeclarationPage {
...
-      text().verifying("supplementary.goodsLocation.postCode.error", isAlphanumericWithAllowedSpecialCharacters and noLongerThan(9))
-    ),
     "city" -> optional(
       text()
         .verifying("supplementary.goodsLocation.city.error", isAlphanumericWithAllowedSpecialCharacters and noLongerThan(35))
     ),
+    "postCode" -> optional(
+      text().verifying("supplementary.goodsLocation.postCode.error", isAlphanumericWithAllowedSpecialCharacters and noLongerThan(9))
+    ),
     "country" ->
```
PatrykRudnicki: This "postCode" is just a name to reuse it in forms. If you changed the order I think that now `postCode` will be stored in the `city` place. At the end it's just `GoodsLocation.apply`.

suryend4u: We checked the CYA screen and came back from there. PostCode & city are getting stored in respective places only.

suryend4u: But DB suggest you are right. We will change the constructor too now.

PatrykRudnicki: Yes, the UI will work similar, because you're saving and reading from the same wrong place, but the declaration and model in database will be wrong

suryend4u: Fixed

suryend4u: Please review and approve

PR #734, opened 2020-01-27
BAU - fix for transport summary section where questions where shown

SteveSugden:
that hadn't been asked.  Bug caused by transposing `meansOfTransportCrossingTheBorder` and `meansOfTransportOnDeparture`.  Bug only revealed by change that caused one of the pages to be skipped (they are normally both shown so the summary was correctly showing the answers to both pages).


alsahid1993:
Awaiting pull request to be merged so that test case can be updated

PR #750, opened 2020-02-27
CEDS-2194 - Skip 5/14 Country of Dispatch for C21

iservport:
For Clearance Requests, data element '5/14 Country of Dispatch' is not required in any circumstances.

### DIFF
--- a/app/controllers/navigation/Navigator.scala
+++ b/app/controllers/navigation/Navigator.scala
@@ -102,12 +102,12 @@ object Navigator {
   val clearance: PartialFunction[DeclarationPage, Mode => Call] = {
     case ContainerFirst              => controllers.declaration.routes.DepartureTransportController.displayPage
     case ContainerAdd                => controllers.declaration.routes.TransportContainerController.displayContainerSummary
-    case Document                    => controllers.declaration.routes.NatureOfTransactionController.displayPage
-    case DestinationCountryPage      => controllers.declaration.routes.OriginationCountryController.displayPage
+    case Document                    => controllers.declaration.routes.OfficeOfExitController.displayPage
+    case DestinationCountryPage      => controllers.declaration.routes.DeclarationHolderController.displayPage
     case RoutingQuestionPage         => controllers.declaration.routes.DestinationCountryController.displayPage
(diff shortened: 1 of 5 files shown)


SteveSugden:
Is that it?  No tests? (Either fixed or new).
I was expecting to see the affected controllers being modified to indicate which journey types are valid, but looking at the code, we don't seem to do that in many places, which is a shame.   See for example https://github.com/hmrc/customs-declare-exports-frontend/blob/fa6854647e3634c3e0d06e23d60651d530fadc4c/app/controllers/declaration/CarrierDetailsController.scala#L47

At the very least you'll have to fix the acceptance tests I guess.

SteveSugden:
Also (before anyone else comments) you need to fix the back-button behaviour which will mean a change in the Navigator class.

bambuchaAdm:
Exactly as @SteveSugden said:
+ missing navigation fixes - on back on next page you would back to page you try to skip
and this means please fix tests for this controller they don't cover different type of journey

app/controllers/navigation/Navigator.scala
```diff
@@ -102,13 +102,12 @@ object Navigator {
...
     case ContainerAdd                => controllers.declaration.routes.TransportContainerController.displayContainerSummary
-    case Document                    => controllers.declaration.routes.OfficeOfExitController.displayPage
-    case OriginationCountryPage      => controllers.declaration.routes.DeclarationHolderController.displayPage
+    case Document                    => controllers.declaration.routes.NatureOfTransactionController.displayPage
     case DestinationCountryPage      => controllers.declaration.routes.OriginationCountryController.displayPage
     case RoutingQuestionPage         => controllers.declaration.routes.DestinationCountryController.displayPage
     case RemoveCountryPage           => controllers.declaration.routes.RoutingCountriesSummaryController.displayPage
     case ChangeCountryPage           => controllers.declaration.routes.RoutingCountriesSummaryController.displayPage
-    case GoodsLocation               => controllers.declaration.routes.DestinationCountryController.displayPage
+    case GoodsLocation               => controllers.declaration.routes.DeclarationHolderController.displayPage
```
PatrykRudnicki: Hmmm, why this change? I think that Goods Location should redirect to the `DestinationCountryController`.

The redirection from Destination Country should be to Declaration Holder

PR #751, opened 2020-02-27
CEDS-2195 - Skip 8/5 'Nature of Transaction' for Clearance Requests

SteveSugden:
CEDS-2188 - Skip 6/18 'Total number of packages' from Clearance Requests

### DIFF
--- a/app/controllers/declaration/OfficeOfExitController.scala
+++ b/app/controllers/declaration/OfficeOfExitController.scala
@@ -25,7 +25,7 @@ import models.requests.JourneyRequest
 import models.{DeclarationType, ExportsDeclaration, Mode}
 import play.api.data.Form
 import play.api.i18n.I18nSupport
-import play.api.mvc.{Action, AnyContent, Call, MessagesControllerComponents, Result}
+import play.api.mvc._
 import play.twirl.api.Html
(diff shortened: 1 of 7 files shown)


app/controllers/declaration/OfficeOfExitController.scala
```diff
@@ -106,9 +106,7 @@ class OfficeOfExitController @Inject()(
     declarationType match {
       case DeclarationType.SUPPLEMENTARY | DeclarationType.STANDARD =>
         controllers.declaration.routes.TotalNumberOfItemsController.displayPage
-      case DeclarationType.CLEARANCE =>
-        controllers.declaration.routes.TotalPackageQuantityController.displayPage
-      case DeclarationType.SIMPLIFIED | DeclarationType.OCCASIONAL =>
+      case DeclarationType.SIMPLIFIED | DeclarationType.OCCASIONAL | DeclarationType.CLEARANCE =>
```
PatrykRudnicki: Is this correct place? Based on the navigator change there is a `NatureOfTheTransactionController`. Shouldn't this change be in place where previously we redirected to this controller?

PatrykRudnicki: It's discussed. This is correct

PR #753, opened 2020-02-28
CEDS-2180 Generate PDF with 128-Barcode based on MRN

PauloPortugal:
* An Export Accompanying Document ('EAD') is a short summary in a
   format agreed access the EU.
 * A lorry driver needs to have a physical copy of an EAD with them
   as they cross a border from the from an EU state to a non-EU state.
 * Declarants need to be able to generate an EAD as a PDF, which they
   can send electronically (or physically) to a driver.

**NOTE:**
This is a version using a XSL template.
@bambuchaAdm enjoyed https://pdfbox.apache.org/ API and is going to create an investigative spike to compare it. The test of fire will be once we have the final template with the DIS data

### DIFF
--- a/app/controllers/pdf/EADController.scala
+++ b/app/controllers/pdf/EADController.scala
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.pdf
+
+import java.time.LocalDate
+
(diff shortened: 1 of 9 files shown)


PauloPortugal:
> Looks good so far, just one observation...
> Would all this image/pdf generating code not be better in a service rather than polluting the controller?

Oh completely of board with that. This is a tactical change, because we do not know if we need to create another PRA, SRA, CAB, TxM, etc.

So we do the work to deliver the programme feature and we raise tech debt
This was discussed with @PatrykRudnicki Damien.

I would have gone with two microservices from the get go: 
 * PDF/Barcode generator microservice
 * DIS Integration Microservice

PauloPortugal:
@SteveSugden Could you please review it again?

app/controllers/pdf/EADController.scala
```diff
@@ -0,0 +1,52 @@
...
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+import views.xml.pdf.pdfTemplate
+
+class EADController @Inject()(pdfTemplate: pdfTemplate, eadService: EADService)(mcc: MessagesControllerComponents, val playFop: PlayFop)
+    extends FrontendController(mcc) with I18nSupport {
+
+  private val mimeType: String = MimeConstants.MIME_PDF
+
+  def generatePdf(mrn: String) = Action { implicit request =>
+    {
```
PatrykRudnicki: Those brackets are unnecessary

PauloPortugal: Me and my Java brackets!
I will pay this off once we integrate DIS into EAD :)

PR #754, opened 2020-03-02
CEDS-2147 - restore 'Seals' to Clearance journey

SteveSugden:
This effectively reverses https://github.com/hmrc/customs-declare-exports-frontend/pull/721


SteveSugden:
Mainly the removal of "special" views and tests I guess.  Take a look at the PR to remove the seals.

app/controllers/declaration/TransportContainerController.scala
```diff
@@ -179,12 +162,7 @@ class TransportContainerController @Inject()(
       .bindFromRequest()
       .fold(
         (formWithErrors: Form[YesNoAnswer]) =>
-          request.declarationType match {
-            case CLEARANCE =>
-              Future.successful(BadRequest(clearanceRemovePage(mode, formWithErrors, request.cacheModel.containers.filter(_.id == containerId).head)))
-            case _ =>
-              Future.successful(BadRequest(removePage(mode, formWithErrors, request.cacheModel.containers.filter(_.id == containerId).head)))
-        },
+          Future.successful(BadRequest(removePage(mode, formWithErrors, request.cacheModel.containers.filter(_.id == containerId).head))),
```
PatrykRudnicki: I know it's an old implementation but instead of `head` I think it'll be better to use `headOption` and handle case if it's empty

SteveSugden: Not looked at the code in detail ,but I suspect head being empty would be a "bad thing" - attempting to remove a container that doesn't exist.  So we'd have the decide what to do if we didn't want the default behaviour of the exception.

PR #759, opened 2020-03-04
CEDS-1771 Add tariff link for two new pages

apodi:
* Missing on Office of exit page in supplementary directory

### DIFF
--- a/app/views/declaration/office_of_exit_supplementary.scala.html
+++ b/app/views/declaration/office_of_exit_supplementary.scala.html
@@ -27,6 +27,10 @@
 
 @(mode: Mode, form: Form[OfficeOfExitSupplementary])(implicit request: JourneyRequest[_], messages: Messages)
 
+@tariffBodyDesc = { @Html(messages("declaration.type.consignmentTariffText", components.details_content_link())) }
+
+@tariffBody = { @components.tariff_body_partial(messages("supplementary.representative-details.help-item1"), messages("supplementary.representative-details.help-item2"))(tariffBodyDesc)}
+
 @main_template(
     title = Title("declaration.officeOfExit.title", "declaration.summary.locations.header")
 ) {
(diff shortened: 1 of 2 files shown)


app/views/declaration/office_of_exit_supplementary.scala.html
```diff
@@ -49,7 +53,9 @@
                   messages("standard.officeOfExit.empty"),
                   AutoCompleteItem.fromOfficeOfExit(OfficeOfExits.all)
               )
-
+        @components.details(messages("site.details.summary_text_these")){
```
PatrykRudnicki: Probably I missed this last time, but we're not using `_` in the messages keys.

I'm not saying this is wrong, but it's just probably different convention.

Can we change `summary_text_these` to `summaryTextThese`?

Additionally what this `these` mean in this message key?

test/views/declaration/OfficeOfExitSupplementaryViewSpec.scala
```diff
@@ -79,6 +80,12 @@ class OfficeOfExitSupplementaryViewSpec extends UnitViewSpec with ExportsTestDat
       val saveAndReturnButton = view.getElementById("submit_and_return")
       saveAndReturnButton.text() mustBe "site.save_and_come_back_later"
     }
+
+    "display Tariff section text" in {
+      val tariffText = view.getElementsByClass("govuk-details__summary-text").first().text()
+      tariffText.text() must be("site.details.summary_text_these")
```
PatrykRudnicki: Just to be consistent with rest of the test cases, we preffer to use `mustBe` not `must be`.

Similar, it's not a mistake, but always we're trying to keep one approach to be consistent

app/views/declaration/office_of_exit_supplementary.scala.html
apodi: I used _ convention beacause already we are using this in lot of places in messages.en
for example :
site.save_and_come_back_later = Save and come back later

we have two different summary link message
site.details.summary_text_this = For more information about this
site.details.summary_text_these = For more information about these
one of this and other for these

PatrykRudnicki: I understand, this convention `site.save_and_come_back_later = Save and come back later` is probably autogenerated by old library, but it's okay

test/views/declaration/OfficeOfExitSupplementaryViewSpec.scala
apodi: Fixed

PR #761, opened 2020-03-10
CEDS-2213 - Tariff Links missing on Additional Fiscal References page

iservport:


app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -41,6 +41,7 @@
 @import views.Title
 @import controllers.navigation.{Navigator, ItemId}
 @import models.requests.JourneyRequest
+@import views.html.components.submit_button
```
SteveSugden: Do you need this import?

iservport: No

PR #768, opened 2020-03-12
CEDS-2171 Integrate Customs Declaration Information service

PauloPortugal:
with @apodi & @PauloPortugal


app/config/AppConfig.scala
```diff
@@ -62,6 +62,27 @@ class AppConfig @Inject()(
   lazy val classificationHelpUrl = loadConfig("urls.classificationHelp")
   lazy val ecicsToolUrl = loadConfig("urls.ecicsTool")
 
+  lazy val customsDeclarationsInformation = servicesConfig.baseUrl("customs-declarations-information")
+  lazy val fetchMrnStatus = servicesConfig.getConfString(
```
SteveSugden: Could maybe have a helper method to get services config or throw exception - like `loadConfig`?

app/models/dis/parsers/MrnStatusParser.scala
```diff
@@ -0,0 +1,73 @@
...
+    documents.foreach { doc =>
+      val id = (doc \ XmlTags.id).text
+      val typeCode = (doc \ XmlTags.typeCode).text
+      previousDocuments :+= PreviousDocument(id, typeCode)
+    }
+    previousDocuments
+  }
+
+  private def timeFormatter(zonedDateTime: String): String = timeFormatter(
+    ZonedDateTime.parse(zonedDateTime, DateTimeFormatter.ofPattern("yyyyMMddHHmmssz"))
```
SteveSugden: I'm fairly sure it's OK to create local constants for these DateTimeFormatters - they are thread-safe

app/models/dis/parsers/MrnStatusParser.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private def previousDocuments(documents: NodeSeq): Seq[PreviousDocument] = {
+    var previousDocuments: Seq[PreviousDocument] = Seq.empty
+    documents.foreach { doc =>
+      val id = (doc \ XmlTags.id).text
+      val typeCode = (doc \ XmlTags.typeCode).text
+      previousDocuments :+= PreviousDocument(id, typeCode)
+    }
+    previousDocuments
```
PatrykRudnicki: Please not use `var`

```
documents.map { doc =>
      val id = (doc \ XmlTags.id).text
      val typeCode = (doc \ XmlTags.typeCode).text
      PreviousDocument(id, typeCode)
    }
```

alsahid1993: this has been addressed

app/models/dis/parsers/MrnStatusParser.scala
alsahid1993: this has been addressed

app/config/AppConfig.scala
alsahid1993: as discussed with Steve, this may not be necessary

SteveSugden: True - it was not a blocker

bambuchaAdm: I would check here is address is correct - ends with slash

PR #773, opened 2020-03-13
CEDS-2243 Add feature flag for the EAD Document generation link

PatrykRudnicki:


app/config/EADDocumentConfig.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package config
+
+import javax.inject.{Inject, Singleton}
+import play.api.Configuration
+
+@Singleton
+class EADDocumentConfig @Inject()(configuration: Configuration) {
```
MaciejRewera: Isn't "D" in EAD for "Document" ? NO need to have it twice ;)

conf/application.conf
```diff
@@ -92,6 +92,7 @@ microservice {
     features {
       welsh-translation = false
       use-improved-error-messages = false
+      ead-document = true
```
MaciejRewera: AFAIR for ILE Query we use values enabled/disabled. Could we stay with one approach?
It seems that we may have more feature flags in the future, so it'd be better to be aligned. Not thinking every time "should I write true or enabled?". 
And not enforcing consistency may lead to even more approaches.

test/config/AppConfigSpec.scala
```diff
@@ -82,7 +82,7 @@ class AppConfigSpec extends UnitSpec {
 
       """.stripMargin
     )
-  private val emptyAppConfig: Config = ConfigFactory.parseString("")
+  private val emptyAppConfig: Config = ConfigFactory.empty()
```
MaciejRewera: 👍

test/config/EADDocumentConfigSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+
+        eadConfig.isEadEnabled mustBe false
+      }
+
+      "ead document feature config doesn't exist" in {
+
+        val eadConfig = new EADDocumentConfig(Configuration(emptyConfig))
+
+        eadConfig.isEadEnabled mustBe false
+      }
```
MaciejRewera: 👍

test/views/DeclarationInformationViewSpec.scala
```diff
@@ -53,7 +68,12 @@ class DeclarationInformationViewSpec extends UnitViewSpec with Injector {
 
   private val notifications = Seq(notification, rejectedNotification)
 
-  private val view = declarationInformationPage(submission, notifications)(request, messages)
+  private val declarationInformationPageWithEad = new declaration_information(gdsMainTemplate, govukSummaryList, govukTable, eadDocumentConfigEnabled)
+
+  private val declarationInformationPageWithoutEad =
+    new declaration_information(gdsMainTemplate, govukSummaryList, govukTable, eadDocumentConfigDisabled)
+
+  private val viewWithEad = declarationInformationPageWithEad(submission, notifications)(request, messages)
```
MaciejRewera: Please have a look at `OverridableInjector':
https://github.com/hmrc/customs-movements-frontend/blob/master/test/unit/base/OverridableInjector.scala
This way you could avoid creating instances of views components by hand and only bind `EADDocumentConfig` to specific instance.

test/views/DeclarationInformationViewSpec.scala
```diff
@@ -19,14 +19,29 @@ package views
...
+  private val govukSummaryList = instanceOf[GovukSummaryList]
+  private val govukTable = instanceOf[GovukTable]
+
+  private val configWithEnabledEAD: Config =
+    ConfigFactory.parseString("microservice.services.features.ead-document=true")
+  private val configWithDisabledEAD: Config =
+    ConfigFactory.parseString("microservice.services.features.ead-document=false")
+
+  private val eadDocumentConfigEnabled = new EADDocumentConfig(Configuration(configWithEnabledEAD))
+  private val eadDocumentConfigDisabled = new EADDocumentConfig(Configuration(configWithDisabledEAD))
```
MaciejRewera: Would it make sense to extract it to a separate class or trait? (depending if you consider inheritance or composition to be more applicable here)

conf/application.conf
PatrykRudnicki: I sticked with the true or false, because of the above configs use the same pattern. we're not using enabled disabled in this application. Good point, but if you're happy with it I'll stay with this configuration for the time being

PatrykRudnicki: Changed to enabled disabled

test/views/DeclarationInformationViewSpec.scala
PatrykRudnicki: Hmm, not sure if it's necessary, this flag is just for a link, so the config for it shouldn't be used in different places

app/config/EADConfig.scala
```diff
@@ -0,0 +1,52 @@
...
+    configuration.getOptional[String](key).getOrElse(throw new Exception(s"Missing configuration key: $key"))
+
+  private def feature2Key(feature: Feature): String = s"microservice.services.features.$feature"
+
+  private def str2FeatureStatus(str: String): FeatureStatus = FeatureStatus.withName(str)
+
+  lazy val defaultFeatureStatus: features.FeatureStatus.Value =
+    FeatureStatus.withName(loadConfig(feature2Key(Feature.default)))
+
+  def featureStatus(feature: Feature): FeatureStatus =
```
MaciejRewera: You probably want it to be private. Otherwise you could check any feature using EADConfig class.

app/config/EADConfig.scala
```diff
@@ -0,0 +1,52 @@
...
+package config
+
+import features.Feature.Feature
+import features.{Feature, FeatureStatus}
+import features.FeatureStatus.FeatureStatus
+import javax.inject.{Inject, Singleton}
+import play.api.Configuration
+
+@Singleton
+class EADConfig @Inject()(configuration: Configuration) {
```
MaciejRewera: Nit: Didn't we have agreement not to capitalise the whole acronym, but only the first letter?

app/config/EADConfig.scala
PatrykRudnicki: Good point, changed

app/config/EADConfig.scala
PatrykRudnicki: You're right! Changed

PR #775, opened 2020-03-13
CEDS-2225 - Move Section 1 of the Standard Declaration journey to the Government Design System

SteveSugden:


app/views/components/details.scala.html
```diff
@@ -18,7 +18,7 @@
 
 @(summaryText: String = "")(bodyHtml: Html = HtmlFormat.empty)
 
-<details class="govuk-details details-font" data-module="govuk-details">
+<details id="tariffReference" class="govuk-details details-font" data-module="govuk-details">
```
bambuchaAdm: why `id` - we should use class in component level  ?

app/views/components/gds/saveAsDraft.scala.html
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import controllers.util.SaveAndReturn
+
+@(messageKey: String = "site.save_and_come_back_later")(implicit messages: Messages)
+<br>
```
bambuchaAdm: I am not sure that is good to have structure elements in component.
It is break hermetization from component perspective.

app/views/components/gds/sectionHeader.scala.html
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@this()
+
+@(message: String)
+
+<span id="section-header" class="govuk-caption-xl">@message</span>
```
bambuchaAdm: Same here - id makes it less reusable

app/views/declaration/declaration_choice.scala.html
```diff
@@ -39,32 +44,65 @@
...
-) {
-
-    @components.back_link(backCall, mode)
+@govukLayout(
+    title = Title("declaration.type.description"),
+    backButton = Some(BackButton(messages("site.backToSelectionPage"), backCall))){
 
-    @helper.form(controllers.declaration.routes.DeclarationChoiceController.submitChoice(mode), 'autoComplete -> "off") {
-        @helper.CSRF.formField
+    @formHelper(action = controllers.declaration.routes.DeclarationChoiceController.submitChoice(), 'autoComplete -> "off") {
```
bambuchaAdm: Is mode not missing here ?

SteveSugden: Good spot.  Unfortunately, it looks like no test picked that up

app/views/components/details.scala.html
SteveSugden: This component only does one thing and it's only ever used once on a page.  It's easier to access by a unique id in the acceptance tests,

app/views/declaration/not_eligible.scala.html
```diff
@@ -14,32 +14,43 @@
...
 
-<h1 style="margin-top: 5px;">@messages("notEligible.title")<br>@messages("notEligible.titleLineTwo")</h1>
+<h1 class="govuk-heading-xl">@messages("notEligible.title")</h1>
 
-<p>@messages("notEligible.descriptionPreUrl")
-    <a href="https://secure.hmce.gov.uk/ecom/login/index.html" target="_blank" rel="noopener noreferrer"> @messages("notEligible.descriptionUrl") </a>
-    @messages("notEligible.descriptionPostUrl")
+<p class="govuk-body">
+  @messages("notEligible.description")
+  @messages("notEligible.descriptionPreUrl") <a href="https://www.gov.uk/guidance/dispatching-your-goods-within-the-eu">@messages("notEligible.descriptionUrl")</a>
```
bambuchaAdm: can we refactor this to use parametriesed URL ? It has been used in few places here.
notEligible.description = Some description. {url}
notEligible.urlText = Click

app/views/components/gds/sectionHeader.scala.html
SteveSugden: Similar, although I could see a situation where a page has > 1 section.   I'll have a think/2nd look

app/views/declaration/not_eligible.scala.html
```diff
@@ -14,32 +14,43 @@
...
 </p>
 
-<h3>@messages("notEligible.referenceTitle")</h3>
-
-<p>@messages("notEligible.reference.text")</p>
+<h2 class="govuk-heading-m">@messages("notEligible.referenceTitle")</h2>
+<p class="govuk-body">
+  @messages("notEligible.reference.problems") <b>@messages("notEligible.reference.phone")</b>
```
bambuchaAdm: Iam not sure that phone should be in translation file

app/views/components/gds/saveAsDraft.scala.html
SteveSugden: This component is a bit of a hack to get around the fact that the gds prototyping kit doesn't appear to have a way of rendering a submit button as a link.

Maybe I should render both the "save and continue" and "save as draft" buttons in this component (as they are nearly always used together anyway).    In that case, would the <br> be appropriate?

test/views/declaration/DeclarationChoiceViewSpec.scala
```diff
@@ -117,16 +115,16 @@ class DeclarationChoiceViewSpec extends UnitViewSpec with CommonMessages with St
   }
   private def ensureAllLabelTextIsCorrect(view: Document): Unit = {
     view.getElementsByTag("label").size mustBe 5
-    view.getElementById("STANDARD-label").text() mustBe "declaration.type.standard"
-    view.getElementById("SUPPLEMENTARY-label").text() mustBe "declaration.type.supplementary"
-    view.getElementById("SIMPLIFIED-label").text() mustBe "declaration.type.simplified"
-    view.getElementById("OCCASIONAL-label").text() mustBe "declaration.type.occasional"
-    view.getElementById("CLEARANCE-label").text() mustBe "declaration.type.clearance"
+    view.getElementsByAttributeValue("for", "STANDARD").text() mustBe "declaration.type.standard"
```
bambuchaAdm: :+1: for attribute selector :)

app/views/declaration/not_eligible.scala.html
SteveSugden: Yes, I wasn't happy with this splitting up of the line.   I'll look for other examples

test/views/declaration/DeclarationChoiceViewSpec.scala
SteveSugden: That's what we do in movements to select/test radio options

app/views/declaration/not_eligible.scala.html
SteveSugden: Again you are right.  I should be able to parameterise that same as the urls

app/views/components/details.scala.html
bambuchaAdm: Hmm... - OK
On the other hand - from perspective of component robustness - class would have more pros.
About API in acceptance test - maybe would be worth to access element/s in easier way then this.

app/views/components/gds/saveAsDraft.scala.html
bambuchaAdm: Maybe be having universal `save and continue` and build over it second component `declaration_page_actions` or something like this? We have two birds at once
* we have components that have clear boundaries and could be used outside each other
* we can be sure that each page that have save, have also continue

On the other hand is this should not be secondary button ?

app/views/declaration/not_eligible.scala.html
bambuchaAdm: Feedback page - in [movements](https://github.com/hmrc/customs-movements-frontend/blob/master/conf/messages#L6) but probably same page would be here

app/views/declaration/not_eligible.scala.html
bambuchaAdm: yea, but i think exact phone value should be taken from service configuration - to not rebuild whole service if this would change

PR #776, opened 2020-03-15
CEDS-2225 - Move Section 1 of the Standard Declaration journey to the GDS

SteveSugden:
In addition to pages on ticket, I've updated
- not_eligile page (from answering No to export out of EU question)
- declaration type for ALL journey types as they are implemented as sub-components of the main view

### DIFF
--- a/app/views/components/details_content_link.scala.html
+++ b/app/views/components/details_content_link.scala.html
@@ -16,4 +16,4 @@
 
 @()(implicit messages: Messages)
 
-<a class="govuk-link govuk-link--no-visited-state" target="_blank" href="https://www.gov.uk/government/collections/uk-trade-tariff-volume-3-for-cds"> @messages("declaration.type.tariffUrlText")</a>
+<a target="_blank" href="https://www.gov.uk/government/collections/uk-trade-tariff-volume-3-for-cds"> @messages("declaration.type.tariffUrlText")</a>
--- a/app/views/components/gds/errorSummary.scala.html
+++ b/app/views/components/gds/errorSummary.scala.html
@@ -21,8 +21,9 @@
 @(errors: Seq[FormError], errorFieldName: Option[String] = None)(implicit messages: Messages)
 
 @errorList = @{errors.map { error =>
+    val errorKey = error.key.replace(".", "_").replace('[', '_').replace("]", "")
     ErrorLink(
-        href = Some(s"#${errorFieldName.getOrElse(error.key)}"),
+        href = Some(s"#${errorFieldName.getOrElse(errorKey)}"),
         content = Text(messages(error.message, error.args: _*))
     )
 }}
--- a/app/views/components/gds/exportsInputText.scala.html
+++ b/app/views/components/gds/exportsInputText.scala.html
@@ -1,0 +1,39 @@
+@*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+
+@this(govukInput: GovukInput)
+
+@(field: Field, labelKey: String, hintKey: Option[String] = None, isPageHeading: Boolean = false, headingClasses: String = "govuk-label--xl", labelClasses: String = "govuk-label--m")(implicit messages: Messages)
+
+@buildLabel = @{
+    if(isPageHeading) {
+        Label(content = Text(messages(labelKey)), isPageHeading = true, classes = headingClasses)
+    } else {
+        Label(content = Text(messages(labelKey)), classes = labelClasses)
+    }
+}
+
+@govukInput(Input(
+    id = field.id,
+    name = field.name,
+    value = field.value,
+    label = buildLabel,
+    hint = hintKey.map(key => Hint(content = Text(messages(key)))),
+    errorMessage = field.error.map(err => ErrorMessage(content = Text(messages(err.message)))),
+    classes = "govuk-!-width-two-thirds"
+))
--- a/app/views/components/gds/exportsInputText.scala.html
+++ b/app/views/components/gds/exportsInputText.scala.html
@@ -15,10 +15,11 @@
  *@
 
 @import uk.gov.hmrc.govukfrontend.views.html.components._
+@import views.components.gds.Styles._
 
 @this(govukInput: GovukInput)
 
-@(field: Field, labelKey: String, hintKey: Option[String] = None, isPageHeading: Boolean = false, headingClasses: String = "govuk-label--xl", labelClasses: String = "govuk-label--m")(implicit messages: Messages)
+@(field: Field, labelKey: String, hintKey: Option[String] = None, isPageHeading: Boolean = false, headingClasses: String = gdsPageLabel, labelClasses: String = "govuk-label--m")(implicit messages: Messages)
 
 @buildLabel = @{
     if(isPageHeading) {
--- a/app/views/components/gds/saveButtons.scala.html
+++ b/app/views/components/gds/saveButtons.scala.html
@@ -14,7 +14,15 @@
  * limitations under the License.
  *@
 
-@import controllers.util.SaveAndReturn
+@import views.html.components.gds.saveAndContinue
+@import views.html.components.gds.saveAsDraft
 
-@(messageKey: String = "site.save_and_come_back_later")(implicit messages: Messages)
-<button id="submit_and_return" class="button-link govuk-link" name="@SaveAndReturn.toString">@messages(messageKey)</button>
+@this(
+        saveAndContinue: saveAndContinue
+)
+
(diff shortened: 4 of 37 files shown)


app/views/components/details_content_link.scala.html
```diff
@@ -16,4 +16,4 @@
 
 @()(implicit messages: Messages)
 
-<a target="_blank" href="https://www.gov.uk/government/collections/uk-trade-tariff-volume-3-for-cds"> @messages("declaration.type.tariffUrlText")</a>
\ No newline at end of file
+<a class="govuk-link govuk-link--no-visited-state" target="_blank" href="https://www.gov.uk/government/collections/uk-trade-tariff-volume-3-for-cds"> @messages("declaration.type.tariffUrlText")</a>
```
MaciejRewera: Since this component is using GDS, could you move it to `gds` package?

app/views/components/gds/errorSummary.scala.html
```diff
@@ -21,8 +21,9 @@
 @(errors: Seq[FormError], errorFieldName: Option[String] = None)(implicit messages: Messages)
 
 @errorList = @{errors.map { error =>
+    val errorKey = error.key.replace(".", "_").replace('[', '_').replace("]", "")
```
MaciejRewera: 👍 Something we did not have in Movements ;-)

app/views/components/details_content_link.scala.html
SteveSugden: That's a good idea.   However, to avoid having to change all the existing pages that reference this component, how about I "clone" it to the gds package (with the new gds classes) and then we delete this version when it's no longer needed?

app/views/components/gds/errorSummary.scala.html
SteveSugden: I copied that from movements, once I had worked out why the error links weren't working for the 'ducr.ducr' field.

app/views/components/gds/saveButtons.scala.html
```diff
@@ -0,0 +1,25 @@
...
+
+@this(
+        saveAndContinue: saveAndContinue
+)
+
+@(saveAndContinueKey: String = "site.save_and_continue", saveAsDraftKey: String = "site.save_and_come_back_later")(implicit messages: Messages)
+
+@saveAndContinue(saveAndContinueKey)
+<br>
+@saveAsDraft(saveAsDraftKey)
```
MaciejRewera: Is this working as expected? `saveAsDraft` is not injected.

SteveSugden: It is.  I think it must be because saveAsDraft does not contain any injected GDS components.

app/views/declaration/dispatch_location.scala.html
```diff
@@ -14,72 +14,83 @@
...
+)
 
 @(mode: Mode, form: Form[DispatchLocation])(implicit request: JourneyRequest[_], messages: Messages)
 
-@tariffBodyDesc = { @Html(messages("declaration.type.dispatchLocationSummaryText", components.details_content_link())) }
-
-@tariffBody = {
-<p>
-    <span class="bold">@messages("supplementary.dispatchLocation.summary.euCountries") </span>
+@countriesDetails = {
```
MaciejRewera: Is this piece unique or could be extracted into a component and reused?

test/views/declaration/ConsignmentReferencesViewSpec.scala
```diff
@@ -71,14 +71,14 @@ class ConsignmentReferencesViewSpec extends UnitViewSpec with CommonMessages wit
...
-        createView().getElementById("section-header").text() must include(messages("supplementary.consignmentReferences.heading"))
+        createView().getElementById("section-header").text() must include(messages("declaration.section.1"))
       }
 
       "display empty input with label for DUCR" in {
 
         val view = createView()
 
-        view.getElementById("ducr_ducr-label").text() mustBe messages("supplementary.consignmentReferences.ducr.info")
+        view.getElementsByAttributeValue("for", "ducr_ducr").text() mustBe messages("supplementary.consignmentReferences.ducr.info")
```
MaciejRewera: Have you considered finding elements by class?

test/views/declaration/NotEligibleViewSpec.scala
```diff
@@ -27,51 +26,53 @@ import views.tags.ViewTest
...
+    "display help information" in {
+      view.getElementsByClass("govuk-body").get(0).text() mustBe validatedMsg("notEligible.description", validatedMsg("notEligible.descriptionLink"))
     }
 
-    "display CHIEF link" in {
-      view.select("p:nth-child(3)>a").attr("href") mustBe "https://secure.hmce.gov.uk/ecom/login/index.html"
+    "display read more link" in {
+      view
+        .getElementsByClass("govuk-body")
+        .get(0)
```
MaciejRewera: Nit: There is an alternative method `first()`

test/views/declaration/spec/ViewMatchers.scala
```diff
@@ -140,7 +140,7 @@ trait ViewMatchers {
   class ContainErrorElementWithClassMatcher(tag: String, href: String) extends Matcher[Element] {
     override def apply(left: Element): MatchResult =
       MatchResult(
-        left != null && left.getElementsByClass("govuk-list govuk-error-summary__list").get(0).getElementsByTag(tag).attr("href") == href,
+        left != null && left.getElementsByClass("govuk-list govuk-error-summary__list").get(0).getElementsByTag(tag).eachAttr("href").contains(href),
```
MaciejRewera: 👍

test/views/declaration/additionaldeclarationtype/DeclarationTypeViewSpec.scala
```diff
@@ -181,41 +181,42 @@ class DeclarationTypeViewSpec extends UnitViewSpec with ExportsTestData with Com
...
+    "display header" when {
+      val validated = validatedMessages(request)
 
       "used for Standard Declaration journey" in {
 
-        val view = createView(formStandard, DeclarationType.STANDARD)
+        val view = createView(formStandard, DeclarationType.STANDARD, messages = validated)
 
-        view.getElementById("title").text() mustBe messages("declaration.declarationType.header.standard")
+        view.title() must include(validated("declaration.declarationType.header.standard"))
```
MaciejRewera: This reminds me of `containMessage` matcher used in Movements. Might be worth moving over to Declarations.

test/views/declaration/spec/ViewMatchers.scala
```diff
@@ -140,7 +140,7 @@ trait ViewMatchers {
   class ContainErrorElementWithClassMatcher(tag: String, href: String) extends Matcher[Element] {
```
MaciejRewera: IMHO this name is not accurate.

app/views/components/details_content_link.scala.html
MaciejRewera: Good point.

test/views/declaration/additionaldeclarationtype/DeclarationTypeViewSpec.scala
SteveSugden: Agree.  But in general I've taken the approach of minimal fixes to the tests as I'm relying on the tests to validate that the refactoring of the views is correct.  If we start refactoring the tests as well (more) mistakes might occur.   (Not in this case, but in general).

test/views/declaration/spec/ViewMatchers.scala
SteveSugden: Agree but not part of this ticket.

app/views/components/details_content_link.scala.html
SteveSugden: created new class components.gds.detailsContentLink

test/views/declaration/ConsignmentReferencesViewSpec.scala
SteveSugden: This is specifically testing accessibility i.e. its looking for the element that is the label for a named field.     The pattern is used elsewhere and in movements

app/views/components/gds/exportsInputText.scala.html
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+
+@this(govukInput: GovukInput)
+
+@(field: Field, labelKey: String, hintKey: Option[String] = None, isPageHeading: Boolean = false, headingClasses: String = "govuk-label--xl", labelClasses: String = "govuk-label--m")(implicit messages: Messages)
```
MaciejRewera: Is it correct to be `"govuk-label--xl"` ? I'd expect `"govuk-label--l"`

SteveSugden: I don't this that value is being used (without being over-ridden) on the pages I've done, but good spot.  I'll change it to use the gds.Styles constant and then they can all be easily changed back to "xl" later

app/views/declaration/dispatch_location.scala.html
```diff
@@ -14,72 +14,84 @@
...
+    <b>@messages("supplementary.dispatchLocation.summary.euCountries") </b>
     @Countries.euCountries.mkString(", ")
-</p>
-<p>
-    <span class="bold">@messages("supplementary.dispatchLocation.summary.euSpecialFiscalTerritories") </span>
+    <br><br>
```
MaciejRewera: AFAIR we were usually using some class for this purpose. They were ending with sth like "...margin-9"

SteveSugden: It's a bit of a hack either way.  The prototype uses two 'govuk-details__text' divs but the GDS 'Details' component we are using doesn't seem to support it.

PR #777, opened 2020-03-16
CEDS-2171 Move DIS connection to the backend service

PauloPortugal:
We wanted to create a new Exports DIS microservice that is responsible
for all things EAD/DIS. Due to the governance around a new microservice
we decided to do a tactical approach and implement the integration in
the Exports FE service

Our first approach was to connect straight from the Exports FE to the
`customs-declarations-information` microservice.

```Exports FE -> customs-declaration-info```

Which is was not the best decision, due to who the services and their
connections are configured in the remote environemnts, so we have
decided to refactor the code and push the integration points to the
Exports BE microservice:

```
Exports FE -> Exports BE -> customs-declaration-info
```

### CODE REVIEW

File: app/controllers/pdf/EADController.scala [hunk 1]
```diff
@@ -28,7 +28,7 @@ import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 
 import scala.concurrent.ExecutionContext
 
-class EADController @Inject()(authenticate: AuthAction, eadService: EADService)(implicit mcc: MessagesControllerComponents, ec: ExecutionContext)
+class EADController @Inject()(authenticate: AuthAction, mcc: MessagesControllerComponents, eadService: EADService)(implicit ec: ExecutionContext)
```
PatrykRudnicki: 👍

conf/application.conf
```diff
@@ -78,15 +78,7 @@ microservice {
...
-    }
-
-    customs-declarations-information {
-      host = localhost
-      port = 9834
-      fetch-mrn-status = "/mrn/ID/status"
-      api-version = "1.0"
-      bearer-token = "Bearer customs-declare-exports-frontend"
-      client-id = "customs-declare-exports-frontend"
+      fetch-ead = "/ead"
```
PatrykRudnicki: Probably just naming, because we're not fetching the EAD but the dec information. Of course it's just for EAD, so I think we can stay with this

PatrykRudnicki: But after checking backend, probably it's okay to stay with this naming

PauloPortugal: Let's continue with this and then discuss once we break the microservice

PR #778, opened 2020-03-16
CEDS-2231 Improvements to Goods Location page

PatrykRudnicki:


app/forms/declaration/GoodsLocation.scala
```diff
@@ -19,55 +19,53 @@ package forms.declaration
...
-  postCode: Option[String],
-  country: String
-)
+case class GoodsLocation(code: String)
 
 object GoodsLocation extends DeclarationPage {
-  implicit val format = Json.format[GoodsLocation]
 
-  val formId = "GoodsLocation"
+  implicit val format: OFormat[GoodsLocation] = Json.format[GoodsLocation]
```
SteveSugden: Is this all taken from the movements service?

app/forms/declaration/GoodsLocation.scala
```diff
@@ -19,55 +19,53 @@ package forms.declaration
...
-  typeOfLocation: String,
-  qualifierOfIdentification: String,
-  identificationOfLocation: Option[String],
-  additionalIdentifier: Option[String],
-  addressLine: Option[String],
-  city: Option[String],
-  postCode: Option[String],
-  country: String
-)
+case class GoodsLocation(code: String)
```
SteveSugden: Would making the 'code' optional make the data migration easier?  In fact, you might not have to migrate the data at all.  Unless you are planning on synthesising the code based on the contents of the existing fields.

app/forms/declaration/GoodsLocation.scala
PatrykRudnicki: Yes

app/forms/declaration/GoodsLocation.scala
PatrykRudnicki: Unfortunately not, because previously we had a lot of fields, I'll try to achieve it in easy way, but if not happy to discuss other approaches

PatrykRudnicki: Based on the xml fields I stay with the backend model, so just changed the form model

app/forms/declaration/GoodsLocation.scala
```diff
@@ -17,58 +17,76 @@
...
+        isEmpty or (
+          validateCountry and validateLocationType and validateQualifierCode and noShorterThan(10) and noLongerThan(17)
+        )
+      )
+  )(GoodsLocationForm.apply)(GoodsLocationForm.unapply)
+
+  def form(): Form[GoodsLocationForm] = Form(mapping)
+}
+
+case class GoodsLocation(country: String, typeOfLocation: String, qualifierOfIdentification: String, identificationOfLocation: String) {
```
bambuchaAdm: Maybe this class should leave in declaration exchange

PatrykRudnicki: Good point. I'll move it there

PatrykRudnicki: Moved

app/forms/declaration/GoodsLocationForm.scala
```diff
@@ -0,0 +1,81 @@
...
+    input.drop(3).headOption.map(_.toString).exists(predicate)
+  }
+
+  val mapping = Forms.mapping(
+    "code" -> text()
+      .verifying("declaration.goodsLocation.code.empty", nonEmpty)
+      .verifying(
+        "declaration.goodsLocation.code.error",
+        isEmpty or (
+          validateCountry and validateLocationType and validateQualifierCode and noShorterThan(10) and noLongerThan(17)
```
SteveSugden: Should the "shorter than" test come first?  If the string is too short does the take(2) (for example) not throw an out of bounds?

app/views/declaration/goods_location.scala.html
```diff
@@ -16,92 +16,33 @@
...
-        @components.back_link(Navigator.backLink(GoodsLocation, mode)(request))
+        @components.back_link(Navigator.backLink(GoodsLocationForm, mode)(request))
 
         @components.error_summary(form.errors)
 
         @components.heading(
             sectionHeader = messages("declaration.summary.locations.header"),
-            title = messages("supplementary.goodsLocation.title"),
-            hint = messages("supplementary.goodsLocation.hint")
+            title = messages("declaration.goodsLocation.title"),
```
SteveSugden: 👍

app/forms/declaration/GoodsLocationForm.scala
PatrykRudnicki: It doesn't matter, because methods `take` and `drop` doesn't throw an exception but return the empty string which will still fail the validation. The same validation rules are on the movements

PR #782, opened 2020-03-18
CEDS-2191 - Change validation on Commodity Measurements page

SteveSugden:
For clearance requests
- remove supplementary units question
- make net and gross weight questions optional
- remove supplementary units question from summary page


app/views/declaration/summary/commodity_measure_section.scala.html
```diff
@@ -0,0 +1,47 @@
...
+@import models.DeclarationType._
+@import models.declaration.ExportItem
+@import forms.declaration.CommodityMeasure
+@import models.requests.JourneyRequest
+@import models.viewmodels.HtmlTableRow
+@import models.Mode
+
+@(mode: Mode, item: ExportItem, commodityMeasure: CommodityMeasure)(implicit messages: Messages, journeyRequest: JourneyRequest[_])
+
+@if(!journeyRequest.isType(DeclarationType.CLEARANCE)) {
```
PatrykRudnicki: I know it will be longer but probably it can be more readable without negation. So specify the list off allowed journeys instead of not allowed.

docs/Customs Declare Exports AutoComplete.js
```diff
@@ -1,7 +1,7 @@
 // ==UserScript==
 // @name         Customs Declare Exports AutoComplete
 // @namespace    http://tampermonkey.net/
-// @version      1.27
+// @version      1.28
```
PatrykRudnicki: Just realised that I didn't change the version in my PR

app/views/declaration/summary/commodity_measure_section.scala.html
SteveSugden: Won't that be less maintainable if/when we add another dec type?   I could create a `isNotType` method on JourneyRequest though.

PR #787, opened 2020-03-19
CEDS-2246 Make Declarant details optional for Clearance request

PatrykRudnicki:


app/controllers/declaration/DeclarantDetailsController.scala
```diff
@@ -43,14 +43,14 @@ class DeclarantDetailsController @Inject()(
 
   def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
     request.cacheModel.parties.declarantDetails match {
-      case Some(data) => Ok(declarantDetailsPage(mode, DeclarantDetails.form().fill(data)))
-      case _          => Ok(declarantDetailsPage(mode, DeclarantDetails.form()))
+      case Some(data) => Ok(declarantDetailsPage(mode, DeclarantDetails.form(request.declarationType).fill(data)))
```
SteveSugden: I've created a private method in my controllers to be able to continue construct the form(). E.g.

```private def form()(implicit request: JourneyRequest[_]) = DepartureTransport.form(request.declarationType)```

File: app/forms/common/Eori.scala [hunk 2]
```diff
@@ -39,6 +39,6 @@ object Eori {
   def mapping(errorPrefix: String): Mapping[Eori] =
     text()
       .verifying(s"${errorPrefix}.eori.empty", nonEmpty)
-      .verifying(s"${errorPrefix}.eori.error.format", isValidEORIPattern and noLongerThan(17) and noShorterThan(3))
+      .verifying(s"${errorPrefix}.eori.error.format", isEmpty or (isValidEORIPattern and noLongerThan(17) and noShorterThan(3)))
```
SteveSugden: Is this safe doing this on a "common" component?

app/views/declaration/declarant_details.scala.html
```diff
@@ -36,7 +36,7 @@
 @(mode: Mode, form: Form[DeclarantDetails])(implicit request: JourneyRequest[_], messages: Messages)
 
 @govukLayout(
-    title = Title("supplementary.declarant.title"),
+    title = Title("declaration.declarant.title"),
```
SteveSugden: :+1 for name change

test/forms/declaration/DeclarantDetailsSpec.scala
```diff
@@ -16,38 +16,54 @@
...
+import models.DeclarationType._
 import org.scalatest.{MustMatchers, WordSpec}
 import play.api.libs.json.{JsObject, JsValue}
 
 class DeclarantDetailsSpec extends WordSpec with MustMatchers {
 
   import DeclarantDetailsSpec._
 
-  "DeclarantDetails mapping used for binding data" should {
+  Seq(SUPPLEMENTARY, STANDARD, OCCASIONAL, SIMPLIFIED).map { decType =>
```
SteveSugden: New test for CLEARANCE, empty form, no errors?

app/forms/common/Eori.scala
PatrykRudnicki: This component was created, but had wrong validation, because fo empty show two errors.

I don't like this approach, I like wrapping Strings in class to be more descriptive and type safe, but not sure if there is any value in this Eori class.

test/forms/declaration/DeclarantDetailsSpec.scala
PatrykRudnicki: Good point! Thanks. I'll add this

app/controllers/declaration/DeclarantDetailsController.scala
PatrykRudnicki: Fixed. Thanks!

test/forms/declaration/DeclarantDetailsSpec.scala
PatrykRudnicki: Added

PatrykRudnicki:
@SteveSugden Could you look once again?

PR #788, opened 2020-03-19
CEDS-2189 Make CommodityDetails.descriptionOfGoods optional

PauloPortugal:
For Clearance Requests, data elements '6/14 Commodity code' and
'6/8 Description of Goods' are not required in the majority of cases.

This data element is not required on clearance requests using procedure
codes 00 18 or 00 19 in D.E. 1/10.

**Depends on: https://github.com/hmrc/customs-declare-exports/pull/169**

### DIFF
--- a/app/forms/declaration/CommodityDetails.scala
+++ b/app/forms/declaration/CommodityDetails.scala
@@ -23,7 +23,7 @@ import play.api.data.{Form, Mapping}
 import play.api.libs.json.Json
 import utils.validators.forms.FieldValidator._
 
-case class CommodityDetails(combinedNomenclatureCode: Option[String], descriptionOfGoods: String)
+case class CommodityDetails(combinedNomenclatureCode: Option[String], descriptionOfGoods: Option[String])
 
 object CommodityDetails extends DeclarationPage {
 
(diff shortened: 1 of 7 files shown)


app/forms/declaration/CommodityDetails.scala
```diff
@@ -23,7 +23,7 @@ import play.api.data.{Form, Mapping}
 import play.api.libs.json.Json
 import utils.validators.forms.FieldValidator._
 
-case class CommodityDetails(combinedNomenclatureCode: Option[String], descriptionOfGoods: String)
+case class CommodityDetails(combinedNomenclatureCode: Option[String], descriptionOfGoods: Option[String])
```
SteveSugden: I think I prefer this way of creating optional & required mappings.  The ones I have done so far use different apply/unapply methods (idea borrowed from `TotalPackageQuantity`)

app/views/declaration/summary/item_section.scala.html
```diff
@@ -14,7 +14,10 @@
  * limitations under the License.
  *@
 
-@import models.DeclarationType._
+@import views.html.declaration.summary.onward_supply_relief
+@import views.html.declaration.summary.package_information
+@import views.html.declaration.summary.union_and_national_codes
```
SteveSugden: Why all the new imports?

PauloPortugal: Twirl templates without explicit view imports within IntelliJ Ultimate are flagged as errors.
If we explicitly import them this goes away.
I know that this is meaningless

SteveSugden: That is a useful tip!

PauloPortugal:
Smoke tests have passed as well

PR #789, opened 2020-03-20
CEDS-2190 package information validation clearance journey

MaciejRewera:
This PR makes all fields on package-information optional for Clearance journey.
No empty elements on this page are accepted, but the user can proceed without adding any package information.

Additionally, validation for Number of packages has been fixed to accept values from range 0 - 99999 inclusive.


app/controllers/declaration/PackageInformationController.scala
```diff
@@ -77,22 +77,24 @@ class PackageInformationController @Inject()(
...
+
   private def updateExportsCache(itemId: String, updatedCache: Seq[PackageInformation])(
     implicit r: JourneyRequest[AnyContent]
   ): Future[Option[ExportsDeclaration]] =
     updateExportsDeclarationSyncDirect(model => model.updatedItem(itemId, _.copy(packageInformation = Some(updatedCache.toList))))
 
   private def nextPage(itemId: String, declarationType: DeclarationType): Mode => Call =
     declarationType match {
-      case DeclarationType.SUPPLEMENTARY | DeclarationType.STANDARD | DeclarationType.CLEARANCE =>
+      case DeclarationType.SUPPLEMENTARY | DeclarationType.STANDARD | CLEARANCE =>
```
PatrykRudnicki: Why you have `DeclarationType` everywhere but you removed from CLEARANCE?

app/forms/declaration/PackageInformation.scala
```diff
@@ -17,50 +17,84 @@
...
 object PackageInformation extends DeclarationPage {
 
-  def fromJsonString(value: String): Option[PackageInformation] = Json.fromJson(Json.parse(value)).asOpt
+  private def fromValues(typesOfPackages: String, numberOfPackages: Int, shippingMarks: String): PackageInformation =
+    PackageInformation(typesOfPackages = Some(typesOfPackages), numberOfPackages = Some(numberOfPackages), shippingMarks = Some(shippingMarks))
+
+  private def toValues(packageInformation: PackageInformation): Option[(String, Int, String)] =
+    for {
+      typesOfPackages <- packageInformation.typesOfPackages
+      numberOfPackages <- packageInformation.numberOfPackages
```
PatrykRudnicki: If typesOfPackages will be empty you skip rest of the values

app/forms/declaration/PackageInformation.scala
```diff
@@ -17,50 +17,84 @@
...
+  val mappingAllFieldsMandatory = Forms
     .mapping(
       "typesOfPackages" ->
         text()
           .verifying("supplementary.packageInformation.typesOfPackages.empty", nonEmpty)
           .verifying("supplementary.packageInformation.typesOfPackages.error", isEmpty or isContainedIn(PackageTypes.all.map(_.code))),
       "numberOfPackages" ->
         number()
-          .verifying("supplementary.packageInformation.numberOfPackages.error", q => q > 0 && q <= 999999),
+          .verifying("supplementary.packageInformation.numberOfPackages.error", isInRange(NumberOfPackagesLimitLower, NumberOfPackagesLimitUpper)),
```
PatrykRudnicki: Personally I'm not feeling that moving this range to the variables make sense. For me it's always more complicated. Preferable `isInRange(0, 99999)`

app/forms/declaration/PackageInformation.scala
```diff
@@ -17,50 +17,84 @@
...
+      )
     )(PackageInformation.apply)(PackageInformation.unapply)
+    .verifying("supplementary.packageInformation.empty", validatePackageInformation(_))
 
-  val DUPLICATE_MSG_KEY = "supplementary.packageInformation.global.duplicate"
-  val LIMIT_MSG_KEY = "supplementary.packageInformation.global.limit"
-  val USE_ADD = "supplementary.packageInformation.global.useAdd"
-  val ADD_ONE = "supplementary.packageInformation.global.addOne"
+  private def validatePackageInformation(packageInformation: PackageInformation): Boolean =
+    !(packageInformation.typesOfPackages.isEmpty && packageInformation.numberOfPackages.isEmpty && packageInformation.shippingMarks.isEmpty)
```
PatrykRudnicki: I think it can be useful to have method `isEmpty` inside the PackageInformation class and just invoke `_.isEmpty`

app/controllers/declaration/PackageInformationController.scala
MaciejRewera: This looks like a miss-match with imports. I'll fix it.

app/forms/declaration/PackageInformation.scala
MaciejRewera: I think contrary, so if someone alse would like to share their opinion, we could decide then.

app/forms/declaration/PackageInformation.scala
MaciejRewera: I will return `None`. To be fair, I am not sure how it should behave in that case:
1. Return `None`
2. Return some default values for the fields and `None` only if all are empty

PatrykRudnicki: But for comprehensive it's combination of flatMaps and maps

```
packageInformation.typesOfPackages.flatMap { typesOfPackages: String =>
  packageInformation.numberOfPackages.flatMap { numberOfPackages: Int =>
    packageInformation.shippingMarks.map { shippingMarks: String =>
      (typesOfPackages, numberOfPackages, shippingMarks)
    }
  }
}
```

This is the equivalent of above code, in case that one of the values will be empty e.g. `typesOfPackages` method `toValues` will return `None`. Is it expected behaviour? Because I'm not sure what you want to achieve with this method when the `typeOfPackages` is empty.

In case if the above scenario is correct this method is fine, but I want to confirm this. I'm not sure if you can have a `numberOfPackages` as defined and rest not defined.

In general it can be my wrong understanding what this method should do.

PatrykRudnicki: I see where are you using it. Can you test scenario when you're doing mandatory journey, where those fields are mandatory and you'll fill just the last input?

I'm not sure about behaviour too

PatrykRudnicki: Hmm, it works good. Sorry for the confusion. I will discard my request changes to this. Still I want to understand what should be the behaviour of this method and when it will be invoked

app/forms/declaration/PackageInformation.scala
```diff
@@ -91,7 +94,7 @@ object PackageInformation extends DeclarationPage {
     .verifying("supplementary.packageInformation.empty", validatePackageInformation(_))
 
   private def validatePackageInformation(packageInformation: PackageInformation): Boolean =
-    !(packageInformation.typesOfPackages.isEmpty && packageInformation.numberOfPackages.isEmpty && packageInformation.shippingMarks.isEmpty)
+    !(packageInformation.isEmpty)
```
PatrykRudnicki: I think this is not required to be in the method.

Brackets around `packageInformation.isEmpty` are not necessary

MaciejRewera: They're not, but they make the `!` more visible

MaciejRewera: Ahh.... now I understand. Let me amend this little change.

app/controllers/declaration/PackageInformationController.scala
```diff
@@ -77,24 +77,26 @@ class PackageInformationController @Inject()(
...
-      .saveAndContinue(boundForm, cachedData, isMandatory = true, PackageInformation.limit)
+      .saveAndContinue(boundForm, cachedData, isMandatory(request), PackageInformation.limit)
       .fold(
         formWithErrors => Future.successful(BadRequest(packageInformationPage(mode, itemId, formWithErrors, cachedData))),
         updatedCache =>
           updateExportsCache(itemId, updatedCache)
             .map(_ => navigator.continueTo(mode, nextPage(itemId, request.declarationType)))
       )
 
+  private def isMandatory(journeyRequest: JourneyRequest[_]): Boolean = journeyRequest.declarationType != CLEARANCE
```
SteveSugden: You could make the request implicit here and not have to pass it in

app/controllers/declaration/PackageInformationController.scala
```diff
@@ -49,12 +49,12 @@ class PackageInformationController @Inject()(
 
   def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
     val items = request.cacheModel.itemBy(itemId).flatMap(_.packageInformation).getOrElse(List.empty)
-    Ok(packageInformationPage(mode, itemId, form(), items))
+    Ok(packageInformationPage(mode, itemId, form(request.declarationType), items))
```
SteveSugden: You could have a private method here to construct the form (and below). e.g.
```private def form()(implicit request: JourneyRequest[_]) = PackageInformation.form(request.declarationType)```

File: app/forms/declaration/PackageInformation.scala [hunk 8]
```diff
@@ -17,50 +17,86 @@
...
+
+  def isEmpty: Boolean = typesOfPackages.isEmpty && numberOfPackages.isEmpty && shippingMarks.isEmpty
+
+  def nonEmpty: Boolean = !isEmpty
+
   def toJson: JsValue = Json.toJson(this)(PackageInformation.format)
 
-  def typesOfPackagesText: String = PackageTypes.findByCode(typesOfPackages).asText()
+  def typesOfPackagesText: Option[String] =
+    for {
```
SteveSugden: Nice.  But could you not just map the `typesOfPackages`?

app/forms/declaration/PackageInformation.scala
SteveSugden: Can't this simply be
` = Some(packageInformation.type, packageInformation.number, packageInformation.shipping)`
Or am I missing something

app/forms/declaration/PackageInformation.scala
SteveSugden: `isInRange(0, 99999)` is easier to parse as a reader.  Guess it depends on whether those values are used elsewhere.     The capitalisation of the constants looks wrong somehow though.

app/forms/declaration/PackageInformation.scala
```diff
@@ -17,50 +17,86 @@
...
-  val ADD_ONE = "supplementary.packageInformation.global.addOne"
+  val mappingAllFieldsOptional = Forms
+    .mapping(
+      "typesOfPackages" -> optional(
+        text()
+          .verifying("supplementary.packageInformation.typesOfPackages.error", isContainedIn(PackageTypes.all.map(_.code)))
+      ),
+      "numberOfPackages" -> optional(
+        number()
+          .verifying("supplementary.packageInformation.numberOfPackages.error", isInRange(NumberOfPackagesLimitLower, NumberOfPackagesLimitUpper))
```
SteveSugden: They are used again!

app/services/PackageTypes.scala
```diff
@@ -33,5 +33,5 @@ object PackageTypes {
 
   val packageTypeCodeMap: Map[String, PackageType] = all.map(packageType => (packageType.code, packageType)).toMap
 
-  def findByCode(code: String): PackageType = packageTypeCodeMap(code)
+  def findByCode(code: String): Option[PackageType] = packageTypeCodeMap.get(code)
```
SteveSugden: Why does this return Option now?   Do we need to support unknown codes?

app/forms/declaration/PackageInformation.scala
MaciejRewera: This is driven by scala-style guideline: https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods
Specifically: "Constant names should be in upper camel case. Similar to Java’s static final members, if the member is final, immutable and it belongs to a package object or an object, it may be considered a constant"

MaciejRewera: And as you've noticed, these values are used in 2 places.

SteveSugden: OK.  But when U was a lad and did Java, constants were ALL_CAPS.  You kids with your fancy new ideas!

app/forms/declaration/PackageInformation.scala
MaciejRewera: Unfortunatelly, it can't.
All fields in `PackageInformation` are `Option`s, so I could either provide some default values (which makes problem for `Int`, because '0' is a valid answer), or I could return `None` in case the object is in invalid state.
The latter represents the "fail fast" approach.

SteveSugden: Sorry, I read that method signature wrong.  Thought it was `Option[(Option[String], Option[Int], Option[String])]`

app/controllers/declaration/PackageInformationController.scala
MaciejRewera: I did that initially, but then realized that it is easier for me to read it as a method when I have these parenthesis present.

app/forms/declaration/PackageInformation.scala
MaciejRewera: It could be
`typesOfPackages
    .flatMap(types => PackageTypes.findByCode(types)
    .map(_.asText()))`
But it is more readable with for comprehension in my opinion.

app/services/PackageTypes.scala
MaciejRewera: This stayed after some previous version. Thank you for spotting this.

test/unit/controllers/declaration/PackageInformationControllerSpec.scala
```diff
@@ -46,18 +47,17 @@ class PackageInformationControllerSpec extends ControllerSpec with ErrorHandlerM
...
+    withNewCaching(aDeclaration(withType(SUPPLEMENTARY)))
     setupErrorHandler()
   }
 
   "Package Information Controller" should {
 
     "return 200 (OK)" when {
-      "on Supplementary journey" when {
-
+      onEveryDeclarationJourney() { declaration =>
```
SteveSugden: :+1 for testing all journey types

PR #793, opened 2020-03-23
CEDS-2192 - 7/7 Identity of the means of transport at departure should be optional for Clearance Requests

SteveSugden:
This page (/departure-transport) requires two answers - select mode of transport (radio) and enter identifier (text).  

For clearance requests no answer is required.   The existing page design had the radio options and a single text field for identifier.   During pre-amigos it was decided to implement the new GDS design for this page whereby selecting a radio option would reveal a text input field for entering the identifier.

To allow the user (making a clearance request) to "de-select" an option, a "dummy" ("I don't know") option was added (which does not have an associated text edit field).

The validation (for non-clearance) is that a radio option and text value is required.
The validation for clearance is that either
- a radio for transport type is selected AND the text identifier is provided, or
- the 'I don't know' radio option is selected.

In order to preserve the users "I don't know" choice a dummy value has been created for TransportMode.   This value needs to be filtered in the customs_declare_exports service just before the creation of the XML payload to send to DMS (see BE PR)


app/models/declaration/Transport.scala
```diff
@@ -55,4 +55,6 @@ case class Transport(
 
 object Transport {
   implicit val format: Format[Transport] = Json.format[Transport]
+
+  val optionNone = "option_none"
```
MaciejRewera: Could this be moved to `TransportDetails`? To me it is more related to that scope. Unless there is something I don't understand...

SteveSugden: I thought it might be useful for if/when other modes of transport need a "none" option.  E.g. the ticket you have picked up.

MaciejRewera: My second thought is that since we are sharing this value between two services, maybe we should have a dedicated type? There is "Null object (anti) pattern" and I was thinking about using something similar here. Strings are very susceptible to type errors.

app/views/declaration/summary/transport_section.scala.html
```diff
@@ -29,15 +29,15 @@
...
       transport.meansOfTransportCrossingTheBorderIDNumber.isDefined ||
       transport.meansOfTransportCrossingTheBorderNationality.isDefined ||
       transport.meansOfTransportCrossingTheBorderType.isDefined
 }
 
 @messagesForDepartureMeansOfTransport(transport: Transport) = @{
     (transport.meansOfTransportOnDepartureType, transport.meansOfTransportOnDepartureIDNumber) match {
-        case (Some(meansType), Some(meansId)) => Seq(messages(s"declaration.summary.transport.departure.meansOfTransport.$meansType"), meansId)
+        case (Some(meansType), Some(meansId)) if meansId.nonEmpty => Seq(messages(s"declaration.summary.transport.departure.meansOfTransport.$meansType"), meansId)
+        case (Some(meansType), _) => Seq(messages(s"declaration.summary.transport.departure.meansOfTransport.$meansType"))
```
MaciejRewera: 👍

test/unit/controllers/declaration/DepartureTransportControllerSpec.scala
```diff
@@ -56,6 +61,14 @@ class DepartureTransportControllerSpec extends ControllerSpec with ErrorHandlerM
     case CLEARANCE     => routes.TransportContainerController.displayContainerSummary()
   }
 
+  private def formData(transportType: String, reference: String) =
+    JsObject(
+      Map(
+        DepartureTransport.meansOfTransportOnDepartureTypeKey -> JsString(transportType),
+        s"${DepartureTransport.meansOfTransportOnDepartureIDNumberKey}_$transportType" -> JsString(reference)
+      )
+    )
```
MaciejRewera: Nit: Another way of doing this (which is shorter in writing) is `Json.obj(...)`

test/views/declaration/DepartureTransportViewSpec.scala
```diff
@@ -91,83 +86,132 @@ class DepartureTransportViewSpec extends UnitViewSpec with CommonMessages with S
...
+          view.getElementById("Departure_NameOfInlandWaterwayVessel").attr("value") mustBe NameOfInlandWaterwayVessel
+          view
+            .getElementsByAttributeValue("for", "Departure_NameOfInlandWaterwayVessel")
             .text() mustBe "declaration.transportInformation.meansOfTransport.nameOfInlandWaterwayVessel"
         }
       }
 
-      "display 'Reference' section" which {
-        "have label" in {
+      "display 'Transport details type' id number section " which {
```
MaciejRewera: Could this be tested together with the radio section above? I mean each case with matching case above.

SteveSugden: I thought that would be testing too much at one time?

MaciejRewera: But radio option and its reveal component could be treated as a single "thing" ? You cannot have the revealing component with the radio option.

SteveSugden: Merged the two sets of tests together

SteveSugden:
I think I've addressed all your comments @MaciejRewera - could you take another look please?

test/unit/forms/declaration/DepartureTransportSpec.scala
```diff
@@ -50,94 +51,233 @@ class DepartureTransportSpec extends FormSpec {
...
+        val errorKeys = result.errors.map(_.key)
+        val errorMessages = result.errors.map(_.message)
+
+        errorKeys must be(List(idNumberField))
+        errorMessages must be(List("declaration.transportInformation.meansOfTransport.reference.error.empty"))
+      }
+
+      "fields are incorrect" in {
+
+        val incorrectForm = Map(transportTypeField -> "incorrect", idNumberField -> "correct")
```
MaciejRewera: The value for `idNumberField` is "correct" while test case says "fields are incorrect". Is it just because this field accepts any string except an empty one? And so one can't really make it incorrect?

test/unit/forms/declaration/DepartureTransportSpec.scala
```diff
@@ -50,94 +51,233 @@ class DepartureTransportSpec extends FormSpec {
...
+
+        val result = form.bind(incorrectForm)
+        val errorKeys = result.errors.map(_.key)
+        val errorMessages = result.errors.map(_.message)
+
+        errorKeys must be(List(transportTypeField))
+        errorMessages must be(List("declaration.transportInformation.meansOfTransport.departure.error.incorrect"))
+      }
+
+      "means of transport on departure id number is empty" in {
```
MaciejRewera: Isn't this test the same as "transport id not provided" ?
It looks like they're testing the same scenario but in a slightly different way.

test/unit/forms/declaration/DepartureTransportSpec.scala
MaciejRewera: Just noticed there is a test "means of transport on departure id number is too long", so I'd recommend giving field `idNumberField` some invalid value.

test/unit/forms/declaration/DepartureTransportSpec.scala
```diff
@@ -50,94 +51,233 @@ class DepartureTransportSpec extends FormSpec {
...
-          )
-        )
+        errorKeys must be(List(idNumberField))
+        errorMessages must be(List("declaration.transportInformation.meansOfTransport.reference.error.empty"))
       }
 
       "fields are incorrect" in {
 
-        val incorrectForm = Map("meansOfTransportOnDepartureType" -> "incorrect", "meansOfTransportOnDepartureIDNumber" -> "correct")
+        val incorrectForm = Map(transportTypeField -> "incorrect", idNumberField -> "correct")
```
MaciejRewera: Similar thing here with `idNumberField`

test/unit/forms/declaration/DepartureTransportSpec.scala
```diff
@@ -17,20 +17,21 @@
...
 
 import base.TestHelper
-import forms.declaration.DepartureTransport
+import forms.declaration.{DepartureTransport, TransportCodes}
 import forms.declaration.TransportCodes._
+import models.DeclarationType
+import models.declaration.Transport
 import unit.base.FormSpec
 
 class DepartureTransportSpec extends FormSpec {
```
MaciejRewera: In general, I'd suggest testing the common logic once (using `onJourney` method for e.g.) and then for Clearance journey only test additional cases.

test/views/declaration/summary/TransportSectionViewSpec.scala
```diff
@@ -63,6 +63,31 @@ class TransportSectionViewSpec extends UnitViewSpec with ExportsTestData {
...
+        Mode.Normal,
+        data.copy(transport = data.transport.copy(meansOfTransportOnDepartureType = None, meansOfTransportOnDepartureIDNumber = Some("")))
+      )(messages, journeyRequest())
+
+      view.getElementById("transport-reference-label").text() mustBe messages("declaration.summary.transport.departure.meansOfTransport.header")
+      view.getElementById("transport-reference").text() mustBe ""
+      view.getElementById("transport-reference-change") must haveHref(controllers.declaration.routes.DepartureTransportController.displayPage())
+    }
+
+    "display transport reference if question not applicable" in {
```
MaciejRewera: Nit: I think 'not applicable' is now named differently?

test/unit/forms/declaration/DepartureTransportSpec.scala
SteveSugden: The onJourney stuff is in the view specs, not the form.  I agree there are some duplicate tests here - I'll tidy them up.

test/unit/forms/declaration/DepartureTransportSpec.scala
SteveSugden: Rename test - its really testing invalid radio option value

test/unit/forms/declaration/DepartureTransportSpec.scala
SteveSugden: removed duplicate

SteveSugden:
Tidied up the tests @MaciejRewera

PR #796, opened 2020-03-24
CEDS-2248 ExporterDetails is optional for Clearance Requests

PauloPortugal:


PauloPortugal:
Smoke tests are green

PR #799, opened 2020-03-25
CEDS-2247 Implement Country class to represent Country and make Destination Country Optional for the Clearance request

PatrykRudnicki:


app/forms/declaration/countries/Countries.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  case object NextRoutingCountryPage extends CountryPage {
+    override val id = "routingCountry"
+  }
+
+  private def mapping(page: CountryPage, cachedCountries: Seq[Country] = Seq.empty): Mapping[String] =
+    text()
+      .verifying(s"declaration.${page.id}.empty", _.trim.nonEmpty)
+      .verifying(s"declaration.${page.id}.error", emptyOrValidCountry)
+      .verifying(s"declaration.routingCountries.duplication", !cachedCountries.flatMap(_.code).contains(_))
```
SteveSugden: I approve of this file very much.   Lots of good practices for optional/mandatary mapping and supporting multiple pages.   I like it.

Only comment is about that "duplication" check.  I can see what it's doing but the not combined with the flatmap jars a bit.  Only a nit though

test/utils/JourneyRequestHelper.scala
```diff
@@ -0,0 +1,68 @@
...
+
+package utils
+
+import models.DeclarationType
+import models.DeclarationType.DeclarationType
+import models.requests.JourneyRequest
+import org.scalatest.WordSpec
+import views.declaration.spec.UnitViewSpec
+
+trait JourneyRequestHelper extends WordSpec {
```
SteveSugden: Is this not what @MaciejRewera was talking about doing?

app/forms/declaration/countries/Countries.scala
PatrykRudnicki: This class is the old one, we had it before, just because of changing the name and more logic GitHub recognise it as new file.

We're storing the Country as a model, but the frontend autocomplete contains only country code.  This is the reason that I need map to the code

app/views/declaration/destinationCountries/origination_country.scala.html
```diff
@@ -46,7 +46,7 @@
         @helper.CSRF.formField
 
         @components.fields.field_autocomplete(
-            form("country"),
+            form("code"),
```
PauloPortugal: I would have been explicit and name the variable as `countryCode`, but that looks fine as well

test/utils/JourneyRequestHelper.scala
PauloPortugal: I think we have a similar one for the UnitViewSpec

app/views/declaration/destinationCountries/origination_country.scala.html
PatrykRudnicki: I thought about the same, but I prefer to stay with it right now, to not block Maciej.

I'll prepare PR with this change today.

test/utils/JourneyRequestHelper.scala
PatrykRudnicki: I extracted this from UnitViewSpec, because it's useful for models changes too.

PR #805, opened 2020-03-27
CEDS-2228 - Move Section 4 of the Standard Declaration journey to GDS

SteveSugden:
Pages covered
- What was the nature of the transaction?
- Onward Supply Relief (Y/N page only)
- Commodity Details
- Does the item have a UN Dangerous Goods Code?
- Do these goods have a CUS code?

### DIFF
--- a/app/views/components/gds/exportsInputTextArea.scala.html
+++ b/app/views/components/gds/exportsInputTextArea.scala.html
@@ -1,0 +1,45 @@
+@*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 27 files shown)


app/views/components/gds/exportsInputTextArea.scala.html
```diff
@@ -0,0 +1,45 @@
...
+        labelKey: String,
+        hintKey: Option[String] = None,
+        isPageHeading: Boolean = false,
+        headingClasses: String = gdsPageLabel,
+        labelClasses: String = "govuk-label--m"
+)(implicit messages: Messages)
+
+@buildLabel = @{
+    if(isPageHeading) {
+        Label(content = Text(messages(labelKey)), isPageHeading = true, classes = headingClasses)
```
bambuchaAdm: can we pass here just pageHeading, and remove if ?

SteveSugden: The classes are different as well.   Copied this from almost identical `exportsInputText` and didn't look too closely at the logic I'm afraid.

PR #806, opened 2020-03-27
CEDS-2290 - add 'value' param to divider so it gets filtered when not required

SteveSugden:


SteveSugden:
Thanks.  
I've just re-enstated the filtering of the divider ("Or") based on the presence of "radio_divider" in the list-of-available-declarations in the config file.  So that it works as before my change to GDS.
And there's a test for that functionality now.

PR #807, opened 2020-03-27
CEDS-2226

alsahid1993:
Updated FE elements within
representative_details.scala
goods_location.scala

Jira: https://jira.tools.tax.service.gov.uk/browse/CEDS-2226

### DIFF
--- a/app/views/declaration/declaration_choice.scala.html
+++ b/app/views/declaration/declaration_choice.scala.html
@@ -22,7 +22,6 @@
 @import views.html.components.gds.{errorSummary, gdsMainTemplate, _}
 @import views.components.gds.Styles._
 @import views.{BackButton, Title}
-@import views.components.inputs.RadioOption
 
 @this(
     govukLayout: gdsMainTemplate,
@@ -33,6 +32,7 @@
     formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF,
     appConfig: AppConfig
 )
+
 
 @(mode: Mode, form: Form[_])(implicit request: Request[_], messages: Messages)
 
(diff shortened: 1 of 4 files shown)


app/views/declaration/declaration_choice.scala.html
```diff
@@ -91,16 +91,15 @@
                     checked = form("type").value.contains(DeclarationType.OCCASIONAL.toString)
                 ),
                 RadioItem(
-                    divider = Some(messages("declaration.type.or")),
-                    value = Some(RadioOption.dividerId)
+                    divider = Some(messages("declaration.type.or"))
```
PatrykRudnicki: This change and line 102 has been changed by @SteveSugden  last time. Not sure if this change with removing value is correct

app/views/declaration/goods_location.scala.html
```diff
@@ -54,7 +54,7 @@
         @pageTitle(messages("declaration.goodsLocation.title"))
 
         @exportsInputText(
-        field = form("code"),
+        field = form("dev-goods-location-code"),
```
PatrykRudnicki: why this change?

PatrykRudnicki: This won't work, because this filed must be the same like form field

app/views/declaration/declaration_choice.scala.html
SteveSugden: No.  This looks like it's reverting my changes of this morning.

test/views/declaration/DeclarationChoiceViewSpec.scala
```diff
@@ -17,26 +17,18 @@
 package views.declaration
 
 import base.Injector
```
SteveSugden: I don't think there should be any changes to this test either.  Declaration Choice page as already been GDS - its not part of this ticket

app/views/declaration/goods_location.scala.html
alsahid1993: I was trying to make the the following element consistent with the prototype's:
<input class="govuk-input" id="dev-local-reference-number" name="dev-local-reference-number" type="text" aria-describedby="dev-local-reference-number-hint">

Anu informed me that the tests were failing when the previous elements were listed as:
<input class="govuk-input govuk-!-width-two-thirds" id="code" name="code" type="text" aria-describedby="code-hint">

PR #809, opened 2020-03-27
CEDS-2226 Updated Heading & Elements

alsahid1993:
Updated representative_details.scala.html - Page contains only one h1 element
Updated goods_location.scala.html - title elements
Jira ticket: https://jira.tools.tax.service.gov.uk/browse/CEDS-2226

### DIFF
--- a/app/views/declaration/goods_location.scala.html
+++ b/app/views/declaration/goods_location.scala.html
@@ -51,14 +51,13 @@
         @errorSummary(form.errors)
 
         @sectionHeader(messages("declaration.summary.locations.header"))
-        @pageTitle(messages("declaration.goodsLocation.title"))
 
         @exportsInputText(
         field = form("code"),
-        labelKey = "",
+        labelKey = "declaration.goodsLocation.title",
         hintKey = Some("declaration.goodsLocation.hint"),
         isPageHeading = true,
-        headingClasses = "govuk-label--m"
+        headingClasses = "govuk-label--xl"
 )
 
 
--- a/app/views/declaration/goods_location.scala.html
+++ b/app/views/declaration/goods_location.scala.html
@@ -23,6 +23,7 @@
 @import views.components.gds.Styles._
 @import views.html.components.gds._
 @import views.{BackButton, Title}
+@import views.components.gds.Styles._
 
 @this(
     govukLayout: gdsMainTemplate,
@@ -57,7 +58,7 @@
         labelKey = "declaration.goodsLocation.title",
         hintKey = Some("declaration.goodsLocation.hint"),
         isPageHeading = true,
-        headingClasses = "govuk-label--xl"
+        headingClasses = gdsPageLabel
 )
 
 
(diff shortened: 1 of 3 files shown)


app/views/declaration/goods_location.scala.html
```diff
@@ -51,14 +51,13 @@
...
-        @pageTitle(messages("declaration.goodsLocation.title"))
 
         @exportsInputText(
         field = form("code"),
-        labelKey = "",
+        labelKey = "declaration.goodsLocation.title",
         hintKey = Some("declaration.goodsLocation.hint"),
         isPageHeading = true,
-        headingClasses = "govuk-label--m"
+        headingClasses = "govuk-label--xl"
```
SteveSugden: Can you use the constant `gdsPageLabel` here instead of "govuk-label--xl". (You'll need an extra import - 

@import views.components.gds.Styles._

That "xl" style you have used is the right one according to the prototype but it's larger than the other non-GDS pages.   By using the constant value consistently everywhere, we can hopefully change to "xl" once all the pages have been converted.

app/views/declaration/representative_details.scala.html
```diff
@@ -92,7 +92,7 @@
             field = form("details.eori"),
             labelKey = "supplementary.representative.eori.legend",
             hintKey = Some("supplementary.representative.eori.hint"),
-            isPageHeading = true,
+            isPageHeading = false,
```
SteveSugden: Good spot.  You can get rid of the next line (headingClass) as well

app/views/declaration/goods_location.scala.html
alsahid1993: noted

test/views/declaration/LocationViewSpec.scala
```diff
@@ -72,7 +72,7 @@ class LocationViewSpec extends UnitViewSpec with ExportsTestData with Stubs with
 
     "display header" in {
 
-      view.getElementById("title").text() mustBe "declaration.goodsLocation.title"
+      view.getElementsByClass("govuk-label govuk-label--l").text() mustBe "declaration.goodsLocation.title"
     }
```
SteveSugden: I normally do this to select the title
`view.getElementsByTag("h1").text() mustBe "blah.blah"`
There might be more than one element with those style classes on the page.  (Not in your case because the test would fail as .text() would return all values for all element found (i.e. more than just the title).

So you don't have to change if you don't want to.

app/views/declaration/representative_details.scala.html
```diff
@@ -92,8 +92,8 @@
             field = form("details.eori"),
             labelKey = "supplementary.representative.eori.legend",
             hintKey = Some("supplementary.representative.eori.hint"),
-            isPageHeading = true,
-            headingClasses = "govuk-label--m"
+            isPageHeading = false,
+            headingClasses = ""
```
SteveSugden: I meant you can get rid of the whole "headingClass" line.   The value isn't used if isPageHeading is false.

PR #810, opened 2020-03-27
CEDS-2238 - statistical_value page to GDS

SteveSugden:


app/assets/stylesheets/customsdecexfrontend-app.scss
```diff
@@ -14,7 +14,10 @@
...
 
+
 $govuk-assets-path: '/customs-declare-exports/assets/';
 $govuk-images-path: "/customs-declare-exports/assets/lib/govuk-frontend/govuk/assets/images/";
 $govuk-fonts-path: "/customs-declare-exports/assets/lib/govuk-frontend/govuk/assets/fonts/";
-@import 'lib/govuk-frontend/govuk/all';
\ No newline at end of file
+@import 'lib/govuk-frontend/govuk/all';
+
+@import 'partials/_currency_input.scss';
```
MaciejRewera: Is it required to be imported after `govuk/all` ? If not, then it mightt be better to keep it before. GDS had some rules to be the last imported part.

app/assets/stylesheets/partials/_currency_input.scss
```diff
@@ -0,0 +1,9 @@
+
+span.prefix {
+  position: absolute;
+  padding: 8px;
+  margin-left: 2px;
+}
+input.input-prefix {
+  padding-left: 30px;
+}
```
PatrykRudnicki: 👍 for new file for the styles 🎉

app/assets/stylesheets/partials/_currency_input.scss
```diff
@@ -0,0 +1,9 @@
+
+span.prefix {
```
bambuchaAdm: please put this into separate scope like `.input-currency` - would not apply accidentally

SteveSugden:
I'm going to check with Chris about over-riding the govuk styles.

app/assets/stylesheets/partials/_currency_input.scss
SteveSugden: You mean change `span.prefix` to `span.input-currency`?   In fact, on reflection, I don't know why I didn't make the style names the same.   Any they aren't really limited to currency's - could be any prefix.

I'' have another look at the class names.

app/assets/stylesheets/customsdecexfrontend-app.scss
SteveSugden: Moved it to above govuk.  Turns out it didn't need to be last to work.

app/assets/stylesheets/partials/_currency_input.scss
SteveSugden: Renamed the new css classes to be more specific to the role they are being used for.

app/assets/stylesheets/partials/_currency_input.scss
```diff
@@ -0,0 +1,9 @@
+
+span.currency-prefix {
```
bambuchaAdm: in scss you could write this as 
```
.currency-prefix {
  span {
    // code
  }
  input {
    // code
  }
}
```

SteveSugden: Doesn't seem to work for some reason.  I'll leave as-is if that's ok

PR #815, opened 2020-04-03
CEDS-2237 fix

iservport:
Following post amigos discussions, error messages need to be updated.  Moving the ticket back to in-development

1) Remove button additional space, OK
2) Error message without selecting an option, OK
3) No option selected and saved, next PR
4) Add Error message options, next PR
5) Wordings on the summary page, next PR
6) EORI Error Message, next PR


app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -17,23 +17,28 @@
...
 case class DeclarationAdditionalActors(eori: Option[Eori], partyType: Option[String]) {
 
+  import DeclarationAdditionalActors._
+
   def isDefined: Boolean = eori.isDefined && partyType.isDefined
 
-  def toJson: JsValue = Json.toJson(this)(DeclarationAdditionalActors.format)
+  def isAllowed: Boolean = partyType.exists(allowedPartyTypes.contains)
+
+  def toJson: JsValue = Json.toJson(this)(format)
```
PatrykRudnicki, line 35: Where is it used?

iservport: I consolidated the allowed types in the DeclarationAdditionalActors companion object and added this method to help with validation.

app/models/declaration/DeclarationAdditionalActorsData.scala
```diff
@@ -20,13 +20,14 @@ import forms.declaration.DeclarationAdditionalActors
 import play.api.libs.json.Json
 
 case class DeclarationAdditionalActorsData(actors: Seq[DeclarationAdditionalActors]) {
-  def containsItem(actor: DeclarationAdditionalActors): Boolean = actors.contains(actor)
+  def addActor(actor: DeclarationAdditionalActors): DeclarationAdditionalActorsData =
+    copy(if (actor.isAllowed) actors :+ actor else actors)
```
MaciejRewera, line 24: I would recommend not to mix-in the validation here.
There is no information returned on why the operation did not succeed. I'd rather split that into two separate methods and use them together if needed (similarly to `Iterator` pattern).

app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -72,6 +79,13 @@ object DeclarationAdditionalActors extends DeclarationPage {
 
   def form(): Form[DeclarationAdditionalActors] = Form(mapping)
 
+  def actorsValidator(actor: DeclarationAdditionalActors, actors: Seq[DeclarationAdditionalActors]): Option[Seq[(String, String)]] =
```
MaciejRewera, line 82: Is there a specific reason/situation for this method to be here? At first glance it is more apropriate to be part of `DeclarationAdditionalActorsData` class.

iservport: That's a good idea.

app/controllers/declaration/DeclarationAdditionalActorsController.scala
```diff
@@ -82,22 +78,19 @@ class DeclarationAdditionalActorsController @Inject()(
...
+      updateCache(updatedCache)
+        .map(_ => navigator.continueTo(mode, routes.DeclarationAdditionalActorsController.displayPage))
     }
 
+  private def actorsValidator(mode: Mode, actor: DeclarationAdditionalActors, actors: Seq[DeclarationAdditionalActors])(
+    success: Future[Result]
+  )(implicit request: JourneyRequest[AnyContent], hc: HeaderCarrier): Future[Result] =
+    DeclarationAdditionalActors
+      .actorsValidator(actor, actors)
+      .fold[Future[Result]](success)(errors => handleErrorPage(mode, errors, actor, actors))
```
MaciejRewera, line 92: 👍 Nice!
I could only suggest a different naming for parameters: actors -> existingActors (?), success -> onSuccess

app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -43,16 +48,18 @@ object DeclarationAdditionalActors extends DeclarationPage {
...
 
-  val mapping = Forms.mapping(
+  val mapping: Mapping[DeclarationAdditionalActors] = Forms.mapping(
     "eoriCS" -> eoriMappingFor(PartyType.Consolidator),
     "eoriMF" -> eoriMappingFor(PartyType.Manufacturer),
     "eoriFW" -> eoriMappingFor(PartyType.FreightForwarder),
     "eoriWH" -> eoriMappingFor(PartyType.WarehouseKeeper),
-    "partyType" -> mandatory(text().verifying("supplementary.partyType.error", isContainedIn(allowedPartyTypes + "no")))
+    "partyType" -> requiredRadio("declaration.partyType.error")
+      .verifying("declaration.partyType.error", isContainedIn(allowedPartyTypes + "no"))
```
MaciejRewera, line 57: What is the purpose of `"no"` here ?

app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -43,16 +48,18 @@ object DeclarationAdditionalActors extends DeclarationPage {
...
+      .transform[Option[String]](choice => Option(choice), choice => choice.getOrElse(""))
   )(form2Model)(model2Form)
 
-  private def eoriMappingFor(partyType: String) =
-    mandatoryIfEqual("partyType", partyType, Eori.mapping("supplementary"))
+  private def eoriMappingFor(partyType: String): Mapping[Option[Eori]] =
+    mandatoryIfEqual("partyType", partyType, Eori.mapping("declaration"))
```
MaciejRewera, line 62: 👍

app/views/components/gds/removeButton.scala.html
```diff
@@ -23,7 +23,7 @@
 @(value: Option[String] = None, hiddenLabel: Option[String] = None)(implicit messages: Messages)
 
 @govukButton(Button(
- classes = "govuk-button--secondary",
+ classes = "govuk-button--secondary govuk-!-margin-0",
```
MaciejRewera, line 26: Nit: Just making sure: this change is to be implemented for all remove buttons in the service?

app/models/declaration/DeclarationAdditionalActorsData.scala
iservport: It is also validated elsewhere. I introduced this because we now have a "No" option that is sent as `partyType`. Without this, the "No" option would add a new empty actor to the collection.

app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -17,23 +17,28 @@
...
 
-  def toJson: JsValue = Json.toJson(this)(DeclarationAdditionalActors.format)
+  def isAllowed: Boolean = partyType.exists(allowedPartyTypes.contains)
+
+  def toJson: JsValue = Json.toJson(this)(format)
 }
 
 object DeclarationAdditionalActors extends DeclarationPage {
 
+  val maxNumberOfActors = 99
```
MaciejRewera, line 40: To me this field should be a responsibility of `DeclarationAdditionalActorsData` class.

app/forms/declaration/DeclarationAdditionalActors.scala
iservport: "No" is a valid choice.

app/forms/declaration/DeclarationAdditionalActors.scala
PatrykRudnicki: ahh, sorry, my fault, I didn't see that you changed the existing method :)

app/views/components/gds/removeButton.scala.html
iservport: I don't know, probably yes.

app/forms/declaration/DeclarationAdditionalActors.scala
iservport: Good idea.

app/forms/declaration/DeclarationAdditionalActors.scala
MaciejRewera: So maybe it could be added to `allowedPartyTypes`, or at least to `PartyType` ?

app/models/declaration/DeclarationAdditionalActorsData.scala
MaciejRewera: I am aware of that. This is why I suggested using two methods together. This way the method is more universal.
As it is in Iterator pattern, you have `next()` and `hasNext()` methods. And using `next()` without the other one may lead to `NullPointerException` being thrown. This design decision has been made not to break single-responsibility principle by putting both operations into single method.

PatrykRudnicki: I slightly agree with @iservport .

Method addActor without validation if it's allowed doesn't make any sense, because you can't use it.

I'll rename it to `addActorIfCorrect` or something more descriptive.

Creating method to add actor to the list is like a wrapper for method inside the Sequence which do add

PatrykRudnicki: I'm only not sure if this copy is necessary. I mean I don't know what is better to copy or to create a new object with new actors

MaciejRewera: So it seems that probably the model could be better designed - not to allow adding values that are incorrect. Maybe through the use of types? But this is probably not for now.
Regarding `copy`, in my opinion it is not required, because you will copy the collection already with `:+` call. And in "else" case you are still copying the whole sequence, although it is not needed.
As a side note: is it necessary for the new actor to be appended to the sequence? It requires copying the whole sequence so it has O(n) time complexity. Prepending is O(1) and should be favoured.

app/forms/declaration/DeclarationAdditionalActors.scala
iservport: There are two situations: 1) when I need to check if I can add/validate an eori, and 2) when I need to see if all radios choices are valid. I would need two collections, but I opted for one and the second being a simple result from the first.

app/models/declaration/DeclarationAdditionalActorsData.scala
iservport: Good tip about prepending. And the copy can be avoided when `actor.isAllowed` is false as well.

iservport: `if (actor.isAllowed) DeclarationAdditionalActorsData(actor +: actors) else this`

app/models/declaration/DeclarationAdditionalActorsData.scala
```diff
@@ -20,13 +20,23 @@ import forms.declaration.DeclarationAdditionalActors
...
 object DeclarationAdditionalActorsData {
+
   implicit val format = Json.format[DeclarationAdditionalActorsData]
 
+  val maxNumberOfActors = 99
+
   val formId = "DeclarationAdditionalActorsData"
 
-  val maxNumberOfItems = 99
+  def actorsValidator(actor: DeclarationAdditionalActors, actors: Seq[DeclarationAdditionalActors]): Option[Seq[(String, String)]] =
```
MaciejRewera, line 35: Nit: This is a method so it more appropriate to have a verb form in name: `validateActors` in this case.

app/views/components/gds/removeButton.scala.html
PauloPortugal: Should we go with with buttons or links?

app/forms/Mapping.scala
```diff
@@ -20,16 +20,26 @@ import play.api.data.format.Formatter
...
 object Mapping {
-
   def requiredRadio(requiredKey: String = "error.required"): FieldMapping[String] =
     of(radioFormatter(requiredKey))
 
-  private def radioFormatter(requiredKey: String): Formatter[String] = new Formatter[String] {
+  def requiredRadio(requiredKey: String, choices: Seq[String]): FieldMapping[String] =
+    of(radioFormatter(requiredKey, choices))
+
+  def optionalRadio(requiredKey: String = "error.required", choices: Seq[String]): FieldMapping[String] =
```
MaciejRewera, line 29: Nit: Just realized that `optionalRadio` and `"error.required"` does not seem right here. Probably we need to think about more apropriate naming.

PR #816, opened 2020-04-03
CEDS-2229 - Section 5 of the Standard Declaration journey to GDS

SteveSugden:


app/views/declaration/transport_leaving_the_border.scala.html
```diff
@@ -19,50 +19,110 @@
 @import forms.declaration.ModeOfTransportCode._
 @import forms.declaration.TransportLeavingTheBorder
 @import models.requests.JourneyRequest
-@import views.Title
-@import views.components.inputs.RadioOption
+@import uk.gov.hmrc.govukfrontend.views.html.components.{Empty => _, _}
```
PatrykRudnicki, line 22: Why this rename `Empty => _` or I understand it incorrectly?

SteveSugden: Because there is an `Empty` class in ModeOfTransportCode that conflicts with the one in `govukfrontend`

PR #823, opened 2020-04-07
CEDS-2237 - Improve 'Other Parties' page

iservport:
Cleanup 2.

Additional comments addressed:

1) Add the right message 'Select an option from the list' - ok
2) Change message: Enter a valid EORI. Append "For example, ..." - ok
3) When save and continue is clicked when we go back the No option should be selected - ok


conf/messages.en
```diff
@@ -377,7 +377,7 @@ declaration.additionalActors.partyType = Party type
 declaration.additionalActors.maximumAmount.error = Maximum amount of parties added
 declaration.additionalActors.duplicated.error = Party's details already added
 declaration.additionalActors.add.hint = Add party
-declaration.eori.empty = Enter a valid EORI
+declaration.eori.empty = Enter a valid EORI number. For example, GB123456789098.'
```
MaciejRewera, line 380: You have asterisk at the end of this line

app/forms/Mapping.scala
```diff
@@ -42,4 +42,5 @@ object Mapping {
 }
 case object NoneOfTheAbove {
   val value: String = "no"
+  val option: Option[String] = Some(value)
```
MaciejRewera, line 45: In my opinion this is unnecessary clutter. I'd rather vote for writing `Some(NoneOfTheAbove.value)`.
The length is similar but it saves you from adding fields just because some clients need it in a different format.

PR #826, opened 2020-04-08
CEDS-2275 - Add Transport Payment screen into Clearance Request Journey

iservport:
There is currently no option for Declarants to add Transport Payment (data element 4/2) when completing a Clearance Request. Declarants should provide details of the transport payment when only for an EXS customs clearance request.

### DIFF
--- a/app/controllers/declaration/DepartureTransportController.scala
+++ b/app/controllers/declaration/DepartureTransportController.scala
@@ -68,7 +68,7 @@ class DepartureTransportController @Inject()(
   private def nextPage(mode: Mode)(implicit request: JourneyRequest[AnyContent]): Result =
     request.declarationType match {
       case DeclarationType.CLEARANCE =>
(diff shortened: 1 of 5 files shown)


conf/messages.en
```diff
@@ -640,7 +640,7 @@ declaration.transportInformation.active.help-item1 = The details of the transpor
 declaration.transportInformation.active.help-item2 = The country the transport is registered in, go to data element 7/15 Nationality of active means of transport crossing the border.
 
 
-declaration.transportInformation.transportPayment.paymentMethod = If this is an express consignment, tell us how you paid for the transport
+declaration.transportInformation.transportPayment.paymentMethod = How is the transport paid for?
```
SteveSugden, line 643: I changed this recently.   The previous title is correct I believe - see https://exports-declaration-prototype.herokuapp.com/version31/dev-ready/active-transport-payment-method?what-declaration-type=standard

SteveSugden: Then again, you have a screenshot on your ticket.  I'd check with design

PR #827, opened 2020-04-08
CEDS-2278 - only show 'OSR' pages if procedure code is '1042'

SteveSugden:


app/controllers/declaration/FiscalInformationController.scala
```diff
@@ -43,10 +44,18 @@ class FiscalInformationController @Inject()(
...
+      request.cacheModel
+        .itemBy(itemId)
+        .flatMap(_.procedureCodes)
+        .flatMap(_.procedureCode)
+        .exists(code => !ProcedureCodesData.osrProcedureCodes.contains(code))
 
     if (fastForward && cacheContainsFiscalReferenceData) {
       navigator.continueTo(routes.AdditionalFiscalReferencesController.displayPage(mode, itemId))
+    } else if (fastForward && cacheItemIneligibleForOSR) {
+      navigator.continueTo(routes.ProcedureCodesController.displayPage(mode, itemId))
```
PatrykRudnicki, line 58: Just to confirm this redirection is correct? I thought that ProcedureCodes is a page before Fiscal Information

SteveSugden: That is for "back link" navigation.   Perhaps the "fastforward" variable could be better named "back navigating"?

The logic is when coming back from the next page (Commodity Code) we direct to FiscalInfo.   We then either redirect one page forward to the VAT page if there is existing fiscal data (old implementation), re-direct to the page before if OSR not valid due to existing Procedure Code (new logic) or stay here.

I've tested this manually and it all appears to work.   It is a bit "clunky" though.

SteveSugden: There are no existing tests in the FiscalControllerSpec for "fast-forward" BTW - adding those now including the new ProcedureCode invalid logic.

PatrykRudnicki: Ahh, I forgot about it. Thanks for explanation!

That's right.

app/models/declaration/ExportItem.scala
```diff
@@ -40,19 +40,25 @@ case class ExportItem(
...
-    if (fiscalInformation.exists(_.onwardSupplyRelief == yes)) additionalFiscalReferencesData.isDefined
-    else fiscalInformation.isDefined
+  private def isProcedureCodesAndFiscalInformationComplete = {
+
+    def isFiscalInformationCompleted: Boolean =
+      if (hasFiscalReferences) additionalFiscalReferencesData.isDefined
+      else fiscalInformation.isDefined
+
+    procedureCodes.flatMap(_.procedureCode).isDefined &&
+    (procedureCodes.flatMap(_.procedureCode).exists(code => !ProcedureCodesData.osrProcedureCodes.contains(code)) || isFiscalInformationCompleted)
```
PatrykRudnicki, line 60: I'm just wondering if the `isFiscalInformationCompleted` must be in exists method or outside it.

SteveSugden: I this this is right - i.e. outside.   The procedure code be defined and either not be "1042" or fiscal information must be provided.

PR #830, opened 2020-04-09
CEDS-2064 - Declarant EORI

SteveSugden:
This PR changes the declarant EORI page from asking the user to enter their EORI to presenting them with the EORI from their login credentials and asking them to confirm it's correct.

I've not changed the cache model.  If the user confirms that the EORI is correct I copy it into the existing DeclarantDetails structure.

I've the user select's "No" they are logged out and shown a new "ineligible" page (content subject to change).

Finally, the declarant's EORI is presented on the summary screen as the first value in the Parties Section and the ability to 'change' has been removed.

### DIFF
--- a/app/controllers/declaration/DeclarantDetailsController.scala
+++ b/app/controllers/declaration/DeclarantDetailsController.scala
@@ -18,7 +18,9 @@ package controllers.declaration
 
 import controllers.actions.{AuthAction, JourneyAction}
 import controllers.navigation.Navigator
-import forms.declaration.DeclarantDetails
+import forms.common.Eori
+import forms.common.YesNoAnswer.YesNoAnswers
+import forms.declaration.{DeclarantDetails, DeclarantEoirConfirmation, EntityDetails}
 import javax.inject.Inject
 import models.requests.JourneyRequest
 import models.{ExportsDeclaration, Mode}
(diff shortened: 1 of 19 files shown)


app/controllers/declaration/DeclarantDetailsController.scala
```diff
@@ -43,28 +45,30 @@ class DeclarantDetailsController @Inject()(
...
         formWithErrors => Future.successful(BadRequest(declarantDetailsPage(mode, formWithErrors))),
-        validDeclarantDetails =>
-          updateCache(validDeclarantDetails)
-            .map(_ => navigator.continueTo(mode, controllers.declaration.routes.ExporterDetailsController.displayPage))
+        validForm =>
+          if (validForm.answer == YesNoAnswers.yes)
+            updateCache(DeclarantDetails(EntityDetails(Some(Eori(request.eori)), None)))
+              .map(_ => navigator.continueTo(mode, controllers.declaration.routes.ExporterDetailsController.displayPage))
+          else
+            Future.successful(SeeOther(controllers.declaration.routes.NotEligibleController.displayNotDeclarant().url).withNewSession)
```
PatrykRudnicki, line 63: Why SeeOther and .url?

Is `Redirect(controllers.declaration.routes.NotEligibleController.displayNotDeclarant()).withNewSession`

app/controllers/declaration/DeclarantDetailsController.scala
```diff
@@ -43,28 +45,30 @@ class DeclarantDetailsController @Inject()(
...
       )
   }
 
-  private def form()(implicit request: JourneyRequest[AnyContent]): Form[DeclarantDetails] =
-    DeclarantDetails.form(request.declarationType)
+  private def form()(implicit request: JourneyRequest[AnyContent]): Form[DeclarantEoirConfirmation] =
```
PatrykRudnicki, line 67: Typo in the class name. You have `Eoir` not `Eori`

app/controllers/declaration/DeclarantDetailsController.scala
SteveSugden: Thanks

app/controllers/declaration/DeclarantDetailsController.scala
SteveSugden: Good spot.  Took me a minute to see even after you had pointed it out

test/unit/controllers/declaration/DispatchLocationControllerSpec.scala
```diff
@@ -62,7 +62,7 @@ class DispatchLocationControllerSpec extends ControllerSpec {
 
         withNewCaching(aDeclaration(withDispatchLocation(OutsideEU)))
 
-        val result = controller.displayPage(Mode.Normal)(getRequest())
+        private val result = controller.displayPage(Mode.Normal)(getRequest())
```
PatrykRudnicki, line 65: There is no point of making this result private. This is inside the test suite so this value cannot be accessed outside.

This comment to the whole file

test/unit/controllers/declaration/NotEligibleControllerSpec.scala
```diff
@@ -19,34 +19,45 @@ package unit.controllers.declaration
...
-        val result = controller.displayPage()(getRequest())
+        private val result = controller.displayNotEligible()(getRequest())
 
         status(result) must be(OK)
+        verify(notEligiblePage, times(1)).apply()(any(), any())
+      }
+
+      "display not eligible user is invoked" in new SetUp {
+
+        private val result = controller.displayNotDeclarant()(getRequest())
```
PatrykRudnicki, line 57: Here similar about private

test/unit/controllers/declaration/DispatchLocationControllerSpec.scala
SteveSugden: I made a few things private because IntelliJ had warnings about them (when it does the warning check just before doing a commit).  I normally ignore the warnings but on my new, super-fast machine I thought I'd take a look.

I'm not going to reverse these changes now that I have a precious approval.  :-)

PR #831, opened 2020-04-14
BAU-2275 - Add Transport Payment screen into Clearance Request Journey

iservport:
Ensure SUPPLEMENTARY journey is not valid for the Transport Payment page.

There is currently no option for Declarants to add Transport Payment (data element 4/2) when completing a Clearance Request. Declarants should provide details of the transport payment when only for an EXS customs clearance request.


PatrykRudnicki:
Please run scala formatter. I'll approve once again :)

PR #843, opened 2020-04-20
CEDS-2345 - Move the Transport leaving the border page to GDS

SteveSugden:


app/views/declaration/border_transport.scala.html
```diff
@@ -26,77 +26,107 @@
...
-                radioOption("europeanVesselIDNumber", EuropeanVesselIDNumber),
-                radioOption("nameOfInlandWaterwayVessel", NameOfInlandWaterwayVessel)
-            )
-        )
+@radioOption(id: String, value: String) = @{RadioItem(
+    id = Some(id),
+    value = Some(value),
+    content = Text(messages(s"declaration.transportInformation.meansOfTransport.$id")),
+    conditionalHtml = inputField(id, s"declaration.transportInformation.meansOfTransport.$id.label"),
+    checked = form("borderTransportType").value.contains(value))}
```
PatrykRudnicki, line 70: Is it worth to extract this as a component or it's only connected with this page? We will be able to reuse this part in more places.

PatrykRudnicki: As I see Github complicated my comment with old changes 😂

SteveSugden: Maybe.  But as Paulo has approved it's going in as it is.

PR #844, opened 2020-04-20
CEDS-2343 - revert to links for change and remove seals/containers

SteveSugden:


app/views/declaration/destinationCountries/routing_countries_summary.scala.html
```diff
@@ -51,31 +51,6 @@
 
 @(mode: Mode, form: Form[Boolean], countriesOfRouting: Seq[Country])(implicit request: JourneyRequest[AnyContent], messages: Messages)
 
-@routingCountriesTableOld = {
```
SteveSugden, line 54: Un-used code

PR #845, opened 2020-04-20
CEDS-2329 - Migrate 'Holder of Authorisation' to Gov Design System

iservport:


app/views/declaration/declaration_holder.scala.html
```diff
@@ -56,26 +66,23 @@
             )
         }
 
-        @components.fields.field_autocomplete(
+        @components.fields.field_accessible_autocomplete(
             field = form("authorisationTypeCode"),
             label = messages("supplementary.declarationHolder.authorisationCode"),
-            labelClass = Some("form-label-bold"),
+            labelClass = Some("govuk-label--m"),
```
SteveSugden, line 72: Is that label class needed?  I left this parameter as "None" when I did my first auto complete page (copying what Paulo did for the country of routing stuff)

iservport:
Clean the history.

PR #850, opened 2020-04-21
BAU - Display Tariff's 3rd item from body partial

PauloPortugal:


app/views/components/tariff_body_partial.scala.html
```diff
@@ -27,7 +27,7 @@
     }
     @if(m3.nonEmpty){
     <li>
-        @m2
+        @m3
```
PatrykRudnicki, line 30: Probably it will be more readable if this `m3` will have better name :D

SteveSugden: It's almost like varargs or indeed lists were a new concept.

PR #851, opened 2020-04-21
CEDS-2348 - Migrate NACT page to Gov Design System

SteveSugden:
New page as existing functionality with regards to add-to-list pattern with some design/content tweaks suggested by Chris.


app/config/AppConfig.scala
```diff
@@ -52,6 +52,7 @@ class AppConfig @Inject()(
 
   lazy val customsDeclarationsGoodsTakenOutOfEuUrl = loadConfig("urls.customsDeclarationsGoodsTakenOutOfEu")
   lazy val commodityCodesUrl = loadConfig("urls.commodityCodes")
+  lazy val nactCodesUrl = loadConfig("urls.nactCodes")
```
PatrykRudnicki, line 55: In the Future probably it will be worth to extract URLs to separate config class, but for now it's okay.

SteveSugden: That's probably a good idea - there are a lot of these "information" urls (i.e. just used for links in the UI).    Could probably consider them part of the content and put them in some form of messages file.   

I think AppConfig is just for stuff the application needs to function.  

But I was just following precedents.

PR #852, opened 2020-04-21
CEDS-2339 Move Saved Declarations page to GDS

MaciejRewera:


app/views/ViewDates.scala
```diff
@@ -19,16 +19,18 @@ package views
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 import java.time.temporal.TemporalAccessor
+import java.util.Locale
 
 object ViewDates {
 
   val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm")
-  val submissionDateTimeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM uuu 'at' HH:mm")
+  val submissionDateTimeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM uuu 'at' HH:mm").withZone(ZoneOffset.UTC)
```
SteveSugden, line 27: Is this part of your "dual times" investigations?  Did you mean to make this change in this PR?

app/views/components/gds/PaginationUtil.scala
```diff
@@ -0,0 +1,70 @@
...
+      throw new IllegalArgumentException("parameter pagesTotal must be bigger than 0")
+    if (currentPageIndex > pagesTotal)
+      throw new IllegalArgumentException("parameter currentPageIndex cannot be bigger than pagesTotal")
+    if (neighbourPagesAmount < 0)
+      throw new IllegalArgumentException("parameter neighbourPagesAmount cannot be negative")
+
+    val neighbourPagesNumbers = ((currentPageIndex - neighbourPagesAmount) to (currentPageIndex + neighbourPagesAmount))
+      .filter(pageNumber => pageNumber > 0 && pageNumber <= pagesTotal)
+
+    val leadingElements: List[PaginationItem] = abs(neighbourPagesNumbers.head - 1) match {
```
PatrykRudnicki, line 41: Is it possible that head will be empty?

What do you think about `.headOption.getOrElse()`?

app/views/declaration/transport_container_summary.scala.html
```diff
@@ -64,11 +64,11 @@
                         attributes = Map("id" -> s"containers-row$index-seals")
                     ),
                     TableRow(
-                        content = HtmlContent(changeLink(messages("site.remove"), messages("declaration.transportInformation.containers.remove.hint", container.id), TransportContainerController.submitContainerRemove(mode, container.id))),
+                        content = HtmlContent(changeLink(messages("site.remove"), Some(messages("declaration.transportInformation.containers.remove.hint", container.id)), TransportContainerController.submitContainerRemove(mode, container.id))),
```
SteveSugden, line 67: Good idea "changing" the changeLink to Link.   I realised after I merged that I was using it to "remove"

conf/application.conf
```diff
@@ -98,7 +98,7 @@ countryCodesCsvFilename = "code-lists/mdg-country-codes.csv"
 countryCodesJsonFilename = "code-lists/location-autocomplete-canonical-list.json"
 eu-country-codes-filename = "code-lists/mdg-country-codes-eu.csv"
 draft.timeToLive = "30d"
-pagination.itemsPerPage = 50
+pagination.itemsPerPage = 10
```
PatrykRudnicki, line 101: 👍

app/views/declaration/summary/related_documents.scala.html
```diff
@@ -18,12 +18,12 @@
...
 @import services.DocumentType
 @import uk.gov.hmrc.govukfrontend.views.html.components._
-@import views.html.components.gds.changeLink
+@import views.html.components.gds.link
 @import views.html.components.gds.summary_list
 
 @this(
         govukTable: GovukTable,
-        changeLink: changeLink,
+        changeLink: link,
```
PatrykRudnicki, line 26: What is the difference with this?

Those components are used on the summary pages

conf/application.conf
SteveSugden, line 101: Was that a product team decision to change to 10 or is this for testing only?

app/views/declaration/summary/related_documents.scala.html
SteveSugden: I think the summary page uses "Actions" to create the change links Patryk.    The name change is because this is used to create links for "Remove" as well as "Change"

conf/application.conf
MaciejRewera: Good catch, this was just for testing. I'll revert it.

app/views/components/gds/PaginationUtil.scala
MaciejRewera: I really doubt it since you have 4 parameters checks at the beginning of the method, so this list will be at least one element long.
But let me investigate how to make more error safe.

app/assets/stylesheets/partials/_pagination.scss
```diff
@@ -0,0 +1,47 @@
+
+.ceds-pagination {}
+
+.ceds-pagination__summary {
```
bambuchaAdm, line 4:
```
.ceds-pagination {
  &__summary {
    @extend .govuk-grid-column-one-half;
  }
}
```
Generay this allow you to DRY and not make silly typos in class names - at least in scss file

app/assets/stylesheets/partials/_pagination.scss
```diff
@@ -0,0 +1,47 @@
...
+.ceds-pagination__controls {
+  @extend .govuk-grid-column-one-half;
+  @extend .align-right;
+}
+
+.ceds-pagination__items {
+  display: inline;
+}
+
+// Can it be changed to class?
```
bambuchaAdm, line 17: It still in single block would be enough.
Is just item class as selectro is not enough ?
Maybe would be worth to merge this declaration on next one ?

app/assets/stylesheets/partials/_pagination.scss
```diff
@@ -1,47 +1,48 @@
...
+  &__items {
+    display: inline;
+  }
 
-// Can it be changed to class?
-.ceds-pagination__items .ceds-pagination__item {
-  display: inline;
-  margin: 0 5px;
-}
+  &__items &__item {
```
bambuchaAdm, line 17: I am not 100% sure if this construction in 100% - scss has some unexpected behaviour in terms of multiple & in directives

app/views/components/gds/PaginationUtil.scala
```diff
@@ -0,0 +1,74 @@
...
+
+package views.components.gds
+
+import java.lang.Math.abs
+
+import views.components.gds.PaginationUtil.PaginationItem._
+
+object PaginationUtil {
+
+  def paginationElements(currentPageIndex: Int, pagesTotal: Int, neighbourPagesAmount: Int = 1): Seq[PaginationItem] = {
```
bambuchaAdm, line 25: Is not beter to take here Paginated ? Most of those checks would be gone in this case - parameters are passed exactly from paginated

app/assets/stylesheets/partials/_pagination.scss
MaciejRewera: It is working fine.

app/views/components/gds/PaginationUtil.scala
MaciejRewera: Not `Paginated` but `Page` and it would only reduce the first two parameters.
It is sensible to do, but I'd probably then need to implement all necessary checks in `Page` class as it has none at the moment.

PR #857, opened 2020-04-22
CEDS-2346 Convert page layout to GDS Design System

PauloPortugal:
**Depends on** https://github.com/hmrc/cds-exports-acceptance/pull/418
_(Smoke tests are passing locally)_

### DIFF
--- a/app/views/declaration/supervising_customs_office.scala.html
+++ b/app/views/declaration/supervising_customs_office.scala.html
@@ -14,50 +14,78 @@
  * limitations under the License.
  *@
 
-@import controllers.declaration.routes
-@import models.requests.JourneyRequest
-@import forms.declaration.SupervisingCustomsOffice
 @import controllers.navigation.Navigator
+@import forms.declaration.SupervisingCustomsOffice
 @import services.CustomsOffices
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+@import views.html.components.gds._
+@import models.requests.JourneyRequest
 @import services.view.AutoCompleteItem
 @import views.Title
+@import views.BackButton
+@import views.components.gds.Styles
+@import scala.collection.immutable
+@import play.twirl.api.HtmlFormat
 
-@this(main_template: views.html.main_template)
-
+@this(
+    govukLayout: gdsMainTemplate,
+    govukButton: GovukButton,
+    govukRadios: GovukRadios,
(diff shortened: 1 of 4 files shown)


app/views/declaration/supervising_customs_office.scala.html
```diff
@@ -14,50 +14,78 @@
...
+    @components.details(messages("site.details.summary_text_these")) {
+        @Html(messages("declaration.type.supervisingCustomsOfficeSummaryText", components.details_content_link()))
+    }
+}
 
-            @components.back_link(Navigator.backLink(SupervisingCustomsOffice, mode)(request))
+@autocomplete = {
+    @components.fields.field_accessible_autocomplete(
+        field = form("supervisingCustomsOffice"),
+        label = "",
```
PatrykRudnicki, line 60: I'm thinking if the empty label cause any accessibility issues.

app/views/declaration/supervising_customs_office.scala.html
```diff
@@ -14,50 +14,78 @@
...
+        @errorSummary(form.errors)
 
-            @components.fields.field_autocomplete(
-                field = form("supervisingCustomsOffice"),
-                label = messages(""),
-                labelClass = Some("form-label-bold"),
-                hintText = None,
-                emptySelectValue = messages("supplementary.packageInformation.typesOfPackages.empty"),
-                items = AutoCompleteItem.fromSupervisingCustomsOffice(CustomsOffices.all))
+        @sectionHeader(messages("declaration.warehouse.supervisingCustomsOffice.sectionHeader"))
```
PatrykRudnicki, line 75: I think that the secitonHeader can be moved to the autocomplete as label with custom class, but not sure about it.

app/views/declaration/supervising_customs_office.scala.html
PauloPortugal: Yes I agree, but the prototype does not have one. I have asked this question to Chris, but he said that the screen reader does not find a problem

app/views/declaration/supervising_customs_office.scala.html
PauloPortugal: Hum... I am not sure about that, as this is the way we tend to do things. Admittedly there are two labels missing as part of the autocomplete component: one for the select tag and another for the input text tag.

PR #859, opened 2020-04-23
CEDS-2330 - Migrate 'Consignee Details' to Gov Design System

iservport:


conf/messages.en
```diff
@@ -325,7 +325,7 @@ declaration.declarant-details.help.bodyText = Go to data element 3/18 Declarant
 declaration.declarant.eori.empty = Enter your EORI number
 declaration.declarant.eori.error.format = Enter a valid EORI number, for example GB123456789000
 
-supplementary.consignee.title = What is the consignee’s address?
+declaration.consignee.title = What is the consignee’s address?
 supplementary.consignee.title.hint = Enter the consignee’s EORI number or name and business address.
 supplementary.consignee.eori.info = Consignee’s EORI number
 supplementary.consignee.address.info = Consignee’s name and business address
```
PatrykRudnicki, line 331: Can you change other messages related with this page to use `declaration` in message key?

test/views/declaration/ConsigneeDetailsViewSpec.scala
```diff
@@ -32,18 +32,58 @@ import views.html.declaration.consignee_details
...
+  def assertAllAddressValidation(errorKey: String, fieldName: String, address: Address, fields: Seq[String] = allFields)(
+    implicit request: JourneyRequest[_]
+  ): Unit = {
+
+    val view = createViewWithAddressError(address)
+    fields.filterNot(_ == fieldName).foreach { key =>
+      view must containErrorElementWithTagAndHref("a", s"#details_address_$key")
+      view must containErrorElementWithMessage(s"supplementary.address.$key.$errorKey")
+    }
+  }
```
PatrykRudnicki, line 72: Is it worth to create this method?

This part: `fields.filterNot(_ == fieldName)` I don't understand why we need all fields except the one that is passed to this.

I think this assertion can be made directly inside the code using the second method, where we can iterate through the sequence and invoke `assertAddressValidation`.

If you want to stay with this method it can be like `fields.filterNot(_ == fieldName).foreach(assertAddressValidation(_))`

There is the same logic in both methods, but one of them iterate through the sequence

PatrykRudnicki: In general you can add view as an argument. In that case you should be able to stay with one method and iterate through the sequence in the specific test

iservport: Regarding this fragment `This part: fields.filterNot(_ == fieldName) I don't understand why we need all fields except the one that is passed to this.`,  I used it to replace a repetitive code that was present in the previous version. I preserved the intent of the previous test to verify if multiple errors are shown correctly.

conf/messages.en
iservport: The other messages are not actually used. I will delete them.

PR #863, opened 2020-04-24
CEDS-2351 Migrate Package Information page to GDS

MaciejRewera:


app/views/declaration/package_information.scala.html
```diff
@@ -23,99 +23,109 @@
...
+        @packageInfoTable
+
+        @components.fields.field_accessible_autocomplete(
+          field = form("typesOfPackages"),
+          label = messages("declaration.packageInformation.typesOfPackages"),
+          labelClass = Some("govuk-label--m"),
+          hintText = None,
+          emptySelectValue = messages("declaration.packageInformation.typesOfPackages.empty"),
+          items = AutoCompleteItem.fromPackageType(PackageTypes.all),
+          'otherErrorFields -> Seq("typesOfPackages")
```
PatrykRudnicki, line 108: Why it's with `'` at the beginnig? I see that previously it was the same

MaciejRewera: No idea. I allowed myself to copy the previously working solution as much as it is possible.

PR #869, opened 2020-04-28
CEDS-2325 - Migrate 'Cancel Declaration' to Gov Design System

iservport:


app/views/cancel_declaration.scala.html
```diff
@@ -16,61 +16,90 @@
...
 
-    @helper.form(action = CancelDeclarationController.onSubmit(), 'autoComplete -> "off") {
-        @helper.CSRF.formField
-
-        @components.back_link(controllers.routes.ChoiceController.displayPage(Some(Choice(CancelDec))))
+    @formHelper(action = CancelDeclarationController.onSubmit(), 'autoComplete -> "off") {
+        @errorSummary(form.errors)
 
-        @components.error_summary(form.errors)
+        <div class="govuk-grid-column-two-thirds">
```
SteveSugden, line 48: Are you sure we need this `<div>`?   We don't tend to add that on other pages.

app/views/cancel_declaration.scala.html
```diff
@@ -16,61 +16,90 @@
...
 
-        @components.fields.field_text(
-            field = form("functionalReferenceId"),
-            label = Some(messages("cancellation.functionalReferenceId")),
-            labelClass = Some("bold-small")
-        )
+            @exportsInputText(
+                field = form("functionalReferenceId"),
+                labelKey = "cancellation.functionalReferenceId",
+                labelClasses = ""
```
SteveSugden, line 55: Should there not be a label class?  Prototype seems to be `govuk-label`

app/views/cancel_declaration.scala.html
iservport: It was necessary to solve the problem of showing the content in the two-thirds format, as shown in the prototype.

app/views/cancel_declaration.scala.html
iservport: The `govuk-label` class is added by default:
```
<label class="govuk-label" for="functionalReferenceId">
        Enter the DUCR
      </label>
```

SteveSugden: You are right.  My bad.  With my suggestions merged in we now have double label!
```<label class="govuk-label govuk-label" for="mrn">
        Enter the Movement Reference Number (MRN)
      </label>```

I think that error is repeated elsewhere on our pages.  Not a biggy.  Maybe we can clean up the exportsInputText component later.

PR #875, opened 2020-04-28
CEDS-2340 - Move the Declaration errors page to GDS

SteveSugden:


app/views/rejected_notification_errors.scala.html
```diff
@@ -16,56 +16,68 @@
...
 
-@this(main_template: views.html.main_template, appConfig: AppConfig)
+@this(
+        govukLayout: gdsMainTemplate,
+        pageTitle: pageTitle,
+        sectionHeader: sectionHeader,
+        link: link,
+        govukTable: GovukTable,
+        govukSummaryList : GovukSummaryList,
+        appConfig: AppConfig
```
PatrykRudnicki, line 31: Probably worth to consider in the Future about the new config class for errors descriptions, to not inject the whole appConfig to the view, but it's not something that must be done now.

PR #876, opened 2020-04-29
CEDS-2341 - Move the submissions page to the GDS

iservport:


app/views/submissions.scala.html
```diff
@@ -24,44 +24,80 @@
...
-    fullWidth = true
-) {
+@headCell(key: String) = @{
+    HeadCell( content = Text(messages(s"submissions.$key.header")) )
+}
 
-    @components.back_link(
-        href = routes.ChoiceController.displayPage(Some(Choice(Submissions))),
-        backMessage = "site.backToSelectionPage"
+@ucrLink(submission: Submission) = @{
```
SteveSugden, line 44: You can use the components.gds.link component to do this.  Could even add some accessible "hidden" text using this method (even though the prototype doesn't have this - could check with Chris/James)

PR #881, opened 2020-05-01
CEDS-2065 - split representative_details into three pages

SteveSugden:
This change involves taking one page with 2 questions -
RepresentativeDetails - Eori and Status (representation type)

and splitting it into 3 pages with one question each - 
RepresentativeAgent - representing another agent? yes/no
RepresentativeEntity - representatives Eori
RepresentativeStatus - representative type

I've re-used the existing RepresentativeDetails model with the addition of the new "agent" question to minimise the changes needed to the back-end and database.

The navigation logic depends on the answer to the "agent" question
- Yes - both Eori and type pages are shown and are required.
- No - only the type page is shown and is optional


app/controllers/declaration/RepresentativeStatusController.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  private def nextPage(declarationType: DeclarationType): Mode => Call =
+    declarationType match {
+      case DeclarationType.SUPPLEMENTARY =>
+        controllers.declaration.routes.DeclarationAdditionalActorsController.displayPage
+      case DeclarationType.STANDARD | DeclarationType.SIMPLIFIED | DeclarationType.OCCASIONAL | DeclarationType.CLEARANCE =>
+        controllers.declaration.routes.CarrierDetailsController.displayPage
+    }
+
+  private def navigationForm()(implicit request: JourneyRequest[AnyContent]) =
```
PatrykRudnicki, line 71: I always prefer to have the return type for private methods too.

If we're returning `RepresentativeStatus` or `RepresentativeEntity` it's good to understand what this method return in general.

Sometimes with this it's easy to catch returning `Any`

app/controllers/declaration/RepresentativeStatusController.scala
```diff
@@ -0,0 +1,91 @@
...
+    if (request.cacheModel.parties.representativeDetails.flatMap(_.details).isDefined) RepresentativeStatus else RepresentativeEntity
+
+  override protected def updateExportsDeclarationSync(
+    update: ExportsDeclaration => Option[ExportsDeclaration]
+  )(implicit hc: HeaderCarrier, ec: ExecutionContext, request: JourneyRequest[_]): Future[Option[ExportsDeclaration]] =
+    super.updateExportsDeclarationSync(update)
```
PatrykRudnicki, line 77: Is this method necessary? Why we can't invoke it directly?

app/views/declaration/summary/parties_section.scala.html
```diff
@@ -98,26 +98,47 @@ <h2 class="govuk-heading-m">Parties</h2>
             isEoriDefault = false
         )).getOrElse(Seq.empty)
 
+        ++ (Seq.empty :+ declarationData.parties.representativeDetails.map(_.representingOtherAgent).map(agent =>
```
PatrykRudnicki, line 101: Instead of adding Seq.empty to element is enough to wrap element in Sequence

`Seq.empty :+ element` == `Seq(element)`

app/controllers/declaration/RepresentativeStatusController.scala
SteveSugden: No idea how that method got there.  I thought it might have come from the original RepresentativeDetailsControler, but no.  It was all my own work.  I'll delete it.

PR #884, opened 2020-05-04
CEDS-2371 Add error change link on rejected notifications page

PatrykRudnicki:


app/controllers/declaration/RoutingCountriesSummaryController.scala
```diff
@@ -52,7 +52,7 @@ class RoutingCountriesSummaryController @Inject()(
     if (countries.nonEmpty) {
       Ok(routingCountriesSummaryPage(mode, RoutingQuestionYesNo.form(), countries))
     } else {
-      navigator.redirectTo(mode, controllers.declaration.routes.RoutingCountriesController.displayRoutingQuestion(_))
+      navigator.continueTo(mode, controllers.declaration.routes.RoutingCountriesController.displayRoutingQuestion(_))
```
SteveSugden, line 55: There seems to be no difference between`continueTo(mode: Mode, factory: Mode => Call)` and `redirectTo(mode: Mode, factory: Mode => Call)`

i wonder if we can get rid of one of them?

SteveSugden: Ignore - hadn't seen your proposed Navigator changes

app/controllers/navigation/Navigator.scala
```diff
@@ -37,23 +38,23 @@ import uk.gov.hmrc.http.HeaderCarrier
 
 class Navigator @Inject()(appConfig: AppConfig, auditService: AuditService) {
 
-  def continueTo(call: Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
```
SteveSugden, line 40: 👍

app/controllers/navigation/Navigator.scala
```diff
@@ -37,23 +38,23 @@ import uk.gov.hmrc.http.HeaderCarrier
...
-        auditService.auditAllPagesUserInput(AuditTypes.SaveAndReturnSubmission, req.cacheModel)
-        goToDraftConfirmation()
-      case _ => Results.Redirect(call)
-    }
-
   def continueTo(mode: Mode, factory: Mode => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
     redirectTo(mode.next, factory)
 
-  def redirectTo(mode: Mode, factory: Mode => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
+  private def redirectTo(mode: Mode, factory: Mode => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
```
SteveSugden, line 44: :+1 for "private"

app/controllers/declaration/RoutingCountriesSummaryController.scala
PatrykRudnicki: Yeah, I saw that the methods did similar stuff and we used different method based on the signature. Passing mode is not a big deal, so I tried to use only one method and rest logic make private and inside navigator.

I thought about removing one of those methods, but the PR is big enough so I thought that first cleanup to expose only one method outside of Navigator will be good first step.

app/models/Pointer.scala
```diff
@@ -62,6 +62,42 @@ case class Pointer(sections: Seq[PointerSection]) {
...
+      itemNoOpt.flatMap { itemNo =>
+        declaration.items.find(_.sequenceId.toString == itemNo).map { item =>
+          url.replace(itemIdPlaceholder, item.id)
+        }
+      }.getOrElse(defaultItemsUrl)
+    } else if (url.contains(containerIdPlaceholder)) {
+      val defaultContainersUrl = "/customs-declare-exports/declaration/containers"
+      val containerNoOpt = sequenceArgs.headOption
+
+      containerNoOpt.flatMap { containerNo =>
```
SteveSugden, line 93: Not seen Try-Catch in a long time.  Are we worried that the containerNo might not be an int?

conf/code-lists/errors-dms-rej-list.csv
```diff
@@ -1,199 +1,199 @@
...
-CDS71262,"Licence Error: Document Status Code ""EA"" can only be declared on Type Y declarations",
-CDS73329,Licence Error: Licence cannot be declared on declaration category/declaration type combination,
-CDS77000,Obligation Error: ID must be provided for declared Document Code,
-CDS77002,Obligation Error: Document Status Code must be provided for declared Document Code,
-CDS77003,Licence Error: DUCR must be provided for an Import declaration with a Licence,
-CDS14000,Credibility check: incredible quota value found,
\ No newline at end of file
+CDS10001,Obligation Error: Mandatory Data Element has not been provided,Obligation error: Mandatory Data Element has not been provided,
+CDS10002,Obligation Error: Data Element is not allowed,Obligation error: Data Element is not allowed,
+CDS10003,Obligation Error: Number of Occurrences exceeds what is allowed for the Data Element,Obligation error: Number of Occurrences exceeds what is allowed for the Data Element,
```
SteveSugden, line 3: Why is each line ending in a comma?

test/base/MockConnectors.scala
```diff
@@ -29,11 +29,13 @@ import org.mockito.ArgumentMatchers._
 import org.mockito.Mockito.when
 import org.mockito.invocation.InvocationOnMock
 import org.mockito.stubbing.{Answer, OngoingStubbing}
+import org.scalatest.BeforeAndAfterEach
 import org.scalatestplus.mockito.MockitoSugar
 import uk.gov.hmrc.http.HeaderCarrier
 
 import scala.concurrent.{ExecutionContext, Future}
 
+// TODO This mock should extends BeforeAndAfterEach trait and has methods beforeEach and afterEach
```
SteveSugden, line 38: Is this for Tech Debt?

app/services/model/RejectionReason.scala
```diff
@@ -25,7 +25,7 @@ import play.api.libs.json.{Json, OFormat}
 
 import scala.io.Source
 
-case class RejectionReason(code: String, cdsDescription: String, exportsDescription: String, pointer: Option[Pointer])
+case class RejectionReason(code: String, cdsDescription: String, exportsDescription: String, url: Option[String], pointer: Option[Pointer])
```
SteveSugden, line 28: Where are these url templates coming from?  I was expecting to see them defined in a config file our something similar.

conf/code-lists/errors-dms-rej-list.csv
PatrykRudnicki: because the last argument is `url`, we have only one url error right now, so needed to add the empty values for rest elements.

app/models/Pointer.scala
PatrykRudnicki: Not sure about it. I would like to be sure to don't have a random errors.

I can remove it :)

app/services/model/RejectionReason.scala
PatrykRudnicki: `conf/code-lists/errors-dms-rej-list.csv`

From this file.

https://github.com/hmrc/customs-declare-exports-frontend/pull/884/files#diff-6a90a36baf2cf59e5e17a114f2cd9261R109

test/base/MockConnectors.scala
PatrykRudnicki: Yeah, I thought about this. Not sure if we want to keep it or not.

app/models/Pointer.scala
SteveSugden: So if there is an error we just don't get a url?   If you do keep the try-catch, maybe add an error log message.  Then we could investigate and fix?

app/services/model/RejectionReason.scala
SteveSugden: Yes - I missed that in the massive diff.

SteveSugden: Hence my question about the commas.

conf/application.conf
```diff
@@ -87,6 +87,7 @@ microservice {
       default = disabled
       ead = enabled
       sfus = enabled
+      changeErrorLink = enabled
```
PauloPortugal, line 90: :+1:

app/services/model/RejectionReason.scala
PauloPortugal: We need to move to `Json`, but this ticket is a chance to try and get design/product feedback ASAP.
Probably move a bit of this logic to a separate microservice... again... lets wait for the feedback from stakeholders before committing ourselves.

app/controllers/SubmissionsController.scala
```diff
@@ -60,18 +61,41 @@ class SubmissionsController @Inject()(
...
+    }.getOrElse(Future.successful(None))
+
+    actualDeclaration.flatMap {
+      case Some(_) => Future.successful(Redirect(controllers.declaration.routes.SummaryController.displayPage(Mode.Amend)))
+      case _       => createNewDraftDec(id, redirect)
+    }
+  }
+
+  def amendErrors(id: String, redirectUrl: String): Action[AnyContent] = authenticate.async { implicit request =>
+    val redirectUrlWithMode = redirectUrl + ErrorFix.queryParameter
```
PauloPortugal, line 77: :+1:

PR #885, opened 2020-05-04
Bau navigator enhancement

MaciejRewera:


app/controllers/navigation/Navigator.scala
```diff
@@ -46,14 +46,19 @@ class Navigator @Inject()(appConfig: AppConfig, auditService: AuditService) {
...
-    redirectTo(mode.next, factory)
+    FormAction.bindFromRequest match {
+      case SaveAndReturn =>
+        auditService.auditAllPagesUserInput(AuditTypes.SaveAndReturnSubmission, req.cacheModel)
+        goToDraftConfirmation()
+      case _ => Results.Redirect(factory(mode.next))
+    }
 
-  def redirectTo(mode: Mode, factory: Mode => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
+  def continueTo(mode: Mode, itemId: String, factory: (Mode, String) => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
```
PatrykRudnicki, line 56: What is the usage of this itemId?

When and where you want to use it?

MaciejRewera: I think it might be better to do this way:
`def continueTo(mode: Mode, itemId: ItemId, factory: (Mode, ItemId) => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =`

PatrykRudnicki: Yeah, but this method is handling the non items pages too. The question was why now the `itemId` is the mandatory for this method

PatrykRudnicki:
I think this will be covered in your new PR @MaciejRewera  or different approach.

In general I think we can close this PR. in case that it'll be necessary we can restore branch and reopen it.

PR #886, opened 2020-05-04
CEDS-2401 - Ask if Declarant is the Exporter

SteveSugden:
Add new page after confirming declarants EORI asking if declarant is exporter.
If answer is "Yes" then we skip the Export and all of the Representation pages.


app/views/declaration/summary/parties_section.scala.html
```diff
@@ -83,21 +83,34 @@ <h2 class="govuk-heading-m">Parties</h2>
     )
     @summaryList("declaration-parties-summary", None,
 
-        declarationData.parties.exporterDetails.map(export => eoriOrAddress(
+        (Seq.empty :+ declarationData.parties.declarantIsExporter.map(isExporter =>
```
PatrykRudnicki, line 86: Comment as always :)

Not `Seq.empty :+ but Seq(element)`

SteveSugden: It doesn't work.  I've tried it several times.  When I originally did the new summary page, when you last made that comment and this time.

PatrykRudnicki: I checked this. It's because  `declarationData.parties.declarantIsExporter` is an Option, not Sequence.

So in that case you need to wrap whole block into the Sequence., not just this element.

SteveSugden: I think each "part" of this produces a Seq[Option[SummaryListRow]] - driven mainly by the `EoriOrAddress` helper.

PR #891, opened 2020-05-06
CEDS-2124 - initial implementation

SteveSugden:


conf/application.conf
```diff
@@ -159,3 +159,8 @@ whitelist {
   enabled = false
   eori = []
 }
+
+timeoutDialog {
+  timeout="13min"
+  countdown="2min"
```
MaciejRewera, line 165: Is it working in this form? Thought it has to be sth like `13 MINUTES`. Depends on the parser I guess...

SteveSugden: Yes - seems to work.  I've been testing with 
```
timeoutDialog {
  timeout="130s"
  countdown="2min"
}
```

app/assets/stylesheets/customsdecexfrontend-app.scss
```diff
@@ -21,4 +21,5 @@ $govuk-fonts-path: "/customs-declare-exports/assets/lib/govuk-frontend/govuk/ass
 
 //Below are all the CSS Overwrites that are required in order for the UI to align with the prototype
 @import 'partials/_gdsSummaryLists.scss';
-@import 'partials/_gdsBase.scss';
\ No newline at end of file
+@import 'partials/_gdsBase.scss';
+@import 'partials/_timeout-dialog.scss';
```
PauloPortugal, line 25: :+1:

app/controllers/SessionTimeoutController.scala
```diff
@@ -0,0 +1,40 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future.successful
+
+class SessionTimeoutController @Inject()(authenticate: AuthAction, mcc: MessagesControllerComponents, sessionTimedOut: session_timed_out)(
+  implicit ec: ExecutionContext
+) extends FrontendController(mcc) with I18nSupport {
+
+  def signOut(): Action[AnyContent] = authenticate.async { implicit request =>
+    successful(Results.Redirect(routes.SessionTimeoutController.signedOut()).withNewSession)
+  }
```
PatrykRudnicki, line 35: You can remove `.async` and `successful`.

Async is only necessary when you have a Future. Here you don't have so this is not necessary, like you have in method below.

PR #892, opened 2020-05-06
CEDS-2366 Display cached Non-Uk Office of Exit

PauloPortugal:
The QA team found a couple of bugs:

1. When a Non-UK Office of Exit exists in cache and
we go back to the Office Of Exit page were overriding the Non-UK Office
of Exit data.
2. Fix back button for Simplified and Occasional journeys

### DIFF
--- a/app/forms/declaration/officeOfExit/OfficeOfExit.scala
+++ b/app/forms/declaration/officeOfExit/OfficeOfExit.scala
@@ -25,11 +25,15 @@ object OfficeOfExit {
   def from(officeOfExitOutsideUK: OfficeOfExitOutsideUK): OfficeOfExit =
     OfficeOfExit(Some(officeOfExitOutsideUK.officeId), Some(AllowedUKOfficeOfExitAnswers.no))
 
-  def from(officeOfExitInsideUK: OfficeOfExitInsideUK): OfficeOfExit = officeOfExitInsideUK.isUkOfficeOfExit match {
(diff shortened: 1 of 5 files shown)


app/forms/declaration/officeOfExit/OfficeOfExit.scala
```diff
@@ -30,6 +30,15 @@ object OfficeOfExit {
     case _                                => OfficeOfExit(None, Some(AllowedUKOfficeOfExitAnswers.no))
   }
 
+  def from(officeOfExitInsideUK: OfficeOfExitInsideUK, officeOfExit: OfficeOfExit): OfficeOfExit =
```
SteveSugden, line 33: Would it not be simpler to always call this method from OfficeOfExitController and then to
- always copy over the 'isUkOfficeOfExit' value
- update the 'officeId' only if it exists on the form, otherwise keep the existing value (if any)

e.g.  
`OfficeOfExit(officeOfExitInsideUK.officeId.else(officeOfExit.officeId), officeOfExitInsideUK. isUkOfficeOfExit)`

SteveSugden: Maybe rename `officeOfExit` to `existingValue` and change it to be Option[OfficeOfExit] as you might not have an existing value.  Would make the logic in the controller above easier.

PauloPortugal: The first comment won't work, as we incur the risk of populating the officeOfExit (outside UK) with an office from the UK.
I have updated the logic, hopefully more easier to read. This was actually my first approach with Leszek, but for some stupid reason ended up making it more complex than it needed to be.

PR #900, opened 2020-05-12
CEDS-2383 - fix for correct display on summary screen

SteveSugden:
Because we weren't updating the items.additionalInformation state when submitting the Y/N answer on the first page, the default value of "None" meant that the section was never rendered.

The fix is to update the cache value to Some(AdditionalInformationData(Seq.empty)).   This is the same value as would be set if the user answered 'Y' but then did not provide any data on the next page.     So as it stands, we cannot differentiate between "N" and "Y" followed by no data - both will be rendered on the summary screen as "No".

However, as the AdditionalItems page needs re-writing to remove the old "add-to-list" pattern, I think this is OK for now.

### DIFF
--- a/app/controllers/declaration/AdditionalInformationRequiredController.scala
+++ b/app/controllers/declaration/AdditionalInformationRequiredController.scala
@@ -21,14 +21,17 @@ import controllers.navigation.Navigator
 import forms.common.YesNoAnswer
 import forms.common.YesNoAnswer.YesNoAnswers
 import javax.inject.Inject
-import models.{Mode}
+import models.declaration.AdditionalInformationData
+import models.requests.JourneyRequest
+import models.{ExportsDeclaration, Mode}
 import play.api.data.Form
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, Call, MessagesControllerComponents}
 import services.cache.ExportsCacheService
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import views.html.declaration.additional_information_required
-import scala.concurrent.{ExecutionContext}
+
+import scala.concurrent.{ExecutionContext, Future}
 
 class AdditionalInformationRequiredController @Inject()(
   authenticate: AuthAction,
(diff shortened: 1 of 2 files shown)


app/controllers/declaration/AdditionalInformationRequiredController.scala
```diff
@@ -44,24 +47,27 @@ class AdditionalInformationRequiredController @Inject()(
     Ok(additionalInfoReq(mode, itemId, YesNoAnswer.form()))
   }
 
-  def submitForm(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+  def submitForm(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
```
SteveSugden, line 50: Add "async" because we are now returning a Future

app/controllers/declaration/AdditionalInformationRequiredController.scala
```diff
@@ -44,24 +47,27 @@ class AdditionalInformationRequiredController @Inject()(
...
     YesNoAnswer
       .form()
       .bindFromRequest()
       .fold(
-        (formWithErrors: Form[YesNoAnswer]) => BadRequest(additionalInfoReq(mode, itemId, formWithErrors)),
-        validYesNo => navigator.continueTo(mode, nextPage(yesNoAnswer = validYesNo, itemId))
+        (formWithErrors: Form[YesNoAnswer]) => Future.successful(BadRequest(additionalInfoReq(mode, itemId, formWithErrors))),
```
SteveSugden, line 55: Need to wrap the BadRequest into a Future now

app/controllers/declaration/AdditionalInformationRequiredController.scala
```diff
@@ -44,24 +47,32 @@ class AdditionalInformationRequiredController @Inject()(
...
+        validYesNo =>
+          updateCache(validYesNo, itemId).map { _ =>
+            navigator.continueTo(mode, nextPage(validYesNo, itemId))
+        }
       )
-
+  }
+  private def updateCache(yesNoAnswer: YesNoAnswer, itemId: String)(implicit r: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] = {
+    val updatedAdditionalInformation = yesNoAnswer.answer match {
+      case YesNoAnswers.yes => r.cacheModel.itemBy(itemId).flatMap(_.additionalInformation).getOrElse(AdditionalInformationData(Seq.empty))
```
SteveSugden, line 64: If "Yes" then we either keep the existing data or create a new AdditionalInformationData(Seq.empty)
If "No" we save a new AdditionalInformationData(Seq.empty) which will delete previous data if user changed from Yes to No

PR #901, opened 2020-05-13
CEDS-2376 - Move the TARIC code page to the two-step remove pattern

SteveSugden:
This is basically a clone of the previous PR for CEDS-2377 - NACT codes


app/views/declaration/nact_code_add_first.scala.html
```diff
@@ -75,7 +75,7 @@
                 RadioItem(
                     id = Some("code_yes"),
                     value = Some(YesNoAnswers.yes),
-                    content = Text(messages("declaration.transportInformation.containers.yes")),
+                    content = Text(messages("site.yes")),
```
PatrykRudnicki, line 78: Could you remove the old message keys from messages.en?

app/controllers/declaration/TaricCodeRemoveController.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  def submitForm(mode: Mode, itemId: String, code: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    form()
+      .bindFromRequest()
+      .fold(
+        (formWithErrors: Form[YesNoAnswer]) => Future.successful(BadRequest(taricCodeRemove(mode, itemId, code, formWithErrors))),
+        formData => {
+          formData.answer match {
+            case YesNoAnswers.yes =>
+              updateExportsCache(itemId, request.cacheModel.itemBy(itemId).flatMap(_.taricCodes).getOrElse(Seq.empty).filterNot(_.taricCode == code))
```
PatrykRudnicki, line 58: I think it can be worth to move this logic into updateExportsCache method.

Pass there code and do everything inside (not sure if it's the best idea, because of the method responsibility).

If it's not good idea at least we can extract preparing the data into variable to make it more readable

app/views/declaration/nact_code_add_first.scala.html
SteveSugden: I think that key is still used - by the containers page.  I just copied it in error during the NACT code PR

app/controllers/declaration/TaricCodeRemoveController.scala
SteveSugden: I don't get you scala bods sometimes.  I thought doing it all in a single statement was cool and readability was for wimps.

PatrykRudnicki: Hmm, I think everything is based on the situation, if you will need to pass just parties like `request.cacheModel.parties` that's okay, it's short and probably readable enought to be in the same line.

`request.cacheModel.itemBy(itemId).flatMap(_.taricCodes).getOrElse(Seq.empty).filterNot(_.taricCode == code)` This is quite big and manipulate the data, this is the reason that I thought it can be useful to extract this as a variable.

I approved PR, so this was just a non blocking proposal :)

SteveSugden: Well I'm going to change it.  And same code in NACT controller.  So you'll have to review all over again.  :-)

SteveSugden: Done

PR #899, opened 2020-05-12
CEDS-2406 EXS Declaration question with logic around parties pages

PatrykRudnicki:

PR #902, opened 2020-05-13
CEDS-2400 EIDR pages

MaciejRewera:
This PR adds a sub-journey related to Entry Into Declarant's Records.
As part of this there are also changes to Navigator class that allows to
have back link dependent on the cached data.

### DIFF
--- a/app/controllers/navigation/Navigator.scala
+++ b/app/controllers/navigation/Navigator.scala
@@ -117,16 +117,16 @@ object Navigator {
     case ChangeCountryPage            => controllers.declaration.routes.RoutingCountriesSummaryController.displayPage
     case GoodsLocationForm            => controllers.declaration.routes.DestinationCountryController.displayPage
     case DeclarationHolder            => controllers.declaration.routes.ConsigneeDetailsController.displayPage
-    case ConsignorEoriNumber       => controllers.declaration.routes.ExporterDetailsController.displayPage
-    case ConsignorDetails          => controllers.declaration.routes.ConsignorEoriNumberController.displayPage
-    case RepresentativeAgent       => controllers.declaration.routes.ConsignorDetailsController.displayPage
-    case OfficeOfExitInsideUK         => controllers.declaration.routes.LocationController.displayPage
-    case OfficeOfExitOutsideUK        => controllers.declaration.routes.OfficeOfExitController.displayPage
-    case SupervisingCustomsOffice     => controllers.declaration.routes.WarehouseIdentificationController.displayPage
-    case TransportLeavingTheBorder    => controllers.declaration.routes.SupervisingCustomsOfficeController.displayPage
-    case WarehouseIdentification      => controllers.declaration.routes.ItemsSummaryController.displayPage
-    case TotalPackageQuantity         => controllers.declaration.routes.OfficeOfExitController.displayPage
-    case page                         => throw new IllegalArgumentException(s"Navigator back-link route not implemented for $page on clearance")
+    case ConsignorEoriNumber          => controllers.declaration.routes.ExporterDetailsController.displayPage
+    case ConsignorDetails             => controllers.declaration.routes.ConsignorEoriNumberController.displayPage
+
+    case OfficeOfExitInsideUK      => controllers.declaration.routes.LocationController.displayPage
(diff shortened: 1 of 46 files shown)


app/controllers/navigation/Navigator.scala
```diff
@@ -273,6 +333,28 @@ object Navigator {
         common.orElse(specific)(page)(mode)
     }
 
+  def backLink(page: DeclarationPage, mode: Mode, cacheModel: ExportsDeclaration)(implicit request: JourneyRequest[_]): Call =
```
PatrykRudnicki, line 336: As I mentioned yesterday `cacheModel` is in the request. You're taking this from the same request to pass to the method.

PatrykRudnicki: WIthout this change, you don't need to add this cache model as argument anywhere

app/controllers/navigation/Navigator.scala
```diff
@@ -255,6 +266,55 @@ object Navigator {
...
+      controllers.declaration.routes.RepresentativeStatusController.displayPage(mode)
+
+  private def consignorEoriNumberClearancePreviousPage(cacheModel: ExportsDeclaration, mode: Mode): Call = {
+    val declarantEori: Option[Eori] = cacheModel.parties.declarantDetails.flatMap(_.details.eori)
+    val exporterEori: Option[Eori] = cacheModel.parties.exporterDetails.flatMap(_.details.eori)
+
+    if (declarantEori.isDefined && exporterEori.isDefined && declarantEori == exporterEori)
+      controllers.declaration.routes.DeclarantExporterController.displayPage(mode)
+    else
+      controllers.declaration.routes.ExporterDetailsController.displayPage(mode)
```
PatrykRudnicki, line 309: This logic should be based on the answer Are you Exporter

MaciejRewera: I am not the person to address this comment as I have only copied what was already there.
Seems we have a bug then.

PatrykRudnicki: It has been fixed in https://github.com/hmrc/customs-declare-exports-frontend/pull/899/files

PatrykRudnicki: I'll change it during rebasing

app/views/components/gds/Styles.scala
```diff
@@ -19,6 +19,8 @@ package views.components.gds
 object Styles {
 
   val gdsPageLegend = "govuk-fieldset__legend--l"
+  val gdsPageLegendLarge = "govuk-fieldset__legend--xl"
```
SteveSugden, line 22: Why are we adding "large" styles?   The plan was to change the definitions from "...l" to "...xl" at some point

MaciejRewera: Wasn't aware of that. '--l' did not match the prototype so I created those.
So should it be just '--l' ?

SteveSugden: Yes.  That is the reason for creating the constants - so we can switch to the "larger" styles en-mass when all pages are GDS

app/controllers/navigation/Navigator.scala
```diff
@@ -255,6 +265,55 @@ object Navigator {
...
+      controllers.declaration.routes.DeclarantExporterController.displayPage(mode)
+
+  private def carrierDetailsClearancePreviousPage(cacheModel: ExportsDeclaration, mode: Mode): Call =
+    if (cacheModel.parties.declarantIsExporter.exists(_.isExporter))
+      controllers.declaration.routes.DeclarantExporterController.displayPage(mode)
+    else
+      controllers.declaration.routes.RepresentativeStatusController.displayPage(mode)
+
+  private def consigneeDetailsSupplementaryPreviousPage(cacheModel: ExportsDeclaration, mode: Mode): Call =
+    if (cacheModel.`type` == SUPPLEMENTARY && cacheModel.parties.declarantIsExporter.exists(_.isExporter))
```
SteveSugden, line 296: This is already a "SUPPLEMENTARY" method so no need to check again?

MaciejRewera:
There are some major changes to be done before merging (rebasing onto other branches at least) so I will open a new PR with that.

PR #905, opened 2020-05-13
CEDS 2383 Split the additional information page into the two pages

alsahid1993:
https://jira.tools.tax.service.gov.uk/browse/CEDS-2383


alsahid1993:
Will merge once acceptance tests are complete

PR #911, opened 2020-05-15
CEDS-2387 - fix for submissions with pseudo status 'Pending'

SteveSugden:
These are currently being displayed in all three tabs - Rejected, Actions Required and Other
They should only be in Other until their real status is determined.

### DIFF
--- a/test/controllers/util/SubmissionDisplayHelperSpec.scala
+++ b/test/controllers/util/SubmissionDisplayHelperSpec.scala
@@ -177,14 +177,14 @@ class SubmissionDisplayHelperSpec extends WordSpec with MustMatchers {
       }
 
       "return submissions that have missing notifications" in {
-        val submissions = Seq(submission -> Seq(notification), submission_3 -> Seq(notification_2), submission_3 -> Seq.empty)
+        val submissions = Seq(submission -> Seq(notification), submission_2 -> Seq(notification_2), submission_3 -> Seq.empty)
 
         val results = SubmissionDisplayHelper.filterSubmissions(
           submissions,
           notifications => notifications.isEmpty || notifications.headOption.map(_.status).exists(SubmissionStatus.otherStatuses.contains)
         )
 
-        results must equal(Seq(submission -> Seq(notification), Seq(submission_3 -> Seq.empty)))
+        results must equal(Seq(submission -> Seq(notification), submission_3 -> Seq.empty))
       }
     }
   }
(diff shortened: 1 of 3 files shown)


test/controllers/util/SubmissionDisplayHelperSpec.scala
```diff
@@ -166,18 +166,25 @@ class SubmissionDisplayHelperSpec extends WordSpec with MustMatchers {
...
+        val submissions = Seq(submission -> Seq(notification), submission_3 -> Seq(notification_2), submission_3 -> Seq.empty)
+
+        val results = SubmissionDisplayHelper.filterSubmissions(
+          submissions,
+          notifications => notifications.isEmpty || notifications.headOption.map(_.status).exists(SubmissionStatus.otherStatuses.contains)
+        )
 
-        SubmissionDisplayHelper.filterSubmissions(submissions, _ => true) must equal(submissions)
-        SubmissionDisplayHelper.filterSubmissions(submissions, _ => false) must equal(submissions)
+        results must equal(Seq(submission -> Seq(notification), Seq(submission_3 -> Seq.empty)))
```
MaciejRewera, line 187: Something does not seem right here... the first element is of type `Submission -> Seq[Notification]`, but the second one is `Seq[Submission -> Seq[Notification]]`. Should the second element be wrapped in `Seq`?

SteveSugden: Yep.  Test fails.  But well spotted.

SteveSugden: We can turn off PR build and rely on Maciej code reviews!  :-)

test/controllers/util/SubmissionDisplayHelperSpec.scala
```diff
@@ -166,18 +166,25 @@ class SubmissionDisplayHelperSpec extends WordSpec with MustMatchers {
...
+        val submissions = Seq(submission -> Seq(notification), submission_2 -> Seq(notification_2), submission_3 -> Seq.empty)
+
+        val results = SubmissionDisplayHelper.filterSubmissions(
+          submissions,
+          notifications => notifications.isEmpty || notifications.headOption.map(_.status).exists(SubmissionStatus.otherStatuses.contains)
+        )
 
-        SubmissionDisplayHelper.filterSubmissions(submissions, _ => true) must equal(submissions)
-        SubmissionDisplayHelper.filterSubmissions(submissions, _ => false) must equal(submissions)
+        results must equal(Seq(submission -> Seq(notification), submission_3 -> Seq.empty))
```
MaciejRewera, line 187: 👍

SteveSugden:
Interesting idea @MaciejRewera

PR #912, opened 2020-05-15
CEDS-2400 is EIDR fixes

MaciejRewera:


conf/messages.en
```diff
@@ -1056,6 +1056,12 @@ declaration.summary.references.ducr.change = DUCR
 declaration.summary.references.lrn = LRN
 declaration.summary.references.lrn.change = LRN
 declaration.summary.parties = Parties
+declaration.summary.parties.eidr = Is this an EIDR?
+declaration.summary.parties.eidr.change = Is this an EIDR?
```
PatrykRudnicki, line 1060: I forgot about this. I was wondering during my ticket why we have two message keys for the same field.

It's not something to do with this PR, this is the pattern that we're using right now, but do you know if there is any reason of this?

I think that in the future we should remove those `change` messages and use everywhere one message.

app/views/declaration/summary/parties_section.scala.html
```diff
@@ -60,30 +60,55 @@
...
+                            href = controllers.declaration.routes.PersonPresentingGoodsDetailsController.displayPage(mode).url,
+                            content = Text(messages("site.change")),
+                            visuallyHiddenText = Some(messages("declaration.summary.parties.personPresentingGoods.change"))
+                        )
+                    )
+                ))
+            )
+        })
+
+        ++ (Seq.empty :+ declarationData.parties.declarantDetails.map(declarant =>
```
PatrykRudnicki, line 103: Just style comment, `map` when is bigger and contains body in new lines should have `{ }` curly brackets. More readable.

conf/messages.en
PatrykRudnicki: ah, I see this, so the second message should start with lower case.

We have a few messages incorrect. I'm sure about my ticket.

MaciejRewera: I have no idea, looked how it is implemented so far. Don't see much value in having separate messages for that.

app/views/declaration/summary/parties_section.scala.html
MaciejRewera: I did not write this part. Just had to move it.

PR #913, opened 2020-05-18
CEDS-2298 Display Time always on British time zone

PauloPortugal:
DMS sends the times in UTC time, but our BE microservice converts them
to `LocalDateTime` which strips them of the Zone information.

Converting the type to `Instant` as we did in the the Movements service
allows us to store the date from DMS as `UTC` or `Zulu time` and then on
the FE we just need to inform the view that we want to render that time
on British Time Zone `Europe/London`.

This ensures that we will see the time as `GMT/UTC` and `BST` depending
on the time of the year.

This holds true as long as DMS returns their dates in UTC.

### DIFF
--- a/test/controllers/util/SubmissionDisplayHelperSpec.scala
+++ b/test/controllers/util/SubmissionDisplayHelperSpec.scala
@@ -16,7 +16,8 @@
 
 package controllers.util
 
-import java.time.LocalDateTime
+import java.time.{Instant, ZoneOffset, ZonedDateTime}
+import java.time.temporal.ChronoUnit.{DAYS, HOURS, MINUTES}
 import java.util.UUID
 
 import base.TestHelper.createRandomAlphanumericString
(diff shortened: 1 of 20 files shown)


PatrykRudnicki:
@PauloPortugal You have 4 tests failing

test/controllers/util/SubmissionDisplayHelperSpec.scala
```diff
@@ -217,11 +218,12 @@ object SubmissionDisplayHelperSpec {
...
   private lazy val functionCodesRandomised: Iterator[String] = Random.shuffle(functionCodes).toIterator
+
   private def randomResponseFunctionCode: String = functionCodesRandomised.next()
 
-  val dateTimeIssued: LocalDateTime = LocalDateTime.now()
-  val dateTimeIssued_2: LocalDateTime = dateTimeIssued.plusMinutes(3)
-  val dateTimeIssued_3: LocalDateTime = dateTimeIssued_2.plusMinutes(3)
+  val dateTimeIssued: Instant = Instant.now()
+  val dateTimeIssued_2: Instant = dateTimeIssued.plus(3, MINUTES);
+  val dateTimeIssued_3: Instant = dateTimeIssued_2.plus(3, MINUTES);
```
PatrykRudnicki, line 226: Please remove semicolons :) It's Scala world! :)

PauloPortugal: ohhh god!!!! No soup for me!!

test/views/NotificationsViewSpec.scala
```diff
@@ -56,6 +60,21 @@ class NotificationsViewSpec extends UnitViewSpec with Stubs {
         view mustNot containElementWithID("submission_notifications-table-row0-amend_link")
       }
 
+      "status Accepted during BST" in {
+        val view = createView(
+          submission,
```
SteveSugden, line 65: 👍 - might be clearer if the "before BST" test explicitly defined the test date/time rather than rely on the test default.

PauloPortugal: I have addressed your comment.

PR #915, opened 2020-05-18
CEDS-2378 - Move Package Information page to the two step remove pattern

SteveSugden:
There are a few things in this PR.
- validation for PackageInformation has changed.  Used to be optional for Clearance but there was a bug whereby partial PackageInfo could be sent.   Now PackageInfo is mandatory unless on Clearance journey and Procedure Code is "0019"
- PR currently makes use of a "composite" id for PackageInfo in the "remove" process - will be updated to use new id when this is available


app/controllers/declaration/CommodityMeasureController.scala
```diff
@@ -44,15 +44,9 @@ class CommodityMeasureController @Inject()(
   private def form()(implicit request: JourneyRequest[_]) = CommodityMeasure.form(request.declarationType)
 
   def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
-    val item = request.cacheModel.itemBy(itemId)
-    val packageInformation = item.map(_.packageInformation)
-    val commodityMeasure = item.flatMap(_.commodityMeasure)
-
-    (packageInformation, commodityMeasure) match {
```
SteveSugden, line 51: Could not work out any reason why PackageInfo was required in order to display CommodityMeasure.  It's been like that for a very long time.

app/controllers/declaration/CommodityDetailsController.scala
```diff
@@ -60,7 +60,10 @@ class CommodityDetailsController @Inject()(
 
   private def redirectToNextPage(mode: Mode, itemId: String)(implicit request: JourneyRequest[AnyContent]): Result =
     if (request.isType(DeclarationType.CLEARANCE)) {
-      navigator.continueTo(mode, controllers.declaration.routes.PackageInformationController.displayPage(_, itemId))
+      if (request.cacheModel.itemBy(itemId).exists(_.isExportInventoryCleansingRecord))
+        navigator.continueTo(mode, controllers.declaration.routes.CommodityMeasureController.displayPage(_, itemId))
```
SteveSugden, line 64: Skip PackageInfo if Clearance and Proc Code is "0019"

app/controllers/declaration/PackageInformationRemoveController.scala
```diff
@@ -0,0 +1,75 @@
...
+  journeyType: JourneyAction,
+  override val exportsCacheService: ExportsCacheService,
+  navigator: Navigator,
+  mcc: MessagesControllerComponents,
+  packageTypeRemove: package_information_remove
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable {
+
+  def displayPage(mode: Mode, itemId: String, id: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    Ok(packageTypeRemove(mode, itemId, PackageInformation.fromId(id), YesNoAnswer.form()))
```
SteveSugden, line 47: Will need to change to get PackageInfo from cache based on real id

app/controllers/declaration/ProcedureCodesController.scala
```diff
@@ -126,15 +126,25 @@ class ProcedureCodesController @Inject()(
   private def updateCache(itemId: String, updatedProcedureCodes: ProcedureCodesData)(
     implicit r: JourneyRequest[AnyContent]
   ): Future[Option[ExportsDeclaration]] = {
+
+    def clearDataForProcedureCode(code: String, itemId: String, model: ExportsDeclaration) = {
+      var result = model
```
SteveSugden, line 131: There are now 2 Proc Codes that imply that we need to clear possible invalid data
- fiscal info when code !- 1042
- package info when Clearance and code is 0019

There may me a better way to do this without a var.

app/controllers/navigation/Navigator.scala
```diff
@@ -282,17 +291,31 @@ object Navigator {
...
     case ConsigneeDetails    => consigneeDetailsClearancePreviousPage
     case RepresentativeAgent => representativeAgentClearancePreviousPage
   }
 
+  val clearanceCacheItemDependent: PartialFunction[DeclarationPage, (ExportsDeclaration, Mode, String) => Call] = {
+    case CommodityMeasure => commodityMeasureClearancePreviousPage
+  }
+
+  private def commodityMeasureClearancePreviousPage(cacheModel: ExportsDeclaration, mode: Mode, itemId: String): Call =
+    if (cacheModel.itemBy(itemId).exists(_.isExportInventoryCleansingRecord))
```
SteveSugden, line 314: Back from CommodityMeasure depends on value of Proc Code

app/services/model/PackageType.scala
```diff
@@ -18,5 +18,5 @@ package services.model
 
 case class PackageType(code: String, description: String) {
 
-  def asText(): String = s"$description - $code"
+  def asText(): String = s"$description ($code)"
```
SteveSugden, line 21: Changed in line with latest prototype - both for drop-down and in summary tables

app/controllers/declaration/ProcedureCodesController.scala
PatrykRudnicki: Yeah, this method should be rewrited. I'll try to look at this in the IntelliJ.

app/controllers/navigation/Navigator.scala
```diff
@@ -366,12 +389,15 @@ object Navigator {
...
-          case CLEARANCE     => clearanceItemPage
+          case STANDARD      => standardCacheItemDependent.orElse(standardItemPage)
+          case SUPPLEMENTARY => supplementaryCacheItemDependent.orElse(supplementaryItemPage)
+          case SIMPLIFIED    => simplifiedCacheItemDependent.orElse(simplifiedItemPage)
+          case OCCASIONAL    => occasionalCacheItemDependent.orElse(occasionalItemPage)
+          case CLEARANCE     => clearanceCacheItemDependent.orElse(clearanceItemPage)
+        }
+        commonCacheItemDependent.orElse(commonItem).orElse(specific)(page) match {
+          case mapping: ((Mode, String) => Call)                     => mapping(mode, itemId.id)
+          case mapping: ((ExportsDeclaration, Mode, String) => Call) => mapping(request.cacheModel, mode, itemId.id)
```
PatrykRudnicki, line 400: This has been added by Maciej. Is your branch rebased/updated?

SteveSugden: Not since Friday.  I'll take a look

app/controllers/declaration/ProcedureCodesController.scala
PatrykRudnicki: Proposal, but to be honest I don't have any ideas how to make it more readable a lot of conditional logic...

Some improvements, probably can be better to use
 instead of:
`r.declarationType == DeclarationType.CLEARANCE`
use
`request.isType(CLEARANCE)`

And change `exportInventoryCleansingRecordProcedureCodes` to something shorter. Whit this probably it will be slightly more readable.

```
def clearDataForProcedureCode(code: String, itemId: String, model: ExportsDeclaration): ExportsDeclaration = {
      
      val decWithUpdatedFiscalReferences =
        if (!ProcedureCodesData.osrProcedureCodes.contains(code))
          model.updatedItem(itemId, item => item.copy(fiscalInformation = None, additionalFiscalReferencesData = None))
        else model
      
      if (r.declarationType == DeclarationType.CLEARANCE && ProcedureCodesData.exportInventoryCleansingRecordProcedureCodes.contains(code))
        decWithUpdatedFiscalReferences.updatedItem(itemId, item => item.copy(packageInformation = None))
      else decWithUpdatedFiscalReferences
    }
```

SteveSugden: So basically passing each (potentially) modified model to the next section/test?   I'm not sure that is more readable than a mutable model (i.e. the var).

app/controllers/navigation/Navigator.scala
SteveSugden: As per Slack discussion - Maciej did non-item pages.  My changes add support for item pages.

app/controllers/declaration/ProcedureCodesController.scala
```diff
@@ -126,15 +126,27 @@ class ProcedureCodesController @Inject()(
...
+          model.updatedItem(itemId, item => item.copy(fiscalInformation = None, additionalFiscalReferencesData = None))
+        else model
+
+      val step2Model =
+        if (r.isType(DeclarationType.CLEARANCE) && ProcedureCodesData.eicrProcedureCodes.contains(code))
+          step1Model.updatedItem(itemId, item => item.copy(packageInformation = None))
+        else step1Model
+
+      step2Model
+    }
```
PauloPortugal, line 142: Could you explain these two `stepModels` please?

PatrykRudnicki: That was to avoid using var. There are two changes on the models but sequentional.

I'm not sure if without var it's better.

Background you can find in one of my comments above

app/controllers/declaration/ProcedureCodesController.scala
PatrykRudnicki: @PauloPortugal Check this

app/controllers/declaration/ProcedureCodesController.scala
SteveSugden: They are a result of a suggestion by Patryk.    Basically this method needs to make (currently) 2 (but will be more in future) passes of the ExportDeclaration model to potentially remove data that is not required due to the current procedure code.

I originally had a var holding the current state of the model which may or may not be updated by each step.

In this version, each step takes the output from the previous step.

app/controllers/declaration/PackageInformationRemoveController.scala
```diff
@@ -0,0 +1,79 @@
...
+      .bindFromRequest()
+      .fold(
+        (formWithErrors: Form[YesNoAnswer]) =>
+          Future.successful(BadRequest(packageTypeRemove(mode, itemId, packageInformationToRemove, formWithErrors))),
+        formData => {
+          formData.answer match {
+            case YesNoAnswers.yes =>
+              updateExportsCache(itemId, packageInformationToRemove)
+                .map(_ => navigator.continueTo(mode, routes.PackageInformationSummaryController.displayPage(_, itemId)))
+            case YesNoAnswers.no =>
```
MaciejRewera, line 62: Is it intentional not to store the users response if it's "no" ?

app/controllers/declaration/ProcedureCodesController.scala
SteveSugden: I numbered the models to hopefully make it more obvious that each step is working on the previous steps output.  But obviously not that obvious.

app/controllers/declaration/PackageInformationRemoveController.scala
SteveSugden: Yes.  Why would you store the users decision not to remove a list item each time?

app/controllers/declaration/ProcedureCodesController.scala
PauloPortugal: Tech sorting a business issue.
The Questions should be reverted then...

Thank you.

app/controllers/declaration/PackageInformationSummaryController.scala
```diff
@@ -0,0 +1,81 @@
...
+            case YesNoAnswers.no  => navigator.continueTo(mode, nextPage(itemId))
+        }
+      )
+  }
+
+}
+
+object PackageInformationSummaryController {
+
+  def nextPage(itemId: String)(implicit request: JourneyRequest[_]): Mode => Call =
```
MaciejRewera, line 74: Nit: Unless there is a reason for this, I don't think that introducing a new "pattern" here is necessary. In all other places we place this method in the class body.

SteveSugden: I think I copied that pattern from the NactCodeSummaryController.  In that controller the "nextPage" logic was used by more than one controller.   In this case it isn't, so you are right.

SteveSugden: I can move it if you like

app/controllers/declaration/ProcedureCodesController.scala
```diff
@@ -126,15 +126,27 @@ class ProcedureCodesController @Inject()(
...
+        if (!ProcedureCodesData.osrProcedureCodes.contains(code))
+          model.updatedItem(itemId, item => item.copy(fiscalInformation = None, additionalFiscalReferencesData = None))
+        else model
+
+      val step2Model =
+        if (r.isType(DeclarationType.CLEARANCE) && ProcedureCodesData.eicrProcedureCodes.contains(code))
+          step1Model.updatedItem(itemId, item => item.copy(packageInformation = None))
+        else step1Model
+
+      step2Model
```
MaciejRewera, line 141: I'd change this into a single "if" statement with 3 possible outcomes. Something like this:
`if (r.isType(DeclarationType.CLEARANCE) && ProcedureCodesData.eicrProcedureCodes.contains(code))
model.updatedItem(itemId, item => item.copy(fiscalInformation = None, additionalFiscalReferencesData = None, packageInformation = None))
else if (!ProcedureCodesData.osrProcedureCodes.contains(code))
model.updatedItem(itemId, item => item.copy(fiscalInformation = None, additionalFiscalReferencesData = None))
else
model`
It took me a while to figure out what and why is it done in two steps.

PatrykRudnicki: but the problem here is that this must be in separate ifs. There can be a situation when two updates must be done in the same method

app/controllers/navigation/Navigator.scala
```diff
@@ -258,41 +262,67 @@ object Navigator {
     case TaricCode                 => controllers.declaration.routes.TaricCodeSummaryController.displayPage
     case TaricCodeFirst            => controllers.declaration.routes.CusCodeController.displayPage
     case StatisticalValue          => controllers.declaration.routes.NactCodeSummaryController.displayPage
-    case CommodityMeasure          => controllers.declaration.routes.PackageInformationController.displayPage
-    case DocumentsProduced         => controllers.declaration.routes.AdditionalInformationController.displayPage
   }
 
   val commonCacheDependent: PartialFunction[DeclarationPage, (ExportsDeclaration, Mode) => Call] = Map.empty
 
+  val commonCacheItemDependent: PartialFunction[DeclarationPage, (ExportsDeclaration, Mode, String) => Call] = {
```
MaciejRewera, line 269: Nit: Those should be rather `commonItemCacheDependent`

app/controllers/declaration/ProcedureCodesController.scala
SteveSugden: That approach would fail to make both changes.  In other words, it is possible for both of the "if" statements to be true.  I.e. a code of "0019" is both equal to "0019" and not equal to "1042".
There is a reason for this mess.

app/controllers/declaration/ProcedureCodesController.scala
PatrykRudnicki: I thought about the one approach, but not sure if it will be suitable.

Create a methods to update declaration but which will return Future[ExportsDeclaration], e.g.

```
def updateProcedureCodes(model: ExportDeclaration): Future[ExportsDeclaration] =
  Future.successful(
    if(condition) updateModel
    else model
 )
```

Create a method for every update. and combine it like

`updateProcedurecodes(model).andThen(updateFiscalInformation(_))`

or something like this.

Probably this approach will work somehow, but it will introduce the unnecessary Futures to work on it. It's not a huge deal, because using Future.successful will give us a 100% confidence that it's not something that we need to wait for the result.

PatrykRudnicki: But with this we will have a different signature for every method. Not return ExportsDeclaration, but `Future[ExportsDeclaration]` which can be more problematic and cause more issues.

I think this is really good to some tech ticket or just tech discussion. How to solve that problem.

app/controllers/navigation/Navigator.scala
SteveSugden: Depends how you read it in your head.  "common cache and item dependant" or "common item and cache dependant"?  Both make sense.

app/forms/declaration/PackageInformation.scala
```diff
@@ -47,48 +54,31 @@ object PackageInformation extends DeclarationPage {
   private val NumberOfPackagesLimitLower = 0
   private val NumberOfPackagesLimitUpper = 99999
 
-  private val mappingTypesOfPackagesOptional: Mapping[Option[String]] = optional(
-    text()
-      .verifying("declaration.packageInformation.typesOfPackages.error", isContainedIn(PackageTypes.all.map(_.code)))
-  )
+  private def generateId: String = Random.alphanumeric.take(8).mkString.toLowerCase
```
MaciejRewera, line 57: This is inconsistent with the BE way of creating Id. It is using the same logic as `ExportItemIdGeneratorService` which is not equal to what is here.

SteveSugden: True.  But what difference does it make?  A random string is a random string.

docs/Customs Declare Exports AutoComplete.js
```diff
@@ -710,16 +710,14 @@ function statisticalValue(){
...
-            document.getElementById('add').click();
-        }
+        selectFromAutoPredict(document.getElementById('typesOfPackages-container'), "XD");
+        document.getElementById('numberOfPackages').value ='10';
+        document.getElementById('shippingMarks').value = 'Shipping description';
+        document.getElementById('submit').click();
+    }
+    if (currentPageIs('/customs-declare-exports/declaration/items/.*/packages-list')) {
+        document.getElementById('code_no').checked = 'checked';
+        document.getElementById('submit').click();
```
MaciejRewera, line 720: 👍

test/models/declaration/ExportDeclarationTestData.scala
```diff
@@ -117,7 +117,7 @@ object ExportDeclarationTestData {
           Some(AdditionalFiscalReferencesData(Seq(AdditionalFiscalReference("PL", "12345"), AdditionalFiscalReference("FR", "54321")))),
         procedureCodes = Some(ProcedureCodesData(Some("CUPR"), Seq("CC", "PR"))),
         packageInformation =
-          Some(List(PackageInformation(Some("AA"), Some(2), Some("mark1")), PackageInformation(Some("AB"), Some(4), Some("mark2"))))
+          Some(List(PackageInformation("pkgAA", Some("AA"), Some(2), Some("mark1")), PackageInformation("pkgBB", Some("AB"), Some(4), Some("mark2"))))
```
MaciejRewera, line 120: Nit: This seems not to be updated after introducing non-composite ids

test/services/GoodsItemCachingData.scala
```diff
@@ -31,7 +31,12 @@ trait GoodsItemCachingData {
   def decimalString(): String = Random.nextDouble().toString
 
   def createPackageInformation(): PackageInformation =
-    PackageInformation(Some(createRandomAlphanumericString(2)), Some(Random.nextInt(20) + 1), Some(createRandomAlphanumericString(150)))
+    PackageInformation(
+      createRandomAlphanumericString(5),
```
MaciejRewera, line 35: Nit: This is also not adequate to the new implementation.

test/services/cache/ExportsItemBuilder.scala
```diff
@@ -103,7 +102,12 @@ trait ExportsItemBuilder {
     cache =>
       cache.copy(
         packageInformation = Some(
-          cache.packageInformation.getOrElse(List.empty) :+ PackageInformation(Some(typesOfPackages), Some(numberOfPackages), Some(shippingMarks))
+          cache.packageInformation.getOrElse(List.empty) :+ PackageInformation(
+            UUID.randomUUID().toString,
```
MaciejRewera, line 106: Nit: And here it's inconsistent. I am aware that we can put there anything, but if you consider tests as some kind of a documentation of how should the code work, then consistency here is recommended in my opinion.

test/services/GoodsItemCachingData.scala
SteveSugden: In what way is it "inadequate"?    The id is a string.

test/unit/controllers/declaration/CommodityMeasureControllerSpec.scala
```diff
@@ -45,7 +45,7 @@ class CommodityMeasureControllerSpec extends ControllerSpec {
     authorizedUser()
 
     val item =
-      ExportItem("itemId", packageInformation = Some(List(PackageInformation(Some("123"), Some(123), Some("123")))))
+      ExportItem("itemId", packageInformation = Some(List(PackageInformation("id", Some("123"), Some(123), Some("123")))))
```
MaciejRewera, line 48: Nit: If it makes a difference, `"id"` is not consistent here.

test/services/cache/ExportsItemBuilder.scala
SteveSugden: Nonsense.  You test to the API not the implementation.

test/unit/forms/declaration/PackageInformationSpec.scala
```diff
@@ -41,29 +37,19 @@ class PackageInformationSpec extends UnitViewSpec {
     }
 
     "has correct type of package text" in {
-      val model = PackageInformation(Some("PK"), Some(10), Some("marks"))
+      val model = PackageInformation("id", Some("PK"), Some(10), Some("marks"))
```
MaciejRewera, line 40: Nit: Probabluy it makes no difference here, but this id is not representative

test/views/declaration/ItemSummaryViewSpec.scala
```diff
@@ -89,15 +89,15 @@ class ItemSummaryViewSpec extends UnitViewSpec with ExportsTestData with Stubs w
               procedureCodes = Some(ProcedureCodesData(Some("procedure-code1"), Seq.empty)),
               statisticalValue = Some(StatisticalValue("")),
               commodityDetails = Some(CommodityDetails(Some("item-type1"), Some(""))),
-              packageInformation = Some(List(PackageInformation(None, Some(1), None)))
+              packageInformation = Some(List(PackageInformation("pk1", None, Some(1), None)))
```
MaciejRewera, line 92: Nit: Smae here about the id

test/views/declaration/PackageInformationAddViewSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+import views.declaration.spec.UnitViewSpec
+import views.html.declaration.package_information_add
+import views.tags.ViewTest
+
+@ViewTest
+class PackageInformationAddViewSpec extends UnitViewSpec with ExportsTestData with Stubs with MustMatchers with CommonMessages with Injector {
+
+  import PackageInformationViewSpec._
+
+  private val itemId = "item1"
```
MaciejRewera, line 39: Nit: Same here about the id

test/views/declaration/PackageInformationViewSpec.scala
```diff
@@ -58,183 +58,106 @@ class PackageInformationViewSpec extends UnitViewSpec with ExportsTestData with
...
-
-      view.select(".govuk-table__body > tr:nth-child(2) > td:nth-child(1)").text() mustBe "PB"
-      view.select(".govuk-table__body > tr:nth-child(2) > td:nth-child(2)").text() mustBe "101"
-      view.select(".govuk-table__body > tr:nth-child(2) > td:nth-child(3)").text() mustBe "Shipping Mark"
-    }
   }
 }
+
+object PackageInformationViewSpec {
+  val packageInformation: PackageInformation = PackageInformation("ID", Some("1A"), Some(1), Some("Marks"))
```
MaciejRewera, line 162: Nit: Same here about the id

test/views/declaration/commodityMeasure/CommodityMeasureViewSpec.scala
```diff
@@ -98,7 +99,7 @@ class CommodityMeasureViewSpec extends UnitViewSpec with CommonMessages with Stu
         val backButton = createView().getElementById("back-link")
 
         backButton.text() mustBe messages(backCaption)
-        backButton.attr("href") must endWith(s"/items/$itemId/package-information")
+        backButton.attr("href") mustBe controllers.declaration.routes.PackageInformationSummaryController.displayPage(Mode.Normal, itemId).url
```
MaciejRewera, line 102: 👍

test/views/declaration/summary/PackageInformationViewSpec.scala
```diff
@@ -38,12 +38,15 @@ class PackageInformationViewSpec extends UnitViewSpec with ExportsTestData with
...
-      row must haveSummaryActionsHref(controllers.declaration.routes.PackageInformationController.displayPage(Mode.Normal, "itemId"))
+      row must haveSummaryActionsHref(controllers.declaration.routes.PackageInformationSummaryController.displayPage(Mode.Normal, "itemId"))
     }
 
     "display package information section with multiple package information and change buttons" in {
 
-      val data = Seq(PackageInformation(Some("PB"), Some(123), Some("first-marks")), PackageInformation(Some("QF"), Some(321), Some("second-marks")))
+      val data = Seq(
+        PackageInformation("ID1", Some("PB"), Some(123), Some("first-marks")),
+        PackageInformation("ID2", Some("QF"), Some(321), Some("second-marks"))
```
MaciejRewera, line 48: Nit: Same here about the id

app/controllers/declaration/PackageInformationRemoveController.scala
MaciejRewera: That's  true. Did not look at the class name.

app/controllers/declaration/ProcedureCodesController.scala
MaciejRewera: Ok, just read the long discussion under Paulo's comment. I agree that we can think about some other way of pipelining operations in such scenario.

app/controllers/navigation/Navigator.scala
MaciejRewera: Good point! I was just adding "ItemDependent" to the value names that already existed.

test/services/GoodsItemCachingData.scala
MaciejRewera: Sorry, by inadequate I mean that the id is created in a different way it is generated in the application code.

PR #916, opened 2020-05-19
CEDS-2420 POF displaying error on the page from rejected notifications

PatrykRudnicki:
To simulate this behaviour locally please make a REJECTED declaration and inside notifications collection in db override errors part with:
```
"errors" : [ 
        {
            "validationCode" : "CDS40045",
            "pointer" : "declaration.items.#1.documentProduced"
        }
    ]
```

This will add you a notification with correct pointer which is supported right now


app/controllers/declaration/DocumentsProducedController.scala
```diff
@@ -50,9 +48,17 @@ class DocumentsProducedController @Inject()(
     extends FrontendController(mcc) with I18nSupport with ModelCacheable {
 
   def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    val fieldName = request.flash.get("fieldName")
+    val errorMessage = request.flash.get("errorMessage")
+
+    val errors = (fieldName, errorMessage) match {
+      case (nameOpt, Some(messageKey)) => Seq(FormError(nameOpt.getOrElse(""), messageKey))
+      case _                           => Seq.empty
+    }
```
PatrykRudnicki, line 57: I think this logic can live in the journeyType action and information can be sticked with request. Still we will need to build the form errors here, but the logic about retrieving it etc can be hidden.

PatrykRudnicki: Additionally we can build some helper method there that for every form (`Form[A]` will add errors).

PR #918, opened 2020-05-20
CEDS-2408 Skip CarrierDetails on Clearance journey if it's not EXS

MaciejRewera:


MaciejRewera:
The back button has been already implemented.

PR #919, opened 2020-05-20
CEDS 1881 - User cannot enter data in lowercase letters

SteveSugden:


app/forms/Ducr.scala
```diff
@@ -26,10 +26,12 @@ case class Ducr(ducr: String)
 object Ducr {
   implicit val format = Json.format[Ducr]
 
-  private val ducrFormat = """^[0-9][A-Z][A-Z][0-9A-Z\(\)\-/]{6,32}"""
+  private val ducrFormat = """^[0-9][A-Za-z][A-Za-z][0-9A-Za-z\(\)\-/]{6,32}"""
+
+  def form2Data(ducr: String): Ducr = new Ducr(ducr.toUpperCase)
```
PatrykRudnicki, line 31: What do you think about putting this as apply method?

app/forms/declaration/WarehouseIdentification.scala
```diff
@@ -38,7 +38,7 @@ object WarehouseIdentification extends DeclarationPage {
   private def form2Model: (String, Option[String]) => WarehouseIdentification = {
     case (inWarehouse, warehouseId) =>
       inWarehouse match {
-        case yes => WarehouseIdentification(warehouseId)
+        case yes => WarehouseIdentification(warehouseId.map(_.toUpperCase))
```
PatrykRudnicki, line 41: This logic could be in apply method too.

app/forms/Ducr.scala
SteveSugden: I think I tried that once (when doing the PackageInfo changes).  The mapping didn't seem to like 2 apply methods.   

Error:(27, 36) ambiguous reference to overloaded definition,
both method apply in object Ducr of type (ducr: String)forms.Ducr
and  method apply in object Ducr of type (ducr: String)forms.Ducr
match expected type String => ?
  implicit val format = Json.format[Ducr]

SteveSugden: Ah, I see it's the json format method that is un-happy

PatrykRudnicki: yeah, but I'm talking about `override def apply(ducr: String): Ducr = new Ducr(ducr.toUpperCase)`.

This will affect all data always, but in general we want to have always uppercase. If you override the apply method this ambigousity should disappear

SteveSugden: "ambigousity"?   I'm stealing that.

PatrykRudnicki: Let's leave it how it was, override won't work

PR #921, opened 2020-05-21
CEDS-2416 - Move the Previous Documents Page into the GDS

SteveSugden:


app/views/components/gds/exportsInputText.scala.html
```diff
@@ -45,5 +45,5 @@
     label = buildLabel,
     hint = hintKey.map(key => Hint(content = Text(messages(key)))),
     errorMessage = field.error.map(err => ErrorMessage(content = Text(messages(err.message)))),
-    classes = defaultInputClasses + inputClasses.map(clazz => s" $clazz").getOrElse("")
+    classes = inputClasses.map(clazz => s" $clazz").getOrElse(defaultInputClasses)
```
PatrykRudnicki, line 48: 👍

app/views/declaration/goods_measure.scala.html
```diff
@@ -1,74 +0,0 @@
...
-
-@this(main_template: views.html.main_template)
-
-@(mode: Mode, itemId: String, form: Form[CommodityMeasure])(implicit request: JourneyRequest[_], messages: Messages)
-
-@tariffBodyDesc = { @Html(messages("declaration.type.consignmentTariffText", components.details_content_link())) }
-
-@tariffBody = { @components.tariff_body_partial(messages("declaration.type.commodityMeasureListItemOne"), messages("declaration.type.commodityMeasureListItemTwo"), messages("declaration.type.commodityMeasureListItemThree"))(tariffBodyDesc)}
-
-@main_template(
```
PatrykRudnicki, line 32: Is it some unused page?

PatrykRudnicki: Probably we should be able in that case clear the messages that are not used after removing the file

SteveSugden: Yes, it's not used.  I was looking for pages that were still not using GDS.  Good point about the messages.  I'll do that now.

PatrykRudnicki: I should read the commit messages to not ask stupid questions 😂

SteveSugden: It's early.  And Friday!

SteveSugden: Just checked - there are no messages referenced in this file that aren't either being used elsewhere or not in the messages.en file (i.e. already purged)

PR #923, opened 2020-05-22
CEDS-2450 Remove PersonPresentingGoodsDetails if answered 'no' to EIDR

MaciejRewera:
This is not to keep data that is unnecessary and potentially breaking
for downstream logic.

### DIFF
--- a/app/controllers/declaration/EntryIntoDeclarantsRecordsController.scala
+++ b/app/controllers/declaration/EntryIntoDeclarantsRecordsController.scala
@@ -60,7 +60,14 @@ class EntryIntoDeclarantsRecordsController @Inject()(
   }
 
   private def updateCache(validData: YesNoAnswer)(implicit request: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] =
-    updateExportsDeclarationSyncDirect(model => model.copy(parties = model.parties.copy(isEntryIntoDeclarantsRecords = Some(validData))))
+    validData.answer match {
+      case YesNoAnswers.yes =>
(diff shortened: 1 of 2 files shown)


app/controllers/declaration/EntryIntoDeclarantsRecordsController.scala
```diff
@@ -60,7 +60,9 @@ class EntryIntoDeclarantsRecordsController @Inject()(
   }
 
   private def updateCache(validData: YesNoAnswer)(implicit request: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] =
-    updateExportsDeclarationSyncDirect(model => model.copy(parties = model.parties.copy(isEntryIntoDeclarantsRecords = Some(validData))))
+    updateExportsDeclarationSyncDirect(
+      model => model.copy(parties = model.parties.copy(isEntryIntoDeclarantsRecords = Some(validData), personPresentingGoodsDetails = None))
```
PatrykRudnicki, line 64: If user come back with back button and choose Yes once again he will loose the data.

I think this putting None inside should be only when User choose No

SteveSugden: Good job Patryk is on the ball.  I thought that was the case based on the new "user said No" test.

PatrykRudnicki: The test case is for No, but as I see logic in the controller will always invoke this method, so for Yes we will remove it.

SteveSugden: I agree with you.

PR #928, opened 2020-05-26
CEDS-2460 - swap change and remove seals links on container page

SteveSugden:


conf/messages.en
```diff
@@ -799,9 +799,9 @@ declaration.transportInformation.containers.add = Do you need to add another con
 declaration.transportInformation.containers.add.title = What is the container ID?
 declaration.transportInformation.containerId.title = Container ID
 declaration.transportInformation.containers.remove = Remove container
-declaration.transportInformation.containers.remove.hint = Remove container {0}
-declaration.transportInformation.containers.change = Change seals
-declaration.transportInformation.containers.change.hint = Change seals for container {0}
+declaration.transportInformation.containers.remove.hint = container {0}
+declaration.transportInformation.containers.change = Change
+declaration.transportInformation.containers.change.hint = seals for container {0}
```
MaciejRewera, line 804: Does it make a difference that these start with lower letter?

SteveSugden: The difference was to remove the word 'Change' from the visually hidden hint text.  So that the screen reader doesn't announce "Change Change seals...."

PR #936, opened 2020-05-28
CEDS-2454 - content change for confirmation page

SteveSugden:


SteveSugden:
Could you take another look please @PatrykRudnicki

PR #939, opened 2020-05-29
CEDS-2455 - Move the 'wrong Enrollment' page to GDS

SteveSugden:


app/config/AppConfig.scala
```diff
@@ -62,6 +59,10 @@ class AppConfig @Inject()(
   lazy val classificationHelpUrl = loadConfig("urls.classificationHelp")
   lazy val ecicsToolUrl = loadConfig("urls.ecicsTool")
 
+  lazy val eoriService = loadConfig("urls.eoriService")
```
SteveSugden, line 62: Should I move these (and other "static" urls.xxx values) into it's own "config" class?  E.g. ExternalLinksConfig.

PatrykRudnicki: I think that can be a good idea, but it's up to you :)

test/views/UnauthorisedViewSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+@ViewTest
+class UnauthorisedViewSpec extends UnitViewSpec with Stubs with Injector {
+
+  private val appConfig = mock[AppConfig]
+  when(appConfig.customsDeclarationsGoodsTakenOutOfEuUrl)
+    .thenReturn("appConfig.customsDeclarationsGoodsTakenOutOfEuUrl")
+  when(appConfig.commodityCodesUrl).thenReturn("appConfig.commodityCodesUrl")
+  when(appConfig.relevantLicensesUrl).thenReturn("appConfig.relevantLicensesUrl")
+  when(appConfig.serviceAvailabilityUrl).thenReturn("appConfig.serviceAvailabilityUrl")
+  when(appConfig.customsMovementsFrontendUrl).thenReturn("appConfig.customsMovementsFrontendUrl")
```
PatrykRudnicki, line 37: This `appConfig` is unnecessary.

You're not using it anywhere so this part can be removed :)

SteveSugden: Yes.  I copied this from another view spec.  Good spot.

PR #940, opened 2020-06-01
CEDS-2388 Added support for more errors plus fixes for redirection

PatrykRudnicki:
Multiple items like routing countries require custom redirection with ErrorFix mode, because the journey need to behave like a normal journey for few pages.
This is the reason that I used the customs redirect without using navigation for ErrorFix mode.


app/controllers/declaration/RoutingCountriesSummaryController.scala
```diff
@@ -93,14 +94,18 @@ class RoutingCountriesSummaryController @Inject()(
...
+          if (validAnswer) removeCountry(Country(Some(countryCode))).map(_ => removeRedirect(mode))
+          else Future.successful(removeRedirect(mode))
       )
   }
 
+  private def removeRedirect(mode: Mode)(implicit request: JourneyRequest[AnyContent]): Result =
+    if (mode == ErrorFix) {
+      Redirect(controllers.declaration.routes.RoutingCountriesSummaryController.displayPage(mode))
+    } else {
+      navigator.continueTo(mode, controllers.declaration.routes.RoutingCountriesSummaryController.displayPage)
```
SteveSugden, line 106: This is the 2nd or 3rd time you've done this.  On the face of it it looks like both paths lead to the same re-direct (and so the code os hard to understand).   Can't this logic be added to Navigator.continueTo?

PatrykRudnicki:
Yes, I'll prepare this PR tomorrow

PR #943, opened 2020-06-02
BAU Refactor for ErrorFix navigator

PatrykRudnicki:


app/controllers/navigation/Navigator.scala
```diff
@@ -40,12 +40,12 @@ import uk.gov.hmrc.http.HeaderCarrier
 
 class Navigator @Inject()(appConfig: AppConfig, auditService: AuditService) {
 
-  def continueTo(mode: Mode, factory: Mode => Call)(implicit req: JourneyRequest[AnyContent], hc: HeaderCarrier): Result =
+  def continueTo(mode: Mode, factory: Mode => Call, isErrorFixInProgress: Boolean = false)(
```
MaciejRewera, line 43: Would it make sense to have this `ErrorFixInProgress` as a new `Mode` ?

app/controllers/declaration/RoutingCountriesSummaryController.scala
```diff
@@ -70,12 +70,12 @@ class RoutingCountriesSummaryController @Inject()(
       )
   }
 
+  private def isModeErrorFix(mode: Mode): Boolean = mode == ErrorFix
```
SteveSugden, line 73: Could you not add a method (isErrorFix) to Mode?

app/controllers/declaration/RoutingCountriesSummaryController.scala
```diff
@@ -70,12 +70,12 @@ class RoutingCountriesSummaryController @Inject()(
...
+
   private def redirectFromSummaryPage(mode: Mode, answer: Boolean)(implicit request: JourneyRequest[AnyContent]): Result =
-    mode match {
-      case ErrorFix if answer => Redirect(controllers.declaration.routes.RoutingCountriesController.displayRoutingCountry(mode))
-      case _ if answer        => navigator.continueTo(mode, controllers.declaration.routes.RoutingCountriesController.displayRoutingCountry)
-      case _                  => navigator.continueTo(mode, controllers.declaration.routes.LocationController.displayPage)
-    }
+    if (answer)
+      navigator.continueTo(mode, controllers.declaration.routes.RoutingCountriesController.displayRoutingCountry, isModeErrorFix(mode))
+    else navigator.continueTo(mode, controllers.declaration.routes.LocationController.displayPage)
```
SteveSugden, line 78: i.e. `navigator.continueTo(mode, controllers.declaration.routes.RoutingCountriesController.displayRoutingCountry, mode.isErrorFix)`

app/controllers/navigation/Navigator.scala
MaciejRewera: In my opinion the last parameter is not necessary. In every place you pass it, you are doing the same comparision. Why not doing it inside `continueTo` method? You have access to the `mode` field so you can preserve the logic, but you don't clutter method's interface.
Unless there are some other non-obvious factors that I am not aware of.

PatrykRudnicki: Hmm, that can be a solution too, but with new mode you'll need to have this logic inside the controller, so I'll come back to previous solution

app/controllers/declaration/RoutingCountriesSummaryController.scala
PatrykRudnicki: yeah, I thought about it. I'll do it

app/controllers/navigation/Navigator.scala
PatrykRudnicki: I agree with second comment, but I think that it won't handle the logic as it should. I'll check it right now

app/controllers/declaration/RoutingCountriesController.scala
```diff
@@ -79,11 +79,10 @@ class RoutingCountriesController @Inject()(
...
   private def redirectFromRoutingAnswer(mode: Mode, answer: Boolean)(implicit request: JourneyRequest[AnyContent]): Result =
-    mode match {
-      case ErrorFix if answer => Redirect(controllers.declaration.routes.RoutingCountriesController.displayRoutingCountry(mode))
-      case _ if answer        => navigator.continueTo(mode, controllers.declaration.routes.RoutingCountriesController.displayRoutingCountry)
-      case _                  => navigator.continueTo(mode, controllers.declaration.routes.LocationController.displayPage)
-    }
+    if (answer)
+      navigator.continueTo(mode, controllers.declaration.routes.RoutingCountriesController.displayRoutingCountry, mode == ErrorFix)
+    else
+      navigator.continueTo(mode, controllers.declaration.routes.LocationController.displayPage)
```
SteveSugden, line 85: The new parameter is required for this situation isn't it?  Here the mode might be ErrorFix but we are not passing isErrorFixInProgress=true (I presume that is right)

app/controllers/navigation/Navigator.scala
PatrykRudnicki: Yeah, we need only pass this elemnt in scenario where we're considering different behaviour than normal one. This is the reason that this can't leave inside the Navigator

app/controllers/declaration/RoutingCountriesController.scala
PatrykRudnicki: It's not required, because if the answer is no we always will return to the Rejected notification page.

app/controllers/declaration/RoutingCountriesSummaryController.scala
PatrykRudnicki: Fixed

app/controllers/declaration/RoutingCountriesSummaryController.scala
PatrykRudnicki: Fixed

app/controllers/navigation/Navigator.scala
PatrykRudnicki: I think new mdoe won't help, still there need to be logic for it :(

PR #944, opened 2020-06-02
CEDS-2467 - Show error messages after navigating from error page

SteveSugden:
Not as bad as it looks - lots of repeated changes to the declaration controllers.
Only updated controller specs that were already capturing the mock form (to test that submission errors are added)

### DIFF
--- a/app/controllers/SubmissionsController.scala
+++ b/app/controllers/SubmissionsController.scala
@@ -21,11 +21,12 @@ import connectors.exchange.ExportsDeclarationExchange
 import controllers.actions.AuthAction
 import controllers.util.SubmissionDisplayHelper
 import javax.inject.Inject
-import models.{ExportsDeclaration, Mode}
 import models.Mode.ErrorFix
 import models.requests.{AuthenticatedRequest, ExportsSessionKeys}
+import models.responses.FlashKeys.errorMessage
+import models.{ExportsDeclaration, Mode}
 import play.api.i18n.I18nSupport
-import play.api.mvc.{Action, AnyContent, MessagesControllerComponents, Result}
+import play.api.mvc._
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import views.html.{declaration_information, submissions}
 
@@ -73,9 +74,10 @@ class SubmissionsController @Inject()(
     }
   }
 
-  def amendErrors(id: String, redirectUrl: String): Action[AnyContent] = authenticate.async { implicit request =>
+  def amendErrors(id: String, redirectUrl: String, pattern: String, messageKey: String): Action[AnyContent] = authenticate.async { implicit request =>
     val redirectUrlWithMode = redirectUrl + ErrorFix.queryParameter
-    val redirect = Redirect(redirectUrlWithMode)
+    val flashData = Map(errorMessage -> messageKey)
+    val redirect = Redirect(redirectUrlWithMode).flashing(Flash(flashData))
 
     val actualDeclaration: Future[Option[ExportsDeclaration]] = request.declarationId.map { decId =>
       customsDeclareExportsConnector.findDeclaration(decId)
--- a/app/controllers/SubmissionsController.scala
+++ b/app/controllers/SubmissionsController.scala
@@ -23,10 +23,11 @@ import controllers.util.SubmissionDisplayHelper
 import javax.inject.Inject
 import models.Mode.ErrorFix
 import models.requests.{AuthenticatedRequest, ExportsSessionKeys}
-import models.responses.FlashKeys.errorMessage
+import models.responses.FlashKeys
 import models.{ExportsDeclaration, Mode}
 import play.api.i18n.I18nSupport
 import play.api.mvc._
+import services.model.FieldNamePointer
 import uk.gov.hmrc.play.bootstrap.controller.FrontendController
 import views.html.{declaration_information, submissions}
 
@@ -76,7 +77,13 @@ class SubmissionsController @Inject()(
 
   def amendErrors(id: String, redirectUrl: String, pattern: String, messageKey: String): Action[AnyContent] = authenticate.async { implicit request =>
     val redirectUrlWithMode = redirectUrl + ErrorFix.queryParameter
-    val flashData = Map(errorMessage -> messageKey)
+    val fieldName = FieldNamePointer.getFieldName(pattern)
+    val flashData = fieldName match {
+      case Some(name) if messageKey.nonEmpty => Map(FlashKeys.fieldName -> name, FlashKeys.errorMessage -> messageKey)
+      case Some(name)                        => Map(FlashKeys.fieldName -> name)
+      case None if messageKey.nonEmpty       => Map(FlashKeys.errorMessage -> messageKey)
+      case _                                 => Map.empty[String, String]
+    }
     val redirect = Redirect(redirectUrlWithMode).flashing(Flash(flashData))
 
     val actualDeclaration: Future[Option[ExportsDeclaration]] = request.declarationId.map { decId =>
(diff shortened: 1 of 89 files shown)


app/controllers/SubmissionsController.scala
```diff
@@ -73,9 +74,10 @@ class SubmissionsController @Inject()(
     }
   }
 
-  def amendErrors(id: String, redirectUrl: String): Action[AnyContent] = authenticate.async { implicit request =>
+  def amendErrors(id: String, redirectUrl: String, pattern: String, messageKey: String): Action[AnyContent] = authenticate.async { implicit request =>
     val redirectUrlWithMode = redirectUrl + ErrorFix.queryParameter
-    val redirect = Redirect(redirectUrlWithMode)
+    val flashData = Map(errorMessage -> messageKey)
```
PatrykRudnicki, line 79: I'm wondering if you'll be adding the field name to this method.

The error support for the page require error message and field name to display. Here is just error Message. I think it will be good to add it to have it for the future

SteveSugden: Yes.  When we add the field names, this flashData map will need to be updated.  I believe there is a new ticket to do this.

app/controllers/declaration/AdditionalFiscalReferencesController.scala
```diff
@@ -43,11 +43,12 @@ class AdditionalFiscalReferencesController @Inject()(
   mcc: MessagesControllerComponents,
   additionalFiscalReferencesPage: additional_fiscal_references
 )(implicit ec: ExecutionContext)
-    extends FrontendController(mcc) with I18nSupport with ModelCacheable {
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
 
   def displayPage(mode: Mode, itemId: String): Action[AnyContent] = itemAction(itemId) { implicit request =>
-    request.item.additionalFiscalReferencesData.fold(Ok(additionalFiscalReferencesPage(mode, itemId, form()))) { data =>
-      Ok(additionalFiscalReferencesPage(mode, itemId, form(), data.references))
+    val frm = form().withSubmissionErrors()
```
PatrykRudnicki, line 49: Typo `frm` :)

app/controllers/declaration/AdditionalInformationController.scala
```diff
@@ -45,14 +45,15 @@ class AdditionalInformationController @Inject()(
   mcc: MessagesControllerComponents,
   additionalInformationPage: additional_information
 )(implicit ec: ExecutionContext)
-    extends FrontendController(mcc) with I18nSupport with ModelCacheable {
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
 
   val elementLimit = 99
 
   def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    val frm = form().withSubmissionErrors()
```
PatrykRudnicki, line 53: The same here

app/controllers/declaration/BorderTransportController.scala
```diff
@@ -54,9 +54,10 @@ class BorderTransportController @Inject()(
       case (Some(meansType), Some(meansId), meansNationality) => Some(BorderTransport(meansNationality, meansType, meansId))
       case _                                                  => None
     }
+    val frm = form().withSubmissionErrors()
```
PatrykRudnicki, line 57: It's probably in every controller :(

app/controllers/declaration/AdditionalFiscalReferencesController.scala
SteveSugden: No.  Some controllers wouldn't allow 'form' as a variable name (probably due to the way the 'form()' method was imported.
I opted for the least-impact change!

app/controllers/declaration/PreviousDocumentsController.scala
```diff
@@ -93,6 +94,7 @@ class PreviousDocumentsController @Inject()(
     }
   }
 
+  // TODO - get rid of this method in favour of Navigator backlink
```
PatrykRudnicki, line 97: 👍

app/controllers/declaration/TotalPackageQuantityController.scala
```diff
@@ -39,14 +39,14 @@ class TotalPackageQuantityController @Inject()(
...
 
   private val validTypes = Seq(DeclarationType.STANDARD, DeclarationType.SUPPLEMENTARY)
 
-  def displayPage(mode: Mode) = (authorize andThen journey(validTypes)) { implicit request =>
-    val form = TotalPackageQuantity.form(request.declarationType)
-    val data = request.cacheModel.totalPackageQuantity.fold(form)(form.fill)
-    Ok(totalPackageQuantity(mode, data))
+  def displayPage(mode: Mode): Action[AnyContent] = (authorize andThen journey(validTypes)) { implicit request =>
+    val form = TotalPackageQuantity.form(request.declarationType).withSubmissionErrors()
+    val datax = request.cacheModel.totalPackageQuantity.fold(form)(form.fill)
```
PatrykRudnicki, line 48: Here it looks like a typo too

app/services/model/RejectionReason.scala
```diff
@@ -17,70 +17,43 @@
...
-  val allRejectionReasons: List[RejectionReason] = {
-    val reader =
-      CSVReader.open(Source.fromURL(getClass.getClassLoader.getResource("code-lists/errors-dms-rej-list.csv"), "UTF-8"))
-
-    val errors: List[List[String]] = reader.all()
-
-    errors.map(RejectionReason.apply)
+  def extractErrorDescription(useImprovedErrorMessage: Boolean, cdsDescription: String, exportsDescription: String): String = {
+    val useDescription = if (useImprovedErrorMessage) exportsDescription else cdsDescription
+    if (useDescription.isEmpty) cdsDescription else useDescription
```
PatrykRudnicki, line 38: Not sure if I understand the logic correctly.
Is it the same below? Proposal:
```
val useDescription = if(useImprovedErrorMessage && exportsDescription.nonEmpty) exportsDescription else cdsDescription
```

Putting this in one if can be more readable, because it will show that use `exportsDescription` only when exists and feature flag is on

app/views/rejected_notification_errors.scala.html
```diff
@@ -34,14 +34,16 @@
...
 
 @errorDesciption(error: RejectionReason) = @{
-  val description = if(appConfig.isUsingImprovedErrorMessages) error.exportsDescription else error.cdsDescription
+  val description = error.summaryErrorMessage
   if(description.endsWith(".")) description else s"$description."
 }
 
 @errorChangeAction(error: RejectionReason) = @{
-  if(changeErrorLinkConfig.isEnabled && error.url.isDefined && error.url.get.nonEmpty) {
+  if(changeErrorLinkConfig.isEnabled && error.url.exists(_.nonEmpty)) {
```
PatrykRudnicki, line 42: 👍 I always forgot about `exists` method :D

conf/code-lists/field-name-pointer.csv
```diff
@@ -0,0 +1,2 @@
+#THIS IS JUST, AN EXAMPLE
+declaration.items.$.documentProduced,documentTypeCode
```
PatrykRudnicki, line 2: 👍 Probably worth pointing that incorrect example

app/controllers/declaration/TotalPackageQuantityController.scala
SteveSugden: It's a more trendy form of data.  Like SpaceX!

app/services/model/RejectionReason.scala
SteveSugden: Yes.  That is probably clearer.  I was just keeping a "flavour" of the existing methods.

app/controllers/declaration/AdditionalInformationController.scala
SteveSugden: Addressed this one as an example.  Not sure its worth it though.

PR #948, opened 2020-06-04
CEDS-2453 Move start page to GDS

PatrykRudnicki:


PatrykRudnicki:
@SteveSugden margin top and margin bottom classes can be removed and we can start using the gds classes. it's used in more places.

Dashed list is not supported in GDS, this is the reason that this require the own styles :(

PR #951, opened 2020-06-05
BAU Fix for triangle

PatrykRudnicki:


app/assets/stylesheets/partials/_gdsBase.scss
```diff
@@ -9,8 +9,6 @@ needs to be behind the dropdown arrow.
 .govuk-template__body {
   margin: 0;
   background-color: #fff;
-  position: relative;
```
PauloPortugal, line 12: so with position absolute this works? Interesting

PatrykRudnicki: Nope, There is another change required, which I changed locally in the browser console,

PatrykRudnicki: Now PR contains this, but this require testing

PauloPortugal: You might as well remove the `.govuk-template__body`, as that was only there to override the `z-index`.

app/assets/stylesheets/partials/_autocomplete.scss
```diff
@@ -36,4 +36,9 @@ $custom-combobox-input-border-width: 2px;
   @media screen and (min-width: $media-query-breakpoint) {
     right: calc(50% - #{$custom-combobox-input-border-width});
   }
+}
+
+.autocomplete__wrapper {
```
PauloPortugal: this wont work, as it will mix the dropdown elments with the content on the page. Please check

app/assets/stylesheets/partials/_gdsBase.scss
PauloPortugal: This works...

app/assets/stylesheets/partials/_autocomplete.scss
PatrykRudnicki: Removed this part

PR #952, opened 2020-06-08
CEDS-2504 Fix for the wrapping issue on rejected notification page

PatrykRudnicki:


app/assets/stylesheets/partials/_exports.scss
```diff
@@ -80,10 +80,6 @@ a.button--gray, button.button--gray, .button--gray:visited {
   }
 }
 
-dt.rejected-field-name {
```
SteveSugden, line 83: I was going to comment that perhaps there might be a way to make the field-name column wider (i.e. still have no-wrapping).  But I suppose that wouldn't work if the field name as extra long.

PatrykRudnicki: Yeah, I thought about this, the first fix that I prepared was to add padding left to the middle column (because the first one contains), but this is more safer approach, that we can be 100% that the text never go to the different column.

I'm just wondering about the column width in general.

Change button has 40% of width. First column is probably to small, not sure, it's up to designers

SteveSugden: Just looked at the source for the page - didn't realise that it's constructed as a GDS `SummaryList`.   Perhaps we can also get rid of the class "rejected-field-name" on the "Key" and just get the GDS component default behaviour.  (Not that it matters because you have deleted the css class definition).

PR #953, opened 2020-06-08
CEDS-2439 Fix EORI validation

MaciejRewera:
EORI is now validated according to the following:
- First 2 characters are letters (lower or upper case)
- Followed by 12-15 digits

### DIFF
--- a/app/utils/validators/forms/FieldValidator.scala
+++ b/app/utils/validators/forms/FieldValidator.scala
@@ -151,8 +151,12 @@ object FieldValidator {
   private val emailPattern = Pattern.compile("""^\S+@\S+$""")
   val isValidEmail: String => Boolean = (name: String) => emailPattern.matcher(name).matches()
 
-  private val eoriPattern = Pattern.compile("^[a-zA-Z]{2}[\\da-zA-Z]+$")
-  val isValidEORIPattern: String => Boolean = (name: String) => eoriPattern.matcher(name).matches()
+  private val eoriDigitsAmountMin = 10
+  private val eoriDigitsAmountMax = 15
+  val isValidEori: String => Boolean = (eori: String) => {
+    val (countryCode, number) = eori.splitAt(2)
+    countryCode.forall(_.isLetter) && number.forall(_.isDigit) && isInRange(eoriDigitsAmountMin, eoriDigitsAmountMax)(number.length)
+  }
 
   private val officeOfExitPattern = Pattern.compile("^[a-zA-Z]{2}[0-9]{6}$")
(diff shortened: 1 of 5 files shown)


app/utils/validators/forms/FieldValidator.scala
```diff
@@ -151,8 +151,10 @@ object FieldValidator {
   private val emailPattern = Pattern.compile("""^\S+@\S+$""")
   val isValidEmail: String => Boolean = (name: String) => emailPattern.matcher(name).matches()
 
-  private val eoriPattern = Pattern.compile("^[a-zA-Z]{2}[\\da-zA-Z]+$")
-  val isValidEORIPattern: String => Boolean = (name: String) => eoriPattern.matcher(name).matches()
+  val isValidEori: String => Boolean = (eori: String) => {
+    val (countryCode, number) = eori.splitAt(2)
+    countryCode.forall(_.isLetter) && number.forall(_.isDigit) && isInRange(12, 15)(number.length)
```
PatrykRudnicki, line 156: Are we sure about 12-15 characters?

I remember that we had something like this before. EORI number in Sweden has two country code characters + 10 digits

MaciejRewera: Checking this back with the business.

PatrykRudnicki: I remeber that there was something with adding zeros at the end, not sure.

In general PR looks good, if you confirm this with BAs you can merge it. I'll leave the approve now :)

MaciejRewera: You're right @PatrykRudnicki. It was not taken into account that we need to accept foreign EORI numbers as well. Update-test loop. Again...

PatrykRudnicki: okay :)

app/utils/validators/forms/FieldValidator.scala
```diff
@@ -151,8 +151,12 @@ object FieldValidator {
   private val emailPattern = Pattern.compile("""^\S+@\S+$""")
   val isValidEmail: String => Boolean = (name: String) => emailPattern.matcher(name).matches()
 
-  private val eoriPattern = Pattern.compile("^[a-zA-Z]{2}[\\da-zA-Z]+$")
-  val isValidEORIPattern: String => Boolean = (name: String) => eoriPattern.matcher(name).matches()
+  private val eoriDigitsAmountMin = 10
```
PatrykRudnicki, line 154: Now I'm wondering, because when I looked more I found that it can be 2 country characters plus 8 digits too.

https://en.wikipedia.org/wiki/VAT_identification_number

PR #954, opened 2020-06-08
CEDS-2485 - Authorisations to two-step remove pattern

SteveSugden:


app/views/declaration/declaration_holder_add.scala.html
```diff
@@ -0,0 +1,87 @@
...
+
+@(mode: Mode, form: Form[DeclarationHolder], holders: Seq[DeclarationHolder])(implicit request: JourneyRequest[_], messages: Messages)
+
+@tariffBody = { @Html(messages("declaration.type.holderOfAuthorisationSummaryText", components.details_content_link())) }
+
+@backLink = @{
+    if(holders.nonEmpty)
+        DeclarationHolderController.displayPage(mode)
+    else
+        Navigator.backLink(DeclarationHolder, mode)
```
PatrykRudnicki, line 50: Should this logic be in the Navigator? Similar like clearance requests helper methods?

conf/application.conf
```diff
@@ -126,7 +126,7 @@ google-analytics {
 }
 
 assets {
-  version = "3.8.0"
+  version = "3.3.2"
```
PatrykRudnicki, line 129: Why this change?

`3.3.2` assets has problem with jQuery

docs/Customs Declare Exports AutoComplete.js
```diff
@@ -439,6 +419,10 @@ function holderOfAuthorisation(){
                 break;
         }
     }
+    if (currentPageIs("/customs-declare-exports/declaration/authorisations-required")) {
+        document.getElementById('code_no').checked = 'checked';
+        document.getElementById('submit').click();
+    }
```
PatrykRudnicki, line 425: Do we need case for the remove page?

conf/application.conf
SteveSugden: Shouldn't have committed that change - was done in the "IE panic" changes on Friday

app/views/declaration/declaration_holder_add.scala.html
SteveSugden: Possibly.  But I'd have to create a brand new `DeclarationPage` object for this. `DeclarationHolder` is the Navigator key for navigating from the summary page.  The other "helper" pages - add, remove, change - all always go back to the summary page, except this one instance (no existing holders)

docs/Customs Declare Exports AutoComplete.js
SteveSugden: Do we do that for other pages that are not on the direct path from start -> finish?

PatrykRudnicki: I don't know, I just thought about remove page, not sure why

docs/Customs Declare Exports AutoComplete.js
```diff
@@ -374,38 +374,18 @@ function additionalActors(){
 }
 
 function holderOfAuthorisation(){
-    if (currentPageIs("/customs-declare-exports/declaration/holder-of-authorisation")) {
+    if (currentPageIs("/customs-declare-exports/declaration/add-authorisation-required")) {
```
PatrykRudnicki, line 377: Please bump the version of QS.

I always forgot about it, don't know how I can remember it during review :D

app/views/declaration/declaration_holder_add.scala.html
SteveSugden: Created new `DeclarationPage` object and moved logic to Navigator.  This allowed me to change the view to stop passing in the existing holders collection just for the purpose of back-link creation

conf/application.conf
SteveSugden: Done

docs/Customs Declare Exports AutoComplete.js
SteveSugden: Done

PR #955, opened 2020-06-08
BAU Remove 40% column width for the change link column and cleanup for styles, js files 

PatrykRudnicki:


app/assets/stylesheets/partials/_gdsSummaryLists.scss
```diff
@@ -3,10 +3,6 @@
 CSS Classes to support GDS Design System SummaryLists as implemented by the Exports Prototype
 ====================================
 */
-.govuk-summary-list__actions {
```
SteveSugden, line 6: The more of these we can remove the better (IMO).

SteveSugden: Should have asked before I approved.   How does this look on the summary page?

PatrykRudnicki: Yeah, that's true, I tried to remove more classes (please look at the next commit).

PatrykRudnicki: It looks similar like previously but there is more space for the first and second column. In practice the last column contains only link `Change` and had 40% width for it. This was the reason that there were a huge space between description and change link.

With this change, there is still some space between description and change link, but first column is wider and can display text without splitting it into few lines.

PR #956, opened 2020-06-08
CEDS-2478 Prevent multiple form submissions

PauloPortugal:
This was a problem when adding export items, where we could have added
50+ empty items, simply by clicking multiple times

There was also a problem on the summary page where we could submit the
final declaration multiple times.

### DIFF
--- a/app/views/components/gds/gdsMainTemplate.scala.html
+++ b/app/views/components/gds/gdsMainTemplate.scala.html
@@ -109,3 +109,5 @@
     footerItems = footer
   )(contentBlock)
 }
+
+<script src='@routes.Assets.versioned("javascripts/preventMultipleFormSubmissions.js")'></script>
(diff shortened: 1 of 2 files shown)


app/views/components/gds/gdsMainTemplate.scala.html
```diff
@@ -109,3 +109,21 @@
     footerItems = footer
   )(contentBlock)
 }
+
+<script type="text/javascript">
```
SteveSugden, line 113: Could/should this be in it's own file?
Should it be for Chrome browsers only?

SteveSugden: Maybe not browser specific.
I managed to replicate on QA using Safari

PauloPortugal: @SteveSugden Good call.
I have also made use of ES6 arrow functions.

It also feels like it is only Firefox dealing with this issue and most browsers will have this problem.
Please review again.

PR #959, opened 2020-06-09
CEDS-2462 GDSify pages to include logout button

PauloPortugal:


app/views/error_template.scala.html
```diff
@@ -14,16 +14,20 @@
...
-@govuk_wrapper(title = pageTitle, contentHeader = Some(contentHeader), mainContent = mainContent)
+@govukLayout(
+ title = Title(pageTitle),
+ useCustomContentWidth = true) {
+ <div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+   <h1 class="govuk-heading-xl" >@heading</h1>
+   <p class="govuk-body">@message</p>
+  </div>
+ </div>
```
MaciejRewera, line 32: Is this really the case? I wouldn't expect error page has to be wider.

test/handlers/ErrorHandlerSpec.scala
```diff
@@ -55,11 +56,11 @@ class ErrorHandlerSpec extends UnitSpec with Stubs with OptionValues {
 
     "standardErrorTemplate" in {
 
-      val result = errorHandler.standardErrorTemplate("Page Title", "Heading", "Message")(request).body
+      val result = errorHandler.standardErrorTemplate("title", "heading", "message")(request).body
```
MaciejRewera, line 59: Nit: You could create constants and reference them here and on assertions.

app/views/error_template.scala.html
PauloPortugal: Yes that is the case.

PauloPortugal: Well I mean if we want to maintain the same width as we do now we will have to go with that.

test/handlers/ErrorHandlerSpec.scala
PauloPortugal: @MaciejRewera what do you mean with that? Sorry :)

PR #960, opened 2020-06-10
CEDS-2479 - fixed typo in error message and changed error block title

SteveSugden:


SteveSugden:
I did a quick text search in the acceptance test project - couldn't find either of the changed messages.  But I'll run them to make sure.

PR #966, opened 2020-06-11
CEDS-2496 - Add a link on the Declaration Information page 

SteveSugden:
...(declaration notification page) to the declaration download page


app/views/declaration/summary/EoriOrAddress.scala
```diff
@@ -33,29 +34,34 @@ object EoriOrAddress {
     addressLabel: String,
     addressChangeLabel: String,
     changeController: Call,
+    mode: Mode,
```
PauloPortugal, line 37: Why do we require the Mode here?
Just trying to understand the context.

app/views/declaration/summary/countries_section.scala.html
```diff
@@ -30,6 +30,11 @@
     locations.originationCountry.isDefined || locations.hasRoutingCountries.isDefined || locations.routingCountries.nonEmpty || locations.destinationCountry.isDefined
 }
 
+@actions(action: Actions) = @{
```
PauloPortugal, line 33: :+1:

app/views/declaration/summary/EoriOrAddress.scala
SteveSugden: So that we can re-use this component helper to render the "ActionItem" (i.e. change links) or not

app/views/declaration/summary/countries_section.scala.html
SteveSugden: This was a bit of a fudge so that I could replace Some(Action) with actions(Action) - i.e. to minimise the changes to each section of the summary page.

PR #967, opened 2020-06-12
CEDS-2509 Avoid breaking up secions when printing

PauloPortugal:
From my experimentations I could not make Firefox read the `@page size`
property. According documentation only this is not yet a standard,
instead a few browsers support it, with Chrome being one of them.

### DIFF
--- a/app/assets/stylesheets/partials/_gdsBase.scss
+++ b/app/assets/stylesheets/partials/_gdsBase.scss
@@ -42,7 +42,8 @@ from the view and the page (url/page number, etc.)
 ====================================
 */
 @media print {
-  @page { margin:0 }
+  @page { margin: 10px }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/partials/_gdsBase.scss
```diff
@@ -42,7 +42,8 @@ from the view and the page (url/page number, etc.)
 ====================================
 */
 @media print {
-  @page { margin:0 }
+  @page { margin: 10px }
```
SteveSugden, line 45: Does that add more of a border around the page?

PauloPortugal:
@SteveSugden 
> Does that add more of a border around the page?


Yes. It was requested by James and Les.

PR #969, opened 2020-06-13
CEDS-2486 - Additional Documents to two-step remove pattern

SteveSugden:
Also includes some re-factoring for DeclarationHolder (authorisations) to re-use components for add / change views, "Do you want to..?" questions, etc.


app/utils/ListItem.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object ListItem {
+
+  def createId[A](index: Int, item: A) = s"$index.${item.hashCode()}"
+
+  def findById[A](id: String, items: Seq[A]): Option[A] =
+    try {
+      id.split("\\.") match {
+        case Array(index: String, hashcode: String) if items(index.toInt).hashCode() == hashcode.toInt => Some(items(index.toInt))
+        case _                                                                                         => None
```
PatrykRudnicki, line 27: Is this case necessary? Never sam the case without anything inside `case _`.

I think that without it you'll have an exception which will be handled in catch so you can remove it.

PatrykRudnicki: Ah now I see how it looks like, formatting during review split this in few lines, so it wasn't readable

PR #971, opened 2020-06-15
CEDS-2484 'Add/Summary' pages for AdditionalActors

PauloPortugal:
This introduces the Add/Remove pattern that is in vogue in other pages.
Removes the `Remove` button that would just delete an item without user
confirmation.

https://github.com/hmrc/cds-exports-acceptance/pull/494


app/controllers/declaration/AdditionalActorsAddController.scala
```diff
@@ -0,0 +1,82 @@
...
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
+
+  val validTypes =
+    Seq(DeclarationType.STANDARD, DeclarationType.SUPPLEMENTARY, DeclarationType.SIMPLIFIED, DeclarationType.OCCASIONAL)
+
+  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType(validTypes)) { implicit request =>
+    val frm = form().withSubmissionErrors()
+    request.cacheModel.parties.declarationAdditionalActorsData match {
+      case Some(_) => Ok(declarationAdditionalActorsPage(mode, frm.fill(DeclarationAdditionalActors(None, Some(NoneOfTheAbove.value)))))
```
SteveSugden, line 52: Is "None" presumed to be the previous answer?

app/controllers/declaration/AdditionalActorsRemoveController.scala
```diff
@@ -0,0 +1,77 @@
...
+  authenticate: AuthAction,
+  journeyType: JourneyAction,
+  override val exportsCacheService: ExportsCacheService,
+  navigator: Navigator,
+  mcc: MessagesControllerComponents,
+  removePage: additional_actors_remove
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
+
+  def displayPage(mode: Mode, id: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
```
SteveSugden, line 47: You didn't fancy the "generate id from index.hashcode" pattern then?
Your implementation will work, of course.  But it allows the user to alter the url which will then display and attempt to "remove" an actor that doesn't exist.

app/controllers/declaration/ConsigneeDetailsController.scala
```diff
@@ -69,7 +69,7 @@ class ConsigneeDetailsController @Inject()(
       case DeclarationType.CLEARANCE =>
         controllers.declaration.routes.DeclarationHolderController.displayPage
       case _ =>
```
SteveSugden, line 71: Should this not always continue to Add. Actors Summary controller?  Which will then redirect to Add page if necessary.

app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -36,6 +38,13 @@ case class DeclarationAdditionalActors(eori: Option[Eori], partyType: Option[Str
 
 object DeclarationAdditionalActors extends DeclarationPage {
 
+  def buildObjectInstanceFromId(value: String): DeclarationAdditionalActors = {
+    val attributes: Array[String] = value.split('-')
+
+    if (attributes.length == 0) DeclarationAdditionalActors(None, None)
```
SteveSugden, line 44: Hmmm.  This looks fragile.  What if Eori or party type is None?
I'd advocate for using the `ListItem`  createId and findById methods.

app/controllers/declaration/ConsigneeDetailsController.scala
PauloPortugal: Well spotted!

app/forms/declaration/DeclarationAdditionalActors.scala
PauloPortugal: :thinking: I thought about that as well, but the logic makes both of them mandatory, although it does not protect us against future changes or DB migrations.

Let me have a look at that class

SteveSugden: See here for example of "findBy" - https://github.com/hmrc/customs-declare-exports-frontend/blob/bc47029d16178f2c7f6014392ea6f1e7e83ec26a/app/controllers/declaration/DocumentsProducedRemoveController.scala#L82

SteveSugden: And here how to create the Id on the summary page - https://github.com/hmrc/customs-declare-exports-frontend/blob/bc47029d16178f2c7f6014392ea6f1e7e83ec26a/app/views/declaration/documentsProduced/documents_produced.scala.html#L96

PauloPortugal:
> Looks OK i suppose - some suggestions made.
>Did product team not ask for a "change" facility as well?

No

app/controllers/declaration/AdditionalActorsRemoveController.scala
PauloPortugal: We can alter the URL both ways. I guess this approach is more predictable and memorable.

SteveSugden: What I meant is that with your approach you can do things like this - 

http://localhost:6791/customs-declare-exports/declaration/authorisations-required/AEOC-STEVE_IS_GREAT/change

To produce this 

image

Does that matter?  I don't know.

SteveSugden:
I really wish you wouldn't squash your commits.   Now I've got to review all 23 files again to see what's changed since the last time I reviewed.  And remember how you did it before.

app/forms/declaration/DeclarationAdditionalActors.scala
```diff
@@ -27,6 +27,8 @@ case class DeclarationAdditionalActors(eori: Option[Eori], partyType: Option[Str
 
   import DeclarationAdditionalActors._
 
+  def id: String = s"${partyType.getOrElse("")}-${eori.getOrElse("")}"
```
SteveSugden, line 30: Is this id needed now that you are using the hashcode-based id?
Or the `buildObjectInstanceFromId` method below?

app/controllers/declaration/AdditionalActorsAddController.scala
PauloPortugal: None or _ is just to deal with no AdditionalActors data.

app/forms/declaration/DeclarationAdditionalActors.scala
PauloPortugal: I have removed it from the code. Thank you for spotting it

PR #979, opened 2020-06-18
CEDS-2475 - update all page headings to 'govuk-...-xl' styles

SteveSugden:


app/views/declaration_information.scala.html
```diff
@@ -74,7 +74,7 @@
     backButton = Some(BackButton(messages("site.backToDeclarations"), routes.SubmissionsController.displayListOfSubmissions()))
 ) {
     <div class = "govuk-grid-row">
-        <h1 class="govuk-heading-l">@messages("submissions.declarationInformation")</h1>
+        <h1 class="govuk-heading-xl">@messages("submissions.declarationInformation")</h1>
```
MaciejRewera, line 77: Could this possibly be changed to `gdsPageHeading` ?

SteveSugden: That is a good idea.  In case we need to change it again en-mass.

PR #990, opened 2020-06-23
CEDS-2513 items two step remove

MaciejRewera:


app/views/declaration/summary/item_section.scala.html
```diff
@@ -28,10 +28,10 @@
         supporting_documents: supporting_documents
 )
 
-@(mode: Mode, item: ExportItem)(implicit messages: Messages, journeyRequest: JourneyRequest[_])
+@(mode: Mode, item: ExportItem, actionsEnabled: Boolean = true)(implicit messages: Messages, journeyRequest: JourneyRequest[_])
```
SteveSugden, line 31: What's the new 'actionsEnabled' param for?

conf/declaration.routes
```diff
@@ -241,11 +241,17 @@ POST        /nature-of-transaction               controllers.declaration.NatureO
 
 # Items Summary
 
-GET         /export-items                        controllers.declaration.ItemsSummaryController.displayPage(mode: models.Mode ?= models.Mode.Normal)
+GET         /add-declaration-item                controllers.declaration.ItemsSummaryController.displayAddItemPage(mode: models.Mode ?= models.Mode.Normal)
```
SteveSugden, line 244: Reminder - new urls -> Emily

app/views/declaration/summary/item_section.scala.html
MaciejRewera: So this is doing the same thing as passing `Mode.Print`, but I did not want to use the mode for the purpose of remove page.

SteveSugden: I see.  Didn't notice that you were re-using the item_summary for your new remove_item page.

Might be useful when we re-use these summary components to "preview" a dec we are about to clone.

MaciejRewera:
@SteveSugden Quick Submit: https://github.com/hmrc/customs-declare-exports-frontend/pull/990/commits/4f8a6e209f415bc0f6060a65950b4f743110b158

app/views/declaration/summary/item_section.scala.html
MaciejRewera: Is it worth adding new `Mode.Preview` ?

SteveSugden: Could maybe rename Mode.Print -> Mode.Preview.  Then you could use it for item_remove summary?  (And get rid of your new actionsEnabled parameter?

MaciejRewera: Sounds good! I'd really prefer doing it this way. Is "Preview" a good word for all use cases?

SteveSugden: On second thoughts, is this just an abuse of the "Mode".    Maybe it would be better to get rid of Mode.Print and implement your `actionsEnabled` flag everywhere.

Otherwise on your item_remove page you are going to be ignoring the Mode passed into the page and pass Mode.Preview (or whatever) into the item_section.

My fault really.  I should have done it properly when I did the declaration print view.

PR #997, opened 2020-06-25
BAU - remove Mode.Print and extend the use of 'actionsEnabled' flag

SteveSugden:


SteveSugden:
Once Maciej introduced the `actionsEnabled` flag because he wanted to re-use the item_section component but not in the context of printing, it seemed to me that using a Mode to control rendering, rather than navigation, was a bit of an "abuse".

I only created the "Print" mode because I was originally too lazy to add another parameter to all of the summary view components.

test/views/declaration/summary/SupportingDocumentsViewSpec.scala
```diff
@@ -76,9 +76,9 @@ class SupportingDocumentsViewSpec extends UnitViewSpec with ExportsTestData with
 
     "display all supporting documents without change buttons" when {
 
-      "mode is Print" in {
+      "actionsEnabled is false" in {
```
MaciejRewera, line 79: This case is already tested staring from line 106. So you can remove this block.

test/views/declaration/summary/UnionAndNationalCodesViewSpec.scala
```diff
@@ -76,9 +76,9 @@ class UnionAndNationalCodesViewSpec extends UnitViewSpec with ExportsTestData wi
 
     "display additional information without change buttons" when {
 
-      "mode is Print" in {
+      "actionsEnabled is false" in {
```
MaciejRewera, line 79: This case is already tested staring from line 106. So you can remove this block.

SteveSugden: Good spots!  I just focused on fixing the compile issues!

PR #1012, opened 2020-07-01
CEDS-2566 - allow new-lines for additional information description

SteveSugden:


app/forms/declaration/AdditionalInformation.scala
```diff
@@ -41,7 +41,7 @@ object AdditionalInformation extends DeclarationPage {
         .verifying("declaration.additionalInformation.description.empty", nonEmpty)
         .verifying(
           "declaration.additionalInformation.description.error",
-          isEmpty or (noLongerThan(70) and isAlphanumericWithAllowedSpecialCharacters)
+          isEmpty or (noLongerThan(70) and isAlphanumericWithSpecialCharacters(Set('\r', '\n') ++ allowedSpecialChars))
```
PatrykRudnicki, line 44: Is it only in one place?

We have (based on your message on the channel) three places where we can put multiline text, should we be consistent about the validation for all pages?

I think you can create some val with `Set('\r', '\n') ++ allowedSpecialChars)` to reuse it in every place that need this, it can be `isMultilineWithSpecialCharacters`

MaciejRewera, line 44: Could the method `isAlphanumericWithAllowedSpecialCharacters` stay? To avoid exposing private field from `FieldValidator`.

SteveSugden: The other two places where we use a text-area input field already allow any character(s).   It was just this one that restricted to a alpha-numeric plus a sub-set of "special" characters.  I've just extended that sub-set with carriage return and new-line.

But I agree with Maciej - I could create a new `isAlphanumericWithAllowedSpecialCharactersAndNewLine` comparator to avoid exposing `allowedSpecialChars` collection.

SteveSugden: Done

SteveSugden: Done

PR #1013, opened 2020-07-01
CEDS-2527 Remove aria-expanded

PatrykRudnicki:
Description of the error: Fix any of the following: ARIA attribute is not allowed: aria-expanded="false"

Snippet: 
```
<input class="govuk-radios__input" id="code_yes" name="hasTaric" type="radio" value="Yes" aria-controls="conditional-code_yes" aria-expanded="false">
```

Alert Level: ERROR

Severity: critical


PatrykRudnicki:
Those changes are not necessary and didn't solve the problem. This is biggest issue that we inherit form govuk frontend

https://github.com/alphagov/govuk-frontend/issues/979

PR #1016, opened 2020-07-05
CEDS-2527 Some accessibility fixes

PatrykRudnicki:
Tests are on Jenkins so I'm not able to verify if those fixes will solve issues from accessibility tests

### DIFF
--- a/app/views/components/gds/exportsDateInput.scala.html
+++ b/app/views/components/gds/exportsDateInput.scala.html
@@ -37,7 +37,7 @@
 }
 
 @dateInput(DateInput(
-  id = fieldName,
+  id = fieldName + "-input",
   fieldset = Some(Fieldset(
     legend = Some(Legend(
       content = Text(messages(labelKey)),
--- a/app/views/declaration/supervising_customs_office.scala.html
+++ b/app/views/declaration/supervising_customs_office.scala.html
@@ -23,18 +23,11 @@
 @import services.view.AutoCompleteItem
 @import views.Title
 @import views.BackButton
-@import views.components.gds.Styles
-@import scala.collection.immutable
-@import play.twirl.api.HtmlFormat
+@import views.components.gds.Styles._
 
 @this(
     govukLayout: gdsMainTemplate,
-    govukButton: GovukButton,
-    govukRadios: GovukRadios,
-    govukDetails : GovukDetails,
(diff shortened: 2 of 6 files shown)


app/views/components/gds/exportsDateInput.scala.html
```diff
@@ -37,7 +37,7 @@
 }
 
 @dateInput(DateInput(
-  id = fieldName,
+  id = fieldName + "-input",
```
PatrykRudnicki, line 40: In line 46 there is another ID which represents the whole fieldset. I think it's better to keep two ids (tests are based on those IDs) but to avoid duplication add "input" to input div.

The issue has been found in AXE report from accessibility tests

app/views/declaration/supervising_customs_office.scala.html
```diff
@@ -42,50 +40,32 @@
...
-
 @govukLayout(
     title = Title("declaration.warehouse.supervisingCustomsOffice.title"),
     backButton = Some(BackButton(messages("site.back"), Navigator.backLink(SupervisingCustomsOffice, mode)))){
 
     @formHelper(action = controllers.declaration.routes.SupervisingCustomsOfficeController.submit(mode), 'autoComplete -> "off") {
         @errorSummary(form.errors)
 
-        @sectionHeader(messages("declaration.warehouse.supervisingCustomsOffice.sectionHeader"))
+        @components.fields.field_accessible_autocomplete_with_header(
```
PatrykRudnicki, line 56: Previous implementation used fieldset. On this page AXE show two errors related with labels.

One input on page shouldn't be wrapped in fieldset. I believe that this will solve AXE error, because this error doesn't appear on rest of the pages

MaciejRewera: Nit: I think you could remove `govukFieldset` from constructor.

PatrykRudnicki: @MaciejRewera Could you take a look once again?

Sorry I put changes into the same commit accidentally. Commit is small so it shouldn't be a problem :)

SteveSugden: 👍  I made similar changes to other "single question pages" in order to correctly display validation errors

PR #1017, opened 2020-07-06
CEDS-2531 - accessibility 'change' and 'remove' link hint texts

SteveSugden:


app/views/components/gds/link.scala.html
```diff
@@ -18,4 +18,7 @@
 
 @(text: String, textHidden: Option[String] = None, call: Call, target: String = "_self", id: Option[String] = None, classes: String = "govuk-link govuk-link--no-visited-state")
 
-<a class="@classes" href="@call" target=@target @id.map{id => id="@id" }>@{text}@if(textHidden.isDefined){<span class="govuk-visually-hidden">@{textHidden.getOrElse("")}</span>}</a>
\ No newline at end of file
+<a class="@classes" href="@call" target=@target @id.map{id => id="@id" }>
+ <span @if(textHidden.isDefined){aria-hidden="true"}>@text</span>
+ @if(textHidden.isDefined){<span class="govuk-visually-hidden">@{textHidden.getOrElse("")}</span>}
```
PatrykRudnicki, line 23: I think it's better to check additionally in if if the `textHidden` has some content.

Right now for empty textHidden we will create a empty span

SteveSugden: I think it's unlikely that we would (ab)use the optional hint text like this.  But I'll add the check.

SteveSugden: Done

PR #1019, opened 2020-07-07
Accessibility example for the ticket

PatrykRudnicki:


app/views/declaration/destinationCountries/routing_countries_summary.scala.html
```diff
@@ -133,7 +133,7 @@
             legend = Some(Legend(
                 content = Text(messages("declaration.routingCountries.summary.header", countriesOfRouting.size)),
                 classes = Styles.gdsPageLegend,
-                isPageHeading = true
+                isPageHeading = false
```
PatrykRudnicki, line 136: This can be false because in the classes above (line 135) we added styles to make this text a page heading.

This is the solution for accessibility issues found by VNU tool.

PR #1023, opened 2020-07-09
CEDS-2515 - Show descriptions of Authorisation Type Code in dropdown list

SteveSugden:


app/views/declaration/declarationHolder/declaration_holder_summary.scala.html
```diff
@@ -25,21 +25,22 @@
...
-        govukRadios: GovukRadios,
-        errorSummary: errorSummary,
-        sectionHeader: sectionHeader,
-        pageTitle: pageTitle,
-        exportsInputText: exportsInputText,
-        link: link,
-        addAnotherQuestion: addAnotherQuestion,
-        saveButtons: saveButtons,
-        formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF
+govukLayout: gdsMainTemplate,
```
PatrykRudnicki, line 31: Previous version probably more readable, but I think this is just a formatter.

conf/code-lists/holder-of-authorisation-codes.csv
```diff
@@ -1,55 +1,53 @@
...
+"SAS","Authorisation for self-assessment"
+"SASP","Single Authorisation for Simplified Procedures"
+"SDE","Simplified declaration authorisation (SDP)"
+"SIVA","Simplified Import VAT Accounting"
+"SSE","Authorisation for the use of seals of a special type"
+"TEA","Temporary admission authorisation"
+"TEAH","Temporary Admission authorisation - Auction Houses (no guarantee for VAT required"
+"TRD","Authorisation to use transit declaration with a reduced dataset"
+"TST","Authorisation to operate storage facilities for the temporary storage of goods"
+"UKCS","UK Continental shelf"
```
PatrykRudnicki, line 53: You have two codes less than before. Is it intentional?

app/services/HolderOfAuthorisationCode.scala
```diff
@@ -18,9 +18,24 @@ package services
...
+import scala.util.matching.Regex
+
+case class HolderOfAuthorisationCode(code: String, description: String) {
+
+  def asString = s"${description} (${code})"
+}
 
 object HolderOfAuthorisationCode {
+
+  private val regex: Regex = """^"?(\w+)"?,"?([^"\n]+)"?$""".r
```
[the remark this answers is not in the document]
SteveSugden: Yes, that is correct.  The regex is based on similar list-from-csv-files classes (e.g. CustomsOffices) but I enhanced it so that it can cope with the "code" being quoted or not.
To be honest, I was surprised to see how we handle existing csv files and that we are not using some sort of library.
Alternatively, I could have used json I suppose.  There was no requirement to use csv.  The "source" is a Google doc so I had to create the csv file anyway.

conf/code-lists/holder-of-authorisation-codes.csv
SteveSugden: Yes - the two no-deal codes that have been removed from the tests.  TBH I did not check the actual before and after codes.  I just created the csv file from the contents of the Google doc i.e. assumed that doc was the source of truth.

app/services/HolderOfAuthorisationCode.scala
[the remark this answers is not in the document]
PatrykRudnicki: We have library `scala-csv`. Make everything really easy, but I added this library at the end, when we had a lot of difference codes, so some of the code lists are still handled by `FileReader`

CSV library is really good, the only cons is that they return `List[List[String]]`, so list of rows and every row list of values, so you need to use pattern matching to retrieve the values, but still I think it's good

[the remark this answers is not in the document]
SteveSugden: I didn't spot that library!   Doh!. 

I will change my PR to use it then.  As there isn't much else on shall I change the other regex-based code-list readers as well?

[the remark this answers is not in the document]
PatrykRudnicki: Yeah, if you want :)

app/services/CustomsOffices.scala
```diff
@@ -16,18 +16,18 @@
...
-  private val regex: Regex = """^(\w+),"?([^"\n]+)"?$""".r
+  def all: List[CustomsOffice] = {
 
-  def all: List[CustomsOffice] =
-    FileReader("code-lists/customs-offices.csv").tail.map {
-      case regex(code: String, description: String) =>
-        CustomsOffice(code, description)
-    }.sortBy(_.description)
+    val reader = CSVReader.open(Source.fromURL(getClass.getClassLoader.getResource("code-lists/customs-offices.csv"), "UTF-8"))
+    val codes: List[Map[String, String]] = reader.allWithHeaders()
```
PatrykRudnicki, line 29: 👍  Didn't know about read with headers

SteveSugden: It was the simplest way to get the CSVReader to read and ignore the headers.  Plus getting the values as a map seems nicer than a list.

PatrykRudnicki: Yeah, looks great, I used readAll or something like this, but needed to remove the headers from the file

PR #1024, opened 2020-07-10
CEDS-2564 - convert view specs to use validating MessagesAPI

SteveSugden:
Draft PR to get feedback on direction of travel


test/views/declaration/AdditionalActorsRemoveViewSpec.scala
```diff
@@ -70,7 +67,7 @@ class AdditionalActorsRemoveViewSpec extends UnitViewSpec with ExportsTestData w
         val view = createView()
 
         view.select("dl>div:nth-child(1)>dt").text() mustBe messages("declaration.additionalActors.table.party")
-        view.select("dl>div:nth-child(1)>dd").text() mustBe "declaration.partyType.MF"
+        view.select("dl>div:nth-child(1)>dd").text() mustBe messages("declaration.partyType.MF")
```
MaciejRewera, line 70: Would matcher `containMessage` be more suitable?

app/views/declaration/additional_fiscal_references.scala.html
```diff
@@ -82,7 +82,7 @@
 @inputFields = {
     @components.fields.field_accessible_autocomplete(
         field = form("country"),
-        label = messages("declaration.additionalFiscalReferences.country"),
+        label = "declaration.additionalFiscalReferences.country",
```
MaciejRewera, line 85: IMHO it is better for the label to consume the final message, not the key. This way you have more loose coupling between `field_accessible_autocomplete` and its clients.

PatrykRudnicki: But this is how it was implemented before.

But the new specs will recognise if you use `messages(messages("messagesKey"))` which is a reason of this change

test/views/declaration/AdditionalActorsSummaryViewSpec.scala
```diff
@@ -69,23 +66,23 @@ class AdditionalActorsSummaryViewSpec extends UnitViewSpec with ExportsTestData
...
+        view.getElementsByTag("h1").text() mustBe messages("declaration.additionalActors.table.multiple.heading", "0")
       }
 
       "display page title for multiple items" in {
         createView(actors = Seq.empty).getElementsByTag("h1").text() mustBe messages("declaration.additionalActors.table.multiple.heading", "0")
       }
 
       "display section header" in {
-        view.getElementById("section-header").text() must include("declaration.summary.parties.header")
+        view.getElementById("section-header").text() must include(messages("declaration.summary.parties.header"))
```
MaciejRewera, line 77: `containMessage` matcher seems like a way to go there.

test/views/declaration/spec/UnitViewSpec2.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.jsoup.nodes.Document
+import org.scalatest.Assertion
+import org.scalatest.matchers.{BeMatcher, MatchResult}
+import play.api.i18n.{Lang, Messages, MessagesApi}
+import play.api.mvc.{AnyContent, Request}
+import play.api.test.{FakeRequest, Helpers}
+import services.cache.ExportsTestData
+import unit.base.{JourneyTypeTestRunner, UnitSpec}
+
+class UnitViewSpec2 extends UnitSpec with ViewMatchers with JourneyTypeTestRunner {
```
MaciejRewera, line 29: Nit: Isn't `JourneyTypeTestRunner` part of `UnitSpec` ?

app/views/declaration/additional_fiscal_references.scala.html
MaciejRewera: Alright, didn't realise that. No problem then (at least for now).

app/views/components/fields/field_accessible_autocomplete_with_header.scala.html
```diff
@@ -47,7 +47,7 @@
 
     @if(isPageHeading) {<h1 class="govuk-label-wrapper">}
     <label id="@(fieldNameWithoutBrackets)-label" for="@fieldNameWithoutBrackets" class="govuk-label@if(labelClass.nonEmpty){ @labelClass}" >
-        @messages(label)
+        @if(label.nonEmpty) { @messages(label) }
```
SteveSugden, line 50: This change was added to fix country_of_routing change which is (I believe) incorrectly using an empty <label>

app/views/declaration/exporter_details.scala.html
```diff
@@ -103,7 +103,7 @@
         @govukFieldset(Fieldset(
             attributes = Map("id" -> "details"),
                 legend = Some(Legend(
-                    content = Text(messages("")),
+                    content = Text(""),
```
SteveSugden, line 106: Also looks wrong - label with no text

PatrykRudnicki: In line 110 you have `exportsName`

You can move exportsName content to this legend content and remove exportsName.

With this approach you get rid of empty message here

SteveSugden: Should I fix this bug as part of this ticket?  There was some other place with an empty label as well.

PatrykRudnicki: I think it's not necessary, up to you :)

SteveSugden: I'll raise a new ticket to look at the two places where we have an empty label (from an accessibility point of view).

Could you take another look at my latest commit @PatrykRudnicki ?   This one removes redundant messages and fixes a couple of issues spotted on the way.

conf/messages.en
```diff
@@ -1284,8 +1171,6 @@ declaration.summary.transport.departure.transportCode.no-code = I don’t know
 
 declaration.summary.transport.departure.meansOfTransport.header = Transport details at the border
 declaration.summary.transport.departure.meansOfTransport.header.change = Change transport details at the border
-declaration.summary.transport.departure.meansOfTransport. =
-# this is hack about incorrectness of model ^^
```
PatrykRudnicki, line 1288: Are you sure this is not required?

SteveSugden: Only if we don't have a meansOfTransport and we always do.

PR #1030, opened 2020-07-14
CEDS-2537 - fix bug with changing proc code from WH to non-WH required

SteveSugden:
The bug was that I was checking if any item had a "warehouse required" procedure code _before_ applying the changed procedure code.    So changing from a WH-required to non-WH-required was failing to remove the WH data.

### DIFF
--- a/app/controllers/declaration/ProcedureCodesController.scala
+++ b/app/controllers/declaration/ProcedureCodesController.scala
@@ -150,13 +150,15 @@ class ProcedureCodesController @Inject()(
         .transform(removeWarehouseIdentificationForCode)
     }
 
-    def updatedModel(model: ExportsDeclaration): ExportsDeclaration =
+    def updatedModel(model: ExportsDeclaration): ExportsDeclaration = {
+      val updatedModel = model.updatedItem(itemId, item => item.copy(procedureCodes = Some(updatedProcedureCodes)))
       updatedProcedureCodes.procedureCode match {
         case Some(code) =>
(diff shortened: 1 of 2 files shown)


app/controllers/declaration/ProcedureCodesController.scala
```diff
@@ -150,13 +150,15 @@ class ProcedureCodesController @Inject()(
...
-    def updatedModel(model: ExportsDeclaration): ExportsDeclaration =
+    def updatedModel(model: ExportsDeclaration): ExportsDeclaration = {
+      val updatedModel = model.updatedItem(itemId, item => item.copy(procedureCodes = Some(updatedProcedureCodes)))
       updatedProcedureCodes.procedureCode match {
         case Some(code) =>
-          clearDataForProcedureCode(code, itemId, model).updatedItem(itemId, item => item.copy(procedureCodes = Some(updatedProcedureCodes)))
+          clearDataForProcedureCode(code, itemId, updatedModel)
         case _ =>
-          model.updatedItem(itemId, item => item.copy(procedureCodes = Some(updatedProcedureCodes)))
+          updatedModel
```
PatrykRudnicki, line 159: Ah so the whole change is around updating model before clearing data. Previously we first clean and after that updated.

SteveSugden:
Basically, we need to remove the Warehouse answer if non of the procedure codes on any of the items has a specific ending (aka a "requires WH" proc code).

So when updating an existing proc code, you need to apply that change to the item in question and then check all of the proc codes on all of the items to see if there is one that is WH required.

Previously, I was checking the model before the proc code was changed (from requiring WH to not requiring WH).   So it was failing to remove the WH details even though they were no longer required.

PR #1031, opened 2020-07-14
CEDS-2521 - You can add duplicate parties to declaration

SteveSugden:


test/unit/controllers/declaration/AdditionalActorsAddControllerSpec.scala
```diff
@@ -119,7 +119,7 @@ class AdditionalActorsAddControllerSpec extends ControllerSpec with ErrorHandler
       "user put incorrect data" in {
 
         val longerEori = TestHelper.createRandomAlphanumericString(18)
-        val wrongAction = Seq(("eori", longerEori), ("partyType", "CS"), saveAndContinueActionUrlEncoded)
+        val wrongAction = Seq(("eoriCS", longerEori), ("partyType", "CS"), saveAndContinueActionUrlEncoded)
```
PatrykRudnicki, line 122: Why you changed this?

SteveSugden: Because the test was passing incorrectly.  It was returning a 400 because the eori field was incorrectly named, not because the controller found anything wrong with the data (i.e. duplicate or value too long, etc)

PatrykRudnicki: Yeah, got it, this page has field name for eori with party type.

Thanks

PR #1033, opened 2020-07-16
CEDS-2628 - Use 12-hour time format in UI

SteveSugden:


app/views/ViewDates.scala
```diff
@@ -16,18 +16,23 @@
...
+import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import java.time.temporal.TemporalAccessor
 
 object ViewDates {
 
-  val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm")
-  val submissionDateTimeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM uuu 'at' HH:mm").withZone(ZoneId.of("Europe/London"))
-  val dateFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("d MMMM uuu").withZone(ZoneId.of("Europe/London"))
+  private val formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm")
```
SteveSugden, line 25: Made the formatters private to force consumers to use the "format" methods.
BTW - I have a feeling that the "yyyy-MM-dd HH:mm" above isn't used anywhere.  I'll investigate and create a new PR if that is the case.

app/views/ViewDates.scala
```diff
@@ -16,18 +16,23 @@
...
+  private val submissionDateTimeFormatter: DateTimeFormatter =
+    DateTimeFormatter.ofPattern("d MMMM uuu 'at' h:mma").withZone(ZoneId.of("Europe/London"))
```
PatrykRudnicki, line 27: `mma`? what `a` means? Is it for `am` and `pm`?

SteveSugden: Yes.  But it produces "AM" and "PM" unfortunately - no option for lower-case - hence the string replacement

PR #1035, opened 2020-07-17
CEDS-2621 - retain any existing UN code when updating for 'isExs'

SteveSugden:


SteveSugden:
Yes.  Maybe I should have added a reference to CEDS-2495 in the comment.  Basically, for clearance journey, UN code is only available to the user if isExs is true.  If not, the UN code page is not displayed.

What I was originally trying to do was 
(a) remove any previous UN code if the user said No
(b) add an "empty" UN code if the user said Yes.  This was for the case where user changed IsExs from No to Yes so that on the summary screen the UN code section would be displayed with a "No" answer which would allow the user to change the UN code.  But of course, it was deleting any previously entered UN codes.

PR #1039, opened 2020-07-18
CEDS-2569 - Various issues with declaration UI

SteveSugden:
The majority of the files changed in this PR are views with radio options.  

Problem:
In our UIs, the only "error" that is possible with a radio option is failing to pick one.  When that happens the FormError has a key of the radio group field which is not possible to navigate to from the link in error summary.   A couple of forms attempted to get round this by adding an id with the same value as the field to the radios fieldset but this just scrolled the page down without setting focus.  According to this https://design-system.service.gov.uk/components/error-summary/ we should navigate to the "first option".

Solutions:
I've created a helper object ("ErrorMapper") that can be used by pages that contain radio options.  It changes the FormError key for any error that is keyed on the radio group field to the id of the specified option (i.e. the first option)

ErrorMapper also does something similar with the two pages where we still ask the user to supply EORI or Address.   If neither are supplied the FormError is updated to the id of the EORI field (again, following the design guidelines of focusing on the first field in a group)

The other two big sources of change are changing the order of fields on `TotalNumberOfItems` and `Document`.   I am wondering if I should have gone this far.   I originally just changed the mapping order and then created "form@Model" and "model2Form" methods to match the existing constructors. Perhaps that was enough.


app/controllers/declaration/PreviousDocumentsRemoveController.scala
```diff
@@ -72,6 +72,8 @@ class PreviousDocumentsRemoveController @Inject()(
     }
   }
 
+  private def removeYesNoForm: Form[YesNoAnswer] = YesNoAnswer.form(errorKey = "declaration.previousDocuments.remove.empty")
```
SteveSugden, line 75: Need this to supply a custom message.

app/views/ErrorMapper.scala
```diff
@@ -0,0 +1,30 @@
...
+package views
+
+import play.api.data.FormError
+
+object ErrorMapper {
+
+  def eoriOrAddressErrors(groupKey: String, errors: Seq[FormError]) =
+    errors.map(err => err.copy(key = if (err.key == groupKey) s"$groupKey.eori" else err.key))
+
+  def radioGroupErrors(groupKey: String, firstOption: String, errors: Seq[FormError]) =
```
SteveSugden, line 26: This is for "general" radio groups.
The "yesNo" version is for radio groups using the pre-built 'yesNoRadios' component

app/forms/declaration/Document.scala
```diff
@@ -18,13 +18,15 @@ package forms.declaration
...
+import models.declaration.DocumentCategory
+import models.declaration.DocumentCategory.{RelatedDocument, SimplifiedDeclaration}
 import play.api.data.Forms.{optional, text}
 import play.api.data.{Form, FormError, Forms}
 import play.api.libs.json.{JsValue, Json}
 import services.DocumentType
 import utils.validators.forms.FieldValidator._
 
-case class Document(documentCategory: String, documentType: String, documentReference: String, goodsItemIdentifier: Option[String]) {
+case class Document(documentType: String, documentReference: String, documentCategory: DocumentCategory, goodsItemIdentifier: Option[String]) {
```
SteveSugden, line 29: Re-ordering mapping/constructor params to match field order on form.  Also decided to create DocumentCategory as a type.

app/forms/declaration/TotalNumberOfItems.scala
```diff
@@ -22,7 +22,7 @@ import play.api.data.{Form, Forms}
 import play.api.libs.json.Json
 import utils.validators.forms.FieldValidator._
 
-case class TotalNumberOfItems(totalAmountInvoiced: Option[String], exchangeRate: Option[String])
+case class TotalNumberOfItems(exchangeRate: Option[String], totalAmountInvoiced: Option[String])
```
SteveSugden, line 25: Reordering mapping/constructor to match field order on form

app/models/declaration/DocumentCategory.scala
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.declaration
+
+class DocumentCategory(val value: String) {
```
MaciejRewera, line 19: I think you could make it `abstract`.

app/models/declaration/DocumentCategory.scala
```diff
@@ -0,0 +1,43 @@
...
+  override def equals(obj: Any): Boolean = obj match {
+    case that: DocumentCategory => this.value == that.value
+    case _                      => false
+  }
+
+  override def hashCode(): Int = 31 * (if (value == null) 1 else value.hashCode)
```
MaciejRewera, line 25: If you'd make the class abstract, I don't think you need these 2 methods. You are going to extend it in `case object`s anyway and because they are 'case'-ed they should have `equals` and `hashCode` methods correctly implemented. But I am not entirely sure about that.

app/models/declaration/DocumentCategory.scala
SteveSugden: It was abstract to begin with but I needed to be able to construct one in the "Document" mapper I ended up with this 

```
.transform[DocumentCategory](value => new DocumentCategory(value), category => category.value)
```

app/forms/declaration/Document.scala
```diff
@@ -52,27 +49,30 @@ object Document extends DeclarationPage {
         "declaration.previousDocuments.documentReference.error",
         isEmpty or (isAlphanumericWithSpecialCharacters(Set('-', '/', ':')) and noLongerThan(35))
       ),
+    "documentCategory" -> requiredRadio("declaration.previousDocuments.documentCategory.error.empty")
+      .verifying("declaration.previousDocuments.documentCategory.error.empty", nonEmpty)
+      .verifying("declaration.previousDocuments.documentCategory.error.incorrect", isEmpty or isContainedIn(correctDocumentCategories))
+      .transform[DocumentCategory]({
+        case SimplifiedDeclaration.value => SimplifiedDeclaration
+        case RelatedDocument.value       => RelatedDocument
+      }, category => category.value),
```
MaciejRewera, line 58: 👍 
Nit: I've noticed that the `mapping` gets a bit long here. I don't remember the exact place, but we've had somewhere an example of having single field mappings assigned to values and then used these values in `mapping`.

app/views/declaration/additionalInformtion/additional_information_required.scala.html
```diff
@@ -48,7 +49,7 @@
     backButton = Some(BackButton(messages("site.back"), Navigator.backLink(AdditionalInformationRequired, mode, ItemId(itemId))))) {
 
     @formHelper(action = controllers.declaration.routes.AdditionalInformationRequiredController.submitForm(mode, (itemId)), 'autoComplete -> "off") {
-        @errorSummary(form.errors)
+        @errorSummary(radioGroupErrors("yesNo", "required_Yes", form.errors))
```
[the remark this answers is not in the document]
SteveSugden: Yes.  There are quite a few places where we could now use yesNoErrors instead of all the different variations.    I might look at doing that and see if I can get the PR even bigger.

app/views/ErrorMapper.scala
MaciejRewera: Do you think it is feasable to have `Mapping` type instead of `firstOption` and then taking the first field name from it? I am not sure if it is ordered or accessible in any way. But this could save someone from little errors with order or even a typo.
In a perfect world you could create an extension for `Form` or `FormError` classes to integrate with existing code. But that is probably not worth the effort.

app/views/declaration/exporter_details.scala.html
```diff
@@ -16,35 +16,31 @@
...
+govukRadios: GovukRadios,
+govukDetails : GovukDetails,
+govukFieldset: GovukFieldset,
+errorSummary: errorSummary,
+sectionHeader: sectionHeader,
+saveButtons: saveButtons,
+group_address: group_address,
+exportsInputText: exportsInputText,
+pageTitle: pageTitle,
+formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF
```
MaciejRewera, line 43: ? Same here ?

app/views/declaration/previousDocuments/previous_documents_edit.scala.html
```diff
@@ -0,0 +1,128 @@
...
+
+@sectionHeader(messages("declaration.previousDocuments.heading"))
+
+@govukFieldset(Fieldset(
+    legend = Some(Legend(
+        content = Text(messages("declaration.previousDocuments.title")),
+        classes = gdsPageLegend,
+        isPageHeading = true
+    )),
+    html = HtmlFormat.fill(immutable.Seq(
```
MaciejRewera, line 119: Nit: I don't think you need `immutable` prefix. Immutable implementations are preferable and default. It is a bit misleading (at least to me) that you have this prefix. If you write with prefix I'd expect it to be `mutable`.

app/views/ErrorMapper.scala
SteveSugden: The "firstOption" is the id of the first valid radio button.   It's not in the `mapping`

app/views/declaration/previousDocuments/previous_documents_edit.scala.html
SteveSugden: We have that pattern all over the place where HtmlFormat.fill is used.  Not sure why though.  I'll have a look.

app/views/declaration/consignor_eori_number.scala.html
```diff
@@ -50,7 +50,7 @@
     backButton = Some(BackButton(messages("site.back"), Navigator.backLink(ConsignorEoriNumber, mode)))){
 
     @formHelper(action = controllers.declaration.routes.ConsignorEoriNumberController.submit(mode), 'autoComplete -> "off") {
-        @errorSummary(form.errors)
+        @errorSummary(radioGroupErrors("hasEori", "Yes", form.errors))
```
PatrykRudnicki, line 53: Why it's not `code_yes`? Are we using something different for this page?

SteveSugden: Yes.  Not every page with a yes/no question is using the yesNoRadio component (and hence the code_yes, code_no ids).

PR #1040, opened 2020-07-20
CEDS-2595 Accessibility: empty table headers

janeturner:
Use visually hidden header for 'Remove' link column on the following tables:

- procedure-codes
- saved-declarations
- additional-fiscal-references

### DIFF
--- a/test/views/declaration/AdditionalInformationViewSpec.scala
+++ b/test/views/declaration/AdditionalInformationViewSpec.scala
@@ -124,6 +124,18 @@ class AdditionalInformationViewSpec extends UnitViewSpec with ExportsTestData wi
             .get(1) must containMessage("declaration.additionalInformation.table.headers.description")
         }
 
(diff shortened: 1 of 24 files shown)


test/views/declaration/AdditionalInformationViewSpec.scala
```diff
@@ -124,6 +124,18 @@ class AdditionalInformationViewSpec extends UnitViewSpec with ExportsTestData wi
             .get(1) must containMessage("declaration.additionalInformation.table.headers.description")
         }
 
+        "has visually hidden Change Link header" in {
+          view
+            .select("#additional_information thead tr th")
+            .get(2) must containMessage("site.change.header")
```
SteveSugden, line 130: :+1 This is the preferred way of checking text - `element must containMessage("message.key")`
There is also a `containMessageForElements` version where you are checking Elements (not Element)

PR #1044, opened 2020-07-23
CEDS-2653 - Item added to declaration before starting 'add item' journey

SteveSugden:


test/unit/controllers/declaration/ItemsSummaryControllerSpec.scala
```diff
@@ -177,9 +190,12 @@ class ItemsSummaryControllerSpec extends ControllerWithoutFormSpec with OptionVa
     onEveryDeclarationJourney() { request =>
       "call cache" in {
 
+        reset(mockExportsCacheService)
```
PatrykRudnicki, line 193: Hmm, why you need to reset mock here?

reset of this mock is in afterEach method in line 108. If that reset from line 108 is not enough it can mean that something is wrong in the spec.

SteveSugden: That shouldn't be there.  I added it in an early attempt to fix the tests.

PR #1051, opened 2020-07-30
CEDS-2639 Change 'total cost of exporting' currency pattern

janeturner:
Use currencyInput component on Statistical Value page


app/views/components/gds/currencyInput.scala.html
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @hint(hintKey.map(messages(_)), currencyHintKey.map(messages(_)))
+
+    @allErrors.map { error => <span id="error-message-@{field.name}-input" class="govuk-error-message">@error</span> }
+
+    <div class="hmrc-currency-input__wrapper">
+
+        <span class="hmrc-currency-input__unit" aria-hidden="true">£</span>
+        <input class="@inputClasses" id="@field.id" value="@field.value" name="@field.id" type="text"
```
SteveSugden, line 81: I think you need to add some conditional "error" class here so that the input field is highlighted in red if there is a validation error.   Something like
```class="@{if(field.hasErrors)"govuk-input--error" else inputClass}"```

PR #1065, opened 2020-08-13
BAU - Quick Submit

leszeklorek:
- script now works for all declarations
- if we need to have decs accepted by DMS, we need to get newest data from Khurram

### DIFF
--- a/docs/Customs Declare Exports AutoComplete.js
+++ b/docs/Customs Declare Exports AutoComplete.js
@@ -1,8 +1,8 @@
 // ==UserScript==
 // @name         Customs Declare Exports AutoComplete
 // @namespace    http://tampermonkey.net/
-// @version      1.62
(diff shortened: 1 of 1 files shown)


docs/Customs Declare Exports AutoComplete.js
```diff
@@ -793,19 +797,22 @@ function additionalInformation(){
...
+function addDocuments(){
     if (currentPageIs('/customs-declare-exports/declaration/items/.*/additional-documentation')) {
         switch(getDeclaration())
         {
             case 'C':
                 if (!document.querySelector("#content>article>form>table>tbody>tr>th")) {
                     document.getElementById('documentTypeCode').value ='C601';
                     document.getElementById('documentIdentifier').value ='GBIPO717572504502801';
-                    document.getElementById('add').click();
+                    document.getElementById('submit').click();
```
SteveSugden, line 815: Has the button id changed on these pages?  Or did this never work?

leszeklorek: it was not working - it needs a biggest changes if we want to add more then one thing

PR #1067, opened 2020-08-19
CEDS-1923 simplified journey holder of authorisation validation

MaciejRewera:


SteveSugden (approved):
LVGTM (for what it's worth)

PR #1078, opened 2020-09-11
CEDS-2678 confirmation page width

MaciejRewera:


PatrykRudnicki (approved):
Not agree with the problem which we're fixing, but PR looks good

PR #1081, opened 2020-09-25
CEDS-2659 Upgrade SBT to 1.3.6

MaciejRewera:
This version seems to be the last reliable version.
Higher versions contain some issues with importing a project. That would
need additional steps and workarounds to make it work.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,18 +1,16 @@
 resolvers += Resolver.url("hmrc-sbt-plugin-releases", url("https://dl.bintray.com/hmrc/sbt-plugin-releases"))(Resolver.ivyStylePatterns)
 
-resolvers += "Typesafe Releases" at "http://repo.typesafe.com/typesafe/releases/"
+resolvers += "Typesafe Releases" at "https://repo.typesafe.com/typesafe/releases/"
 
 resolvers += "HMRC Releases" at "https://dl.bintray.com/hmrc/releases"
 
-addSbtPlugin("uk.gov.hmrc" % "sbt-auto-build" % "2.9.0")
-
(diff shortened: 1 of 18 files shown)


MaciejRewera:
Test this please.

project/plugins.sbt
```diff
@@ -1,18 +1,18 @@
...
-resolvers += "Typesafe Releases" at "http://repo.typesafe.com/typesafe/releases/"
+resolvers += "Typesafe Releases" at "https://repo.typesafe.com/typesafe/releases/"
 
 resolvers += "HMRC Releases" at "https://dl.bintray.com/hmrc/releases"
 
-addSbtPlugin("uk.gov.hmrc" % "sbt-auto-build" % "2.9.0")
+addSbtPlugin("uk.gov.hmrc" % "sbt-auto-build" % "2.10.0")
 
-addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
PatrykRudnicki, line 9: Not sure if this should be a dependency inside the project. Didn't know that we had this in the project.

This is a kind of dependency that should probably leave in the global configuration for everyone, not inside the project

PatrykRudnicki (dismissed):

project/plugins.sbt
MaciejRewera: Did not think about it, just updated all plugins to the newest version. I'll have a look if it can be removed.

PatrykRudnicki: Yeah, I know, that's the reason I approved this :)

This is just to generate the dependency graphs. Codebase is not using it on 100%

PatrykRudnicki (approved):

PR #1083, opened 2020-10-06
CEDS-2720 Google form feedback link

MaciejRewera:


PatrykRudnicki (dismissed):
LGTM, just I think that `lazy vals` doesn't make any sense in that specific scenarios.

They can do performance worse in some specific cases (of course I'm not saying that this PR looks like some specific cases).

So I think it's safe and better to use normal vals instead of lazy. Use lazy only when you know that this will work as excepted and make sense

PatrykRudnicki (approved):

PR #1087, opened 2020-10-16
CEDS-2730 commodity measurements validation 2

MaciejRewera:


MaciejRewera:
Test this please.

app/forms/declaration/CommodityMeasure.scala
```diff
@@ -61,13 +84,34 @@ object CommodityMeasure extends DeclarationPage {
...
   )(CommodityMeasure.applyClearance)(CommodityMeasure.unapplyClearance)
 
+  private def isFirstGreaterOrEqual(first: String, second: String): Boolean = {
+    val firstNumOpt = Try(BigDecimal(first)).toOption
+    val secondNumOpt = Try(BigDecimal(second)).toOption
+
+    (for {
+      firstNum <- firstNumOpt
+      secondNum <- secondNumOpt
+    } yield firstNum >= secondNum).getOrElse(true)
```
PatrykRudnicki, line 109: If the second will be defined and first not defined you will have a true here. Is that correct? or Is that possible?

PatrykRudnicki (dismissed):

app/forms/declaration/CommodityMeasure.scala
MaciejRewera: It is possible and it is correct in this case, but the wording could be probably a little bit better.

PatrykRudnicki (approved):

PR #1090, opened 2020-11-02
CEDS-2722 Update duplication errors for single field pages

janeturner:
Update duplication errors to include red indentation and error message for pages with single fields.

### DIFF
--- a/app/controllers/declaration/NactCodeAddController.scala
+++ b/app/controllers/declaration/NactCodeAddController.scala
@@ -82,7 +82,7 @@ class NactCodeAddController @Inject()(
     implicit request: JourneyRequest[AnyContent]
   ): Future[Result] =
     MultipleItemsHelper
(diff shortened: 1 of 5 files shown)


PauloPortugal (approved):

app/controllers/declaration/NactCodeAddController.scala
```diff
@@ -82,7 +82,7 @@ class NactCodeAddController @Inject()(
     implicit request: JourneyRequest[AnyContent]
   ): Future[Result] =
     MultipleItemsHelper
-      .add(boundForm, cachedData, nactCodeLimit)
+      .add(boundForm, cachedData, nactCodeLimit, "nactCode")
```
MaciejRewera, line 85: What do you think about having a field in the Form class with the field name ("nactCode") and using the field instead of hard-coded Strings? At the moment it is typo away from errors that may be difficult to find.

PR #1091, opened 2020-11-03
CEDS-2723 Remove indentation from Declaration Type page section header

MaciejRewera:
This is an example on how to fix the issue in CEDS-2723.

Note:
This is done for a single page for standard declaration ONLY.
For this particular page there are separate templates for each
declaration type.


PauloPortugal (approved):
Awesome! Ship it!

PR #1092, opened 2020-11-03
CEDS-2755 Add a Pointer Based URL logic

PauloPortugal:
In some instances the WCO Declaration pointer will be returned but it
can map to two different pages on our user journey, which is depending on the
declaration type and the questions answered.

This means that we need to dynamically send the user to the correct page
based on the above.

This is an approach that keep the responsibility of this task/decision
in the RejectionReason class.

### DIFF
--- a/app/services/model/RejectionReason.scala
+++ b/app/services/model/RejectionReason.scala
@@ -20,7 +20,7 @@ import com.github.tototoshi.csv._
 import config.AppConfig
 import javax.inject.{Inject, Singleton}
 import models.declaration.notifications.Notification
-import models.{ExportsDeclaration, Pointer}
+import models.{DeclarationType, ExportsDeclaration, Pointer}
 import play.api.Logger
 import play.api.i18n.Messages
 import play.api.libs.json.{Json, OFormat}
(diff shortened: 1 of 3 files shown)


test/unit/services/model/RejectionReasonSpec.scala
```diff
@@ -243,6 +246,71 @@ class RejectionReasonSpec extends UnitSpec with ExportsTestData with Injector {
...
+      val pointer = Pointer(sections)
+      val url = "/customs-declare-exports/declaration/URL_PATH"
+      val declaration = aDeclaration(withType(DeclarationType.CLEARANCE), withIsExs(), withPersonPresentingGoodsDetails(Some(Eori("GB201920200000"))))
+
+      val expectedUrl = s"/customs-declare-exports/declaration/person-presenting-goods"
+
+      RejectionReason.url(url, declaration, Some(pointer)) mustBe expectedUrl
+    }
+
+    "correctly build a pointer [declaration.declarantDetails.details.eori] base url for a STANDARD Declaration" in {
```
MaciejRewera, line 283: Is it possible to visit this page for other journeys as well? If so, maybe change this test to be run against all of them using `JourneyTypeTestRunner` ?

test/unit/services/model/RejectionReasonSpec.scala
```diff
@@ -243,6 +246,71 @@ class RejectionReasonSpec extends UnitSpec with ExportsTestData with Injector {
...
+      val pointer = Pointer(sections)
+      val url = "/customs-declare-exports/declaration/URL_PATH"
+      val declaration = aDeclaration(withType(DeclarationType.CLEARANCE))
+
+      val expectedUrl = s"/customs-declare-exports/declaration/declarant-details"
+
+      RejectionReason.url(url, declaration, Some(pointer)) mustBe expectedUrl
+    }
+
+    "correctly build a pointer [declaration.declarantDetails.details.eori] base url for a CLEARANCE Declaration with EIX" in {
```
MaciejRewera, line 266: Nit: Should it not be "...with EXS" ?

app/services/model/RejectionReason.scala
```diff
@@ -91,8 +92,20 @@ object RejectionReason {
...
   }
+
+  private def pointerBasedUrl(url: String, declaration: ExportsDeclaration, pointerOpt: Option[Pointer]) =
+    pointerOpt match {
+      case Some(pointer) if pointer.toString == "declaration.declarantDetails.details.eori" =>
+        if (declaration.`type` == DeclarationType.CLEARANCE && declaration.isExs && declaration.parties.personPresentingGoodsDetails.nonEmpty) {
+          url.replace(urlPathPlaceholder, "person-presenting-goods")
+        } else url.replace(urlPathPlaceholder, "declarant-details")
+      case Some(pointer) => throw new IllegalArgumentException(s"Pointer [${pointer.toString}] does not have a corresponding URL")
+    }
```
MaciejRewera, line 107: 👍

MaciejRewera (approved):
LGTM
Just want to mention, that if we'll get another case like this one, then probably we'll need another level of conditionality.
If that happens, then IMHO it would be more "clean" to have separate objects extending common trait, instead of private methods (?)

glidester (approved):

PR #1105, opened 2020-11-20
CEDS-2667 Add HMRCReportTechnicalssue component

PauloPortugal:
This component is a non-ajax implementation and is the recommended
approach by the HMRC Play UI team, since the previous "GetHelpLink" was
not accessible.

This component does not work locally and it should simply open a new
page. We have been told that this will work in all the other
environments.


app/views/components/gds/gdsMainTemplate.scala.html
```diff
@@ -62,6 +66,7 @@
     @if(useTimeoutDialog) {
         <script src='@routes.Assets.versioned("lib/hmrc-frontend/hmrc/all.js")'></script>
         <script src='@routes.Assets.versioned("javascripts/timeoutDialog.js")'> </script>
+        <script>window.HMRCFrontend.initAll();</script>
```
russellthorn, line 69: 👍

glidester (dismissed):

glidester (approved):

PR #1108, opened 2020-11-23
[CEDS-2758] Fix back button flow on clearance journey

glidester:
After splitting carrier details into two pages the Navigator logic
was incorrectly modified and left out some required checks.

### DIFF
--- a/app/controllers/navigation/Navigator.scala
+++ b/app/controllers/navigation/Navigator.scala
@@ -411,7 +411,7 @@ object Navigator {
     else
       controllers.declaration.routes.RepresentativeStatusController.displayPage(mode)
 
-  private def carrierEoriNumberClearancePreviousPage(cacheModel: ExportsDeclaration, mode: Mode): Call = {
+  private def carrierEoriNumberClearancePreviousPage(cacheModel: ExportsDeclaration, mode: Mode): Call =
     if (!cacheModel.parties.declarantIsExporter.exists(_.isExporter))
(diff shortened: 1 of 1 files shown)


app/controllers/navigation/Navigator.scala
```diff
@@ -412,10 +412,13 @@ object Navigator {
...
-    if (cacheModel.parties.declarantIsExporter.exists(_.isExporter))
-      controllers.declaration.routes.DeclarantExporterController.displayPage(mode)
-    else {
+    if (!cacheModel.parties.declarantIsExporter.exists(_.isExporter))
       controllers.declaration.routes.RepresentativeStatusController.displayPage(mode)
+    else {
+      if (cacheModel.parties.consignorDetails.flatMap(_.details.address).isDefined)
+        controllers.declaration.routes.ConsignorDetailsController.displayPage(mode)
+      else
+        controllers.declaration.routes.ConsignorEoriNumberController.displayPage(mode)
```
MaciejRewera, line 421: Why the previous page relates to Consignor now? Was there a change in the journey flow, or the previous version was incorrect?

glidester: Previous version was incorrect. I've run the smoke and regression tests for standard and clearance journeys and they now pass (were failing before which is why Leszek pushed the ticket back again) :)

MaciejRewera (approved):

PR #1118, opened 2020-12-03
CEDS-2889 Set up the Exit Survey for Declaration Service

janeturner:
Add link to Exit Survey on /customs-declare-exports/declaration/confirmation page and on the user log out page.

### DIFF
--- a/app/config/AppConfig.scala
+++ b/app/config/AppConfig.scala
@@ -53,6 +53,7 @@ class AppConfig @Inject()(
   lazy val relevantLicensesUrl = loadConfig("urls.relevantLicenses")
   lazy val serviceAvailabilityUrl = loadConfig("urls.serviceAvailability")
   lazy val customsMovementsFrontendUrl = loadConfig("urls.customsMovementsFrontend")
+  lazy val feedbackFrontendUrl = loadConfig("urls.feedbackFrontendUrl")
 
   lazy val govUkUrl = loadConfig("urls.govUk")
   lazy val tradeTariffUrl = loadConfig("urls.tradeTariff")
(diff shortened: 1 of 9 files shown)


app/views/components/feedback_frontend.scala.html
```diff
@@ -0,0 +1,34 @@
...
+@import views.html.components.gds._
+@import config.AppConfig
+
+@this(paragraphBody: paragraphBody,
+        link: link,
+        appConfig: AppConfig)
+
+@()(implicit messages: Messages)
+
+<div>
```
russellthorn, line 26: very small comment, but does this page need this div? It doesn't seem to be doing anything, especially without a class.

janeturner: Good point, I copied this from the submission page into the component and didn't change it. Removed it now.

app/config/AppConfig.scala
```diff
@@ -53,6 +53,7 @@ class AppConfig @Inject()(
   lazy val relevantLicensesUrl = loadConfig("urls.relevantLicenses")
   lazy val serviceAvailabilityUrl = loadConfig("urls.serviceAvailability")
   lazy val customsMovementsFrontendUrl = loadConfig("urls.customsMovementsFrontend")
+  lazy val feedbackFrontendUrl = loadConfig("urls.feedbackFrontendUrl")
```
glidester, line 56: Using the name 'feedback' here for the exist survey I think is a bit confusing as it clashes with the 'giveFeedbackLink' below (line 66, used in the phaseBanner component). Could we call this new one 'exitSurvey' to more clearly distinguish it from the header feedback link?

glidester (changes requested):

app/config/AppConfig.scala
PauloPortugal: @glidester bike-shedding big time, but he is actually right! It is just a shame that he did not give the same suggestion on my SFUS work :stuck_out_tongue:

glidester (approved):
LGTM :slightly_smiling_face:

PR #1119, opened 2020-12-16
CEDS-2670 Code Lists update

PauloPortugal:
This is an automated way of parsing the tariff codes and
transforming them into our own CDS Exports service.

Essentially we read the files directly from the GovUK Tariff service, transform 
them and then overwrite the code lists on our microservice.

Going forward we would need to ensure the URLs are the most recent ones
and update them on the `urls.js` file and then execute `start.js` in order
to get the most up to date codes.
 
This is simply an quick approach to produce quick wins and generate a discussion
about the approach taken.

The `README.md` file explains on how to execute this task.


conf/code-lists/package-types.csv
```diff
@@ -1,379 +1,379 @@
...
+"8B","Crate, wooden"
+"8C","Bundle, wooden"
+"AA","Intermediate bulk container, rigid plastic"
+"AB","Receptacle, fibre"
+"AC","Receptacle, paper"
+"AD","Receptacle, wooden"
+"AE","Aerosol"
+"AF","Pallet, modular, collars 80cms   60cms"
+"AG","Pallet, shrinkwrapped"
+"AH","Pallet, 100cms   110cms"
```
PauloPortugal, line 37: Keep the * character

conf/code-lists/office-of-exits.csv
```diff
@@ -1,132 +1,135 @@
...
-Tilbury,GB000093
-Uk North Auth Consignor/Nees,GB000081
-UK South Auth Consignor/Nees,GB000246
-UKBA East Midlands Airport,GB000245
-UKBF Bristol International Airport,GB000029
-Warrenpoint,GB005160
-Workington,GB003280
\ No newline at end of file
+"Description","Code"
+"Central Community Transit Office","GB000001"
```
PauloPortugal, line 2: try to order by `Description`

README.md
```diff
@@ -42,3 +42,28 @@ or
 Auto Complete Script to help speed through the form journey.
 
 These scripts can be found in the docs directory.
+
+### Updating Tariff Code lists
```
MaciejRewera, line 46: Could you add some info on why do we take this data from publicly available place?

MaciejRewera (dismissed):
I am not a specialist in JS, but it looks organised and I think I would be able to make changes understanding what I am doing there.

I only find tests for these codes extractors a bit odd. I would question their value as they are not testing our logic, but rather the data and they do it in a very "picky" way.
Appreciate that was not part of your ticket, but might be something worth creating tech debt ticket for.

README.md
PauloPortugal: Done! :+1:

PauloPortugal:
@MaciejRewera  did not create tests. I have updated the JSON data on the Acceptance tests
so that updates the static data on:
* customs-declare-exports-frontend
* cds-exports-acceptance

MaciejRewera (approved):

PR #1138, opened 2021-01-19
CEDS-2785 Redirect start page to the choice page

glidester:
This work is prep for removing the start page completely
from the user journey (as it will be replaced by the
Gov.uk page).


app/controllers/StartController.scala
```diff
@@ -25,6 +25,6 @@ import views.html.start_page
 class StartController @Inject()(mcc: MessagesControllerComponents, startPage: start_page) extends FrontendController(mcc) with I18nSupport {
```
udinobi, line 25: Do we still need the start_page view? Will it maybe removed in future?

glidester: Yes, next ticket 2965 we will remove all links to the start page and remove the controller and view then.

udinobi: OK, but do we still need to inject `startPage`? It's not used anywhere.

udinobi (approved):

app/controllers/StartController.scala
glidester: Just ended up pulling the whole start_page view in the end :+1:

PR #1139, opened 2021-01-21
[CEDS-2870] Update tariff links throughout the service.

glidester:
Please take a look at my proposed solution in line with what we discussed.

There are no unit tests atm but I've implemented the new solution on the /declaration/dispatch-location page with examples of how the solution deals with multiple tariff links for a single page and also how different links for a page can be specified dependent on Journey type.


app/views/components/gds/tariffExpander.scala.html
```diff
@@ -0,0 +1,34 @@
...
+
+@htmlElements = @{ page.defineTariffContentKeys(journey).map( key =>
+ messages(s"$key.text", components.gds.tariffLink(key))
+).mkString("</br></br>")}
+
+@govukDetails(Details(
+ attributes = Map("id" -> "tariffReference"),
+ summary = Text(messages("site.details.summary_text")),
+ content = HtmlContent(htmlElements)
+))
```
udinobi, line 34: Just a cosmetic one... Increasing the indentation of lines 27 and 31-33 of at least one blank?

app/views/components/gds/tariffLink.scala.html
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import config.AppConfig
+
+@(key: String)(implicit messages: Messages, appConfig: AppConfig)
+
+<a class="govuk-link govuk-link--no-visited-state" target="_blank" href="@appConfig.tariffGuideUrl(s"urls.$key")">@messages(s"$key.linkText")</a>
```
udinobi, line 21: Just a cosmetic one... newline is missing

app/config/AppConfig.scala
```diff
@@ -151,4 +151,7 @@ class AppConfig @Inject()(
   lazy val allowListEnabled: Boolean = servicesConfig.getBoolean(allowListed)
 
   lazy val gtmContainer: String = servicesConfig.getString("tracking-consent-frontend.gtm.container")
+
+  def tariffGuideUrl(key: String): String =
+    runModeConfiguration.getOptional[String](key).getOrElse(throw new Exception(s"Missing tariff guide url key: $key"))
```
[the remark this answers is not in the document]
glidester: Same as line 40-41 in the loadConfig method, it allows us to customise the exception that is thrown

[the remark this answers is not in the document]
udinobi: Yes, yes. My point is that in any case you have the stacktrace in the log with the offending line. However, not a big deal :wink:

udinobi (approved):

app/views/components/gds/tariffLink.scala.html
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@import config.AppConfig
+
+@(key: String)(implicit messages: Messages, appConfig: AppConfig)
```
MaciejRewera, line 19: Instead of passing a key only, I'd pass the whole message. This way it is more generic.
In some places it turned out after some time that passing a key is limiting and either a new component needs to be built, or the old one is changed. In both cases there is some extra work to do, keeping in mind not to break existing things.

conf/tariffLinks.conf
```diff
@@ -0,0 +1,25 @@
...
+  tariff {
+    common {
+      dispatchLocation = "https://www.gov.uk/government/publications/uk-trade-tariff-cds-volume-3-export-declaration-completion-guide/group-1-message-information-including-procedure-codes#de-11-declaration-type-box-1-declaration-first-subdivision"
+    }
+
+    clearance {
+      dispatchLocation = "https://www.gov.uk/government/publications/uk-trade-tariff-cds-volume-3-export-declaration-completion-guide/group-1-message-information-including-procedure-codes#de-11-declaration-type-box-1-declaration-first-subdivision"
+    }
+  }
+}
```
MaciejRewera, line 25: How about switching the ordering to have structure like: `urls.tariff.<page>.<journey>`? This way you'd have links for given page closer to each other and it should be easier to update/change them.
I'd rather expect changes to be "change tariff links for page XYZ", instead of "change tariff links for journey XYZ".

app/forms/declaration/DispatchLocation.scala
```diff
@@ -48,4 +49,10 @@ object DispatchLocation extends DeclarationPage {
     val OutsideEU = "EX"
     val SpecialFiscalTerritory = "CO"
   }
+
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[String] =
+    decType match {
+      case CLEARANCE => Seq("tariff.clearance.dispatchLocation")
+      case _         => Seq("tariff.common.dispatchLocation", "tariff.common.dispatchLocation")
+    }
```
MaciejRewera, line 57: I'd change this method's name to sth a bit "simpler", like just `getTariffContent`. 

But more importantly, I'm considering a different return type. `Seq[String]` is very generic and you can't actually provide correct return values without going through several other files to see how it is being used.
I'd suggest creating a `case class` that would hold the keys in more descriptive way, or I'd consider returning ready text (extracted from `Messages`).

app/forms/DeclarationPage.scala
```diff
@@ -16,4 +16,18 @@
...
+
+  def defineJourneySpecificKey(decType: DeclarationType): String = "common"
+}
+
+abstract class ClearanceModifierPage extends DeclarationPage {
+  override def defineJourneySpecificKey(decType: DeclarationType): String = decType match {
+    case CLEARANCE => "clearance"
+    case _         => "common"
+  }
+}
```
glidester, line 33: I would have liked to split the page specific logic out from the forms objects as (Maciej originally mentioned) there is not an actual one-to-one relationship between pages and these forms objects (e.g. YesNoAnswers). Due to time constraints I've had to settle for defining new objects where necessary. I'm hoping in the next tariff links ticket we can resolve this.

app/forms/DeclarationPage.scala
```diff
@@ -16,4 +16,18 @@
...
+import models.DeclarationType.{CLEARANCE, DeclarationType}
+import models.viewmodels.TariffContentKey
+
+trait DeclarationPage {
+  def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] = Seq.empty
+
+  def defineJourneySpecificKey(decType: DeclarationType): String = "common"
+}
+
+abstract class ClearanceModifierPage extends DeclarationPage {
```
udinobi, line 28: Why an abstract class? Couldn't it still be a trait?

app/views/declaration/previousDocuments/previous_documents_summary.scala.html
```diff
@@ -137,7 +138,7 @@
             html = HtmlFormat.fill(immutable.Seq(
```
udinobi, line 138: If you replace `Seq` with `List` then you do not need to import `scala.collection.immutable`

glidester: Ok good to know but my ticket didn't touch this section of the code so probably out of scope of this review.

app/forms/DeclarationPage.scala
glidester: :+1:

app/forms/common/YesNoAnswer.scala
```diff
@@ -44,3 +47,13 @@ object YesNoAnswer {
   def form(fieldName: String = "yesNo", errorKey: String = "error.yesNo.required"): Form[YesNoAnswer] =
     Form(mapping(fieldName, errorKey))
 }
+
+object IsAuthorisationRequired extends ClearanceModifierPage {
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] =
+    Seq(TariffContentKey(s"tariff.declaration.isAuthorisationRequired.${defineJourneySpecificKey(decType)}"))
+}
+
+object PreviousDocumentsRemove extends DeclarationPage {
```
MaciejRewera, line 56: I am wondering, why any 'remove' page would require tariff to be present?

app/forms/DeclarationPage.scala
```diff
@@ -16,4 +16,18 @@
 
 package forms
 
-trait DeclarationPage
+import models.DeclarationType.{CLEARANCE, DeclarationType}
+import models.viewmodels.TariffContentKey
+
+trait DeclarationPage {
+  def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] = Seq.empty
```
MaciejRewera, line 23: What do you think about making this method abstract? To enforce implementation when creating a new page?

app/config/AppConfig.scala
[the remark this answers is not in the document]
MaciejRewera: On line 59 there is `tradeTariffUrl`. Is it related and can be removed? Or maybe the config can be moved to the new file?

app/forms/declaration/DeclarationHolder.scala
```diff
@@ -61,6 +63,9 @@ object DeclarationHolder extends DeclarationPage {
       case _ => DeclarationHolder(Some(dividedString(0).trim), Some(Eori(dividedString(1).trim)))
     }
   }
+
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] =
+    Seq(TariffContentKey(s"tariff.declaration.addAuthorisationRequired.${defineJourneySpecificKey(decType)}"))
 }
 
 object DeclarationHolderRequired extends DeclarationPage
```
MaciejRewera, line 71: There you have the page for which you created another one.

app/forms/common/YesNoAnswer.scala
```diff
@@ -44,3 +47,13 @@ object YesNoAnswer {
...
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] =
+    Seq(TariffContentKey("tariff.declaration.previousDocuments.remove.clearance"))
+}
```
MaciejRewera, line 59: You can find appropriate objects in files `forms.declaration.Document` and `forms.declaration.DeclarationHolder`.

app/forms/declaration/BorderTransport.scala
```diff
@@ -150,4 +152,7 @@ object BorderTransport extends DeclarationPage {
     } else {
       None
     }
+
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] =
+    Seq(TariffContentKey("tariff.declaration.borderTransport.1.common"), TariffContentKey("tariff.declaration.borderTransport.2.common"))
```
MaciejRewera, line 157: Nit: Could still make use of `defineJourneySpecificKey`

app/forms/declaration/CommodityDetails.scala
```diff
@@ -89,4 +90,19 @@ object CommodityDetails extends DeclarationPage {
...
+      case CLEARANCE =>
+        Seq(
+          TariffContentKey("tariff.declaration.item.commodityDetails.1.clearance"),
+          TariffContentKey("tariff.declaration.item.commodityDetails.2.clearance"),
+          TariffContentKey("tariff.declaration.item.commodityDetails.3.clearance")
+        )
+      case _ =>
+        Seq(
+          TariffContentKey("tariff.declaration.item.commodityDetails.1.common"),
+          TariffContentKey("tariff.declaration.item.commodityDetails.2.common")
```
MaciejRewera, line 105: Nit: Could make use of `defineJourneySpecificKey`.

app/forms/declaration/CommodityMeasure.scala
```diff
@@ -108,4 +109,21 @@ object CommodityMeasure extends DeclarationPage {
...
+          TariffContentKey("tariff.declaration.item.commodityMeasure.1.clearance"),
+          TariffContentKey("tariff.declaration.item.commodityMeasure.2.clearance"),
+          TariffContentKey("tariff.declaration.item.commodityMeasure.3.clearance"),
+          TariffContentKey("tariff.declaration.item.commodityMeasure.4.clearance")
+        )
+      case _ =>
+        Seq(
+          TariffContentKey("tariff.declaration.item.commodityMeasure.1.common"),
+          TariffContentKey("tariff.declaration.item.commodityMeasure.2.common"),
+          TariffContentKey("tariff.declaration.item.commodityMeasure.3.common")
```
MaciejRewera, line 125: Nit: Could make use of `defineJourneySpecificKey`.

app/forms/declaration/ConsignmentReferences.scala
```diff
@@ -32,4 +34,20 @@ object ConsignmentReferences extends DeclarationPage {
...
+        Seq(
+          TariffContentKey("tariff.declaration.consignmentReferences.1.clearance"),
+          TariffContentKey("tariff.declaration.consignmentReferences.2.clearance"),
+          TariffContentKey("tariff.declaration.consignmentReferences.3.clearance")
+        )
+      case _ =>
+        Seq(
+          TariffContentKey("tariff.declaration.consignmentReferences.1.common"),
+          TariffContentKey("tariff.declaration.consignmentReferences.2.common"),
+          TariffContentKey("tariff.declaration.consignmentReferences.3.common")
```
MaciejRewera, line 50: Nit: Could make use of `defineJourneySpecificKey`.

app/forms/declaration/ContainerAdd.scala
```diff
@@ -38,4 +40,7 @@ object ContainerAdd extends DeclarationPage {
   )(ContainerAdd.apply)(ContainerAdd.unapply)
 
   def form(): Form[ContainerAdd] = Form(mapping)
+
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] =
+    Seq(TariffContentKey("tariff.declaration.container.change.common"))
```
MaciejRewera, line 45: Nit: Could make use of `defineJourneySpecificKey`.

app/forms/declaration/ContainerFirst.scala
```diff
@@ -66,4 +68,16 @@ object ContainerFirst extends DeclarationPage {
...
+
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] =
+    decType match {
+      case CLEARANCE =>
+        Seq(
+          TariffContentKey("tariff.declaration.container.1.clearance"),
+          TariffContentKey("tariff.declaration.container.2.clearance"),
+          TariffContentKey("tariff.declaration.container.3.clearance")
+        )
+      case _ => Seq(TariffContentKey("tariff.declaration.container.common"))
```
MaciejRewera, line 80: Nit: Could make use of `defineJourneySpecificKey`.

app/forms/declaration/CusCode.scala
```diff
@@ -69,4 +71,7 @@ object CusCode extends DeclarationPage {
     )(form2Model)(model2Form)
 
   def form(): Form[CusCode] = Form(mapping)
+
+  override def defineTariffContentKeys(decType: DeclarationType): Seq[TariffContentKey] =
+    Seq(TariffContentKey("tariff.declaration.item.cusCode.common"))
```
MaciejRewera, line 76: Nit: Could make use of `defineJourneySpecificKey`.

app/forms/DeclarationPage.scala
```diff
@@ -16,4 +16,18 @@
...
+
+  def defineJourneySpecificKey(decType: DeclarationType): String = "common"
```
MaciejRewera, line 25: I would like to challenge the presence of this method (and the `ClearanceModifierPage` class). This method is not used everywhere and therefore it makes me wonder if it is not introducing unnecessary complexity.
For e.g. in `ConsignmentReferences` there are still 'clearance' and 'common' being explicitly repeated in every key. On the other hand in `CusCode` there is only one possible key but it contains `common` suffix. Even though the default implementation of `defineJourneySpecificKey` would return the same suffix.

I'd suggest to remove this method if it is possible to have only a single one. From the PR it seems to me that it is not necessary. And actually, maybe it will be easier to read through those keys if they are not built from several methods?

MaciejRewera (changes requested):
I'd like to know your opinion on `defineJourneySpecificKey` method and the 2 extra pages (`IsAuthorisationRequired` and `PreviousDocumentsRemove`) that were actually already present in the codebase.
More details are in the comments.

app/forms/common/YesNoAnswer.scala
glidester: This is in keeping with previous behavior.

app/forms/DeclarationPage.scala
glidester: Not all pages require tariff links so would force them to implement a method that they have no use for.

app/forms/declaration/DeclarationHolder.scala
glidester: Removed.

app/forms/common/YesNoAnswer.scala
glidester: Thanks, removed these and used the existing objects.

app/forms/DeclarationPage.scala
glidester: After our conversation I have removed the `ClearanceModifierPage` class as it really was not adding anything useful and moved its implementation of `defineJourneySpecificKey` (renamed for clarity to `getJourneyTypeSpecialisation`) to the base `DeclarationPage` trait.

I have not moved the match statement into every concretion of the `DeclarationPage` trait as it offended by DRY principles too much :smile: I tried to think of other ways to implement the logic but it all boils down to the fact we have 2 different concerns that are determined by declaration type: 1) The number of paragraphs in a tariff expander, 2) The content of each paragraph in the tariff expander.

The potential solutions to centralise these two concerns in one base trait caused far too much complexity and obfuscation of the code's purpose IMO. We can discuss again on Monday.

app/forms/common/YesNoAnswer.scala
```diff
@@ -16,14 +16,17 @@
...
+import models.DeclarationType.DeclarationType
+import models.viewmodels.TariffContentKey
 import play.api.data.{Form, Forms, Mapping}
 import play.api.libs.json.Json
 import utils.validators.forms.FieldValidator.isContainedIn
 
 case class YesNoAnswer(answer: String)
 
-object YesNoAnswer {
+object YesNoAnswer extends DeclarationPage {
```
MaciejRewera, line 29: I don't think extending `DeclarationPage` is necessary in this case.
It'd make sense if it was used in `Navigator`, but I can't find it being used there.

app/views/declaration/additionalActors/additional_actors_add.scala.html
```diff
@@ -37,9 +37,11 @@
         errorSummary: errorSummary,
         sectionHeader: sectionHeader,
         exportsInputText: exportsInputText,
-        tariffDetails: tariffDetails,
+        tariffExpander: tariffExpander,
         saveButtons: saveButtons,
-        formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF)
+        formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF,
+        appConfig: AppConfig
```
MaciejRewera, line 43: Maybe better inject `AppConfig` as implicit? This is a practise in some other places.
You could then avoid passing it to `tariffExpander` explicitly.

conf/messages.en
```diff
@@ -284,19 +470,15 @@ declaration.entryIntoDeclarantRecords.answer.no = No
 declaration.entryIntoDeclarantRecords.error = Select yes if this is an entry into the declarant’s records.
 
 declaration.personPresentingGoodsDetails.title = What is the EORI number of the person presenting the goods to customs?
-declaration.personPresentingGoodsDetails.help-item = Go to data element 3/41 Presenter Identification Number in the {0}
 
-declaration.exporter-address.help-item1 = Go to data element 3/1 Exporter’s name and business address in the {0}
```
MaciejRewera, line 289: 👍

MaciejRewera (commented):
Implementation looks really good. I've left one comment reg. templates and having `AppConfig` implicit. It might save some space.

However, I can hardly find any tests for the feature you've implemented. There is only a few test cases fixed that were failing due to changes in messages.
Please add tests for the classes in `forms` package at least. It would be also good to have some tests that could detect missing/changed links, but I am not sure if this is feasible.

app/views/declaration/additionalActors/additional_actors_add.scala.html
glidester: Yes but I'd have to alter all the calling controllers to have AppConfig as an implicit parameter (so basically adding more code to them to save a little in the templates). So I'm not sure I see the benefit of changing approaches tbh.

glidester: I did consider this but I'd have to alter every controller also to add AppConfig as an implicit param and as time is short I've decided the small gain isn't worth the effort.

app/forms/common/YesNoAnswer.scala
glidester: Good spot :+1: Done

glidester:
I've implemented tests for all the form classes that define tariff content keys as requested.

test/forms/common/DeclarationPageHelper.scala
```diff
@@ -0,0 +1,60 @@
...
+
+package forms.common
+
+import forms.DeclarationPage
+import models.DeclarationType.{DeclarationType, _}
+import models.viewmodels.TariffContentKey
+import models.DeclarationType
+import org.scalatest.{MustMatchers, WordSpec}
+
+trait DeclarationPageHelper extends WordSpec with MustMatchers {
```
MaciejRewera, line 25: Nit: I'd not extend these test traits here to avoid coupling.

test/forms/common/DeclarationPageHelper.scala
```diff
@@ -0,0 +1,60 @@
...
+          }
+
+          page.defineTariffContentKeys(CLEARANCE) mustBe getClearanceTariffKeys(i18nKey)
+        }
+      }
+    }
+
+  def testTariffContentKeysNoSpecialisation(
+    page: DeclarationPage,
+    i18nKey: String,
```
MaciejRewera, line 44: Nit: This is logically the same thing being called `messageKey` in other places. I'd rather avoid calling the same thing in different ways.

MaciejRewera (approved):
LGTM
Left a few comments but they're not crucial. More of a personal preference or improvement suggestion.

test/forms/common/DeclarationPageHelper.scala
glidester: :+1:

test/forms/common/DeclarationPageHelper.scala
glidester: :+1:

test/forms/common/DeclarationPageBaseSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+          for (decType: DeclarationType <- DeclarationType.values) {
+            page.defineTariffContentKeys(decType) mustBe getKeys(messageKey)
+          }
+        }
+      }
+    }
+
+  def getCommonTariffKeys(i18nKey: String) = Seq(TariffContentKey(s"$i18nKey.common"))
+
+  def getClearanceTariffKeys(i18nKey: String) = Seq(TariffContentKey(s"$i18nKey.clearance"))
```
MaciejRewera, line 59: Since you've changed parameter names for the two other methods, maybe you could change them for these two as well?

MaciejRewera (approved):
LGTM

PR #1148, opened 2021-02-16
[CEDS-2821] Receiving MRN via URL param to skip entry page

glidester:
As part of the usability improvements we are allowing the link
to SFUS from the Exports UI to include the MRN of the declaration
to pre-populate that field and remove the need for the user
to re-type it.

### DIFF
--- a/app/views/declaration_information.scala.html
+++ b/app/views/declaration_information.scala.html
@@ -43,10 +43,9 @@
     }
 
     @if(status == SubmissionStatus.ADDITIONAL_DOCUMENTS_REQUIRED && sfusConfig.isSfusEnabled) {
-        @link(text = messages("submissions.sfus"), call = Call("GET", sfusConfig.sfusLink), target="_blank")
+        @link(text = messages("submissions.sfus"), call = Call("GET", s"${sfusConfig.sfusLink}/${submission.mrn.getOrElse("")}"), target="_blank")
     }
 }
-
 
 @renderDeclarationLinks = {
     <div class="govuk-list">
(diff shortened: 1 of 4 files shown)


app/views/declaration_information.scala.html
```diff
@@ -43,11 +43,10 @@
     }
 
     @if(status == SubmissionStatus.ADDITIONAL_DOCUMENTS_REQUIRED && sfusConfig.isSfusEnabled) {
-        @link(text = messages("submissions.sfus"), call = Call("GET", sfusConfig.sfusLink), target="_blank")
+        @link(text = messages("submissions.sfus"), call = Call("GET", s"${sfusConfig.sfusLink}/${submission.mrn.getOrElse("")}"), target="_blank")
```
udinobi, line 46: IMO too much Scala logic in this template. You could pass directly the entire url as an `Option[String]` parameter from the controller, where you could in addition check if DMSDOC and isSfusEnabled.
The idea is that in general you should always try to include less/none (wherever feasible of course) Scala logic in a template, so to make easier for a frontend dev who maybe does not know Scala to work with the template.

glidester: This page is quite dynamic in nature. I could do as you suggest but to do so just for this simple SFUS link would make the approach inconsistent with how the rest of the page is coded. It would also IMO hide the logic of when an SFUS link should be present and when it shouldn't be from the reader (and I don't like that personally). I understand the principle your referring to but our UI guys are familiar with twirl and anyway they have their own prototype apps that they code so I don't see the value of doing this for us in this specific context.

MaciejRewera (approved):

PR #1153, opened 2021-03-10
CEDS-2964 Use HmrcTimeoutDialogHelper component

janeturner:
Replace current Timeout dialog component with HmrcTimeoutDialogHelper


udinobi:
If feasible (but I'm not sure) it would be nice to have a test for this feature. If not, an acceptance test should also work.

udinobi (approved):

PR #1156, opened 2021-03-19
CEDS-3105 Content amendments

udinobi:
Content amendments to dashboard and "declaration information" pages.

The big chunk of changes is in commit" CEDS-3105 Refactor views' helpers", which just moves a few helpers scala files into package _views/helpers_. What is mostly worth to be reviewed is in the other commits.


app/models/declaration/submissions/SubmissionStatus.scala
```diff
@@ -47,6 +47,10 @@ object SubmissionStatus extends Enumeration {
     case UNKNOWN                         => "Unknown status"
   }
 
+  def formatOnDeclInfoPages(status: SubmissionStatus): String = status match {
+    case ADDITIONAL_DOCUMENTS_REQUIRED => "Documents Required"
+    case _                             => format(status)
+  }
```
glidester, line 53: Should this (and the format method above) not return i18n message keys rather than text to render on a view? Display logic don't belong in the data model.

glidester (changes requested):

app/models/declaration/submissions/SubmissionStatus.scala
udinobi: You are absolutely right, but I think we need a proper tech ticket for that. I do not think should be done within this ticket.

MaciejRewera: Going this route I'd ask if `format` method should not also be mmoved from this place to some view-dedicated class/object?

app/controllers/SubmissionsController.scala
```diff
@@ -59,7 +60,8 @@ class SubmissionsController @Inject()(
     customsDeclareExportsConnector.findSubmission(id).flatMap {
       case Some(submission) =>
         customsDeclareExportsConnector.findNotifications(id).map { notifications =>
-          Ok(declarationInformationPage(submission, notifications))
+          val hasDmsdocNotifications = notifications.exists(_.status == SubmissionStatus.ADDITIONAL_DOCUMENTS_REQUIRED)
+          Ok(declarationInformationPage(submission, notifications, hasDmsdocNotifications))
```
MaciejRewera, line 64: This logic seems rather simple. Wouldn't it be worth having this inside `declaration_information` template instead of an extra flag parameter?
From testing perspective, I don't think it makes much of a difference too.

MaciejRewera: Also, this logic is related strictly to what will be present on the page, so maybe controller should not really care about such details?

app/views/declaration_information.scala.html
```diff
@@ -70,72 +70,98 @@
...
+                @if(hasDmsdocNotifications) {
+                    <p id="content-on-dmsdoc" class="govuk-body">
+                        @messages("submissions.content.on.dmsdoc")
+                        @if(sfusConfig.isSfusSecureMessagingEnabled) {
+                            @Html(messages("submissions.content.on.dmsdoc.sfus", sfusInboxLink))
+                        }
+                    </p>
+                }
+
+                <div class="govuk-grid-column">
```
MaciejRewera, line 141: Nit: Out of curiosity: how does this class work? What is the width of the element?

app/controllers/SubmissionsController.scala
udinobi: NP, since you already follow this approach project-wide. I did this way because I generally prefer the idea "no logic in templates".

app/views/declaration/summary/notifications_section.scala.html
```diff
@@ -37,9 +38,9 @@
         ))
         ++ notifications.map(notification =>
             SummaryListRow(
-                classes = s"${notification.displayStatus.toLowerCase}-row",
```
udinobi, line 40: It seems to me that this was wrong in the former implementation as we would end up with class names that have spaces between the words. Please let me know in case I'm wrong.

test/views/DeclarationInformationViewSpec.scala
```diff
@@ -67,11 +68,13 @@ class DeclarationInformationViewSpec extends UnitViewSpec with Injector {
 
   private val submission: Submission = submission()
 
-  private val zone: ZoneId = ZoneId.of("Europe/London")
+  private val dateTimeIssued = ZonedDateTime.now
+  private val dateTimeAsText = ViewDates.formatDateAtTime(dateTimeIssued)
```
glidester, line 72: Using these values in the test instead of the old literal values seems to invalidate the usefulness of these test as you are using the same function `ViewDates.formatDateAtTime` to create the expected result string as the view that is being tested.

app/views/declaration/summary/notifications_section.scala.html
glidester: :+1:

glidester (approved):

test/views/DeclarationInformationViewSpec.scala
udinobi: `formatDateAtTime` is tested in `test/views/helpers/ViewDatesSpec.scala`.
The tests here are mainly to verify that the date is shown, nor really to verify their format.

PR #1158, opened 2021-03-22
CEDS-3099 exports inbox feature flag

MaciejRewera:


app/config/SecureMessagingFeatureFlagConfig.scala
```diff
@@ -0,0 +1,61 @@
...
+
+package config
+
+import features.Feature.Feature
+import features.SecureMessagingFeatureStatus.SecureMessagingFeatureStatus
+import features.{Feature, SecureMessagingFeatureStatus}
+import javax.inject.Inject
+import play.api.Configuration
+
+class SecureMessagingFeatureFlagConfig @Inject()(config: Configuration) {
```
udinobi, line 25: The class name seems a bit redundant. What about a simple _SecureMessagingConfig_? Even because it includes not only flags.

udinobi: Uhm... I see that we already have a _SecureMessagingConfig_ :confused:

MaciejRewera: Yes, together with Glidester we named 2 classes the same way. Added a commit to rename this class again. Maybe the new name is more explanatory.

app/config/SecureMessagingInboxConfig.scala
```diff
@@ -0,0 +1,61 @@
...
+
+package config
+
+import features.Feature.Feature
+import features.SecureMessagingFeatureStatus.SecureMessagingFeatureStatus
+import features.{Feature, SecureMessagingFeatureStatus}
+import javax.inject.Inject
+import play.api.Configuration
+
+class SecureMessagingInboxConfig @Inject()(config: Configuration) {
```
udinobi, line 25: :+1:

conf/application.conf
```diff
@@ -98,12 +98,12 @@ microservice {
...
 
     features {
+      default = disabled
       betaBanner = enabled
       changeErrorLink = enabled
-      default = disabled
       ead = enabled
       sfus = enabled
-      sfusSecureMessaging = enabled
+      secureMessaging = sfus
```
udinobi, line 106: Is this actually working? Should not `sfus` be between braces (... or something like that, I do not remember the correct syntax now)

udinobi, line 106: Ok, got it. But a bit confusing just `sfus`. Maybe values like `useSfusInbox` or `viaSfusInbox` (or something like that), as well as `useExportsInbox` or `viaExportsInbox`. Or whatever, but still something more explicit.

MaciejRewera: Maybe I will rename the flag's name? Do you think `secureMessagingInbox` is more explicit?

udinobi: That's fine. I would still use something more explicit for the value too. Maybe just `useSfus` or `viaSfus`.

udinobi (approved):

PR #1162, opened 2021-03-26
CEDS-3059 Add navigation links

udinobi:
Add navigation links to a few pages when "Secure Messaging" flag is enabled


app/views/helpers/PageWithNavigation.scala
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package views.helpers
+
+object PageWithNavigation extends Enumeration {
```
glidester, line 19: This construct confuses me. It seems to be serving two different purposes, enabling and disabling the navigation links and optionally specifying what page we are on (dashboard or inbox). 

Expressions like `PageWithNavigation.OnPageWithoutNavigation` read as a contradiction of terms.

Would not a simpler solution be that each view that requires the navigation links pass them as an optional component to the gdsMainTemplate (disabling whichever link might need disabling)?

MaciejRewera: Not sure if it helps, but I have an idea to instead of having this Enumeration object, change `gdsMainTemplate` parameter to be the an optional navigation banner itself instead of a flag.
This way you could have default of `None` and for pages where you need the navigation banner you could provide it as any other parameter. 

I'm happy to talk about it if it is not explained well enough ;-)

MaciejRewera: And now I realized that I might be talking about the same thing as @glidester

glidester: Its just optionally disabling one of the two links isn't it? The same functionality could be achieved with just two Boolean parameters on 'navigationBanner' where the loading view could disable either link as appropriate. Alternatively, if 'PageWithNavigation' just had two values of 'DashboardPage' and 'MessageInboxPage' then that would be clearer.

udinobi: Sorry, Friday I was in a hurry, so I didn't make myself clear.

The banner has to be always shown on four pages (dashboard, decl information, inbox, single conversation) but only if the Secure Messaging flag is enabled. What is different between the four pages is the style to be used
- Dashboard and inbox have one link only, the other, for the page itself, become just a span tag.
- "decl information" and "single conversation" have instead two links.

While your solution, passing to gdsMainTemplate what every page needs, is simpler also a few cons
- If not there yet you need to inject SecureMessagingConfig to every page.
- On the other side, if you do inject SecureMessagingConfig only in gdsMainTemplate you will end up creating the html in every page but they won't be shown if the flag is disabled.
- If in future you need to change the href url for one of the two links you have to change it in all pages.
- If in future you need to add a link to the banner you have to add it to all pages.

I have removed now the confusing enum item and setting the banner as optional in gdsMainTemplate. This solution still seems quite clean, to me at least.

app/views/components/gds/gdsMainTemplate.scala.html
```diff
@@ -44,7 +44,8 @@
 @(
   title: Title,
   backButton: Option[BackButton] = None,
-  useCustomContentWidth: Boolean = false
+  useCustomContentWidth: Boolean = false,
+  navigationBanner: Html = HtmlFormat.empty
```
MaciejRewera, line 48: Nit: It's not that important, but you can put not only navigation banner here. You can put anything that needs to go before the page's main content.

app/views/project/build.properties
```diff
@@ -1 +0,0 @@
-sbt.version=1.4.4
```
MaciejRewera, line 1: Do we have SBT version defined somewhere else? I wonder why would you delete this file?

app/views/declaration_information.scala.html
```diff
@@ -14,26 +14,34 @@
...
     sfusConfig: SfusConfig,
-    secureMessagingInboxConfig: SecureMessagingInboxConfig
+    secureMessagingInboxConfig: SecureMessagingInboxConfig,
+    secureMessagingConfig: SecureMessagingConfig
 )
 
 @(submission: Submission, notifications: Seq[Notification])(implicit request: Request[_], messages: Messages)
 
+@navBanner = {
+  @if(secureMessagingConfig.isSecureMessagingEnabled) {
```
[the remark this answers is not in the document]
udinobi: The point to have `secureMessagingConfig.isSecureMessagingEnabled` is exactly to not have anymore references to `SecureMessagingInboxConfig` once Secure Messaging for Exports is complete. At that point in Exports we do not care anymore for the SFUS implementation. This is the reason for not using `SecureMessagingInboxConfig`

test/views/DeclarationInformationViewSpec.scala
```diff
@@ -155,12 +168,36 @@ class DeclarationInformationViewSpec extends UnitViewSpec with Injector {
       messages must haveTranslationFor("submissions.timeline")
     }
 
-    "contains page header" in {
+    "contain the navigation banner" when {
+      "the Secure Messaging flag is set to 'exports'" in {
```
MaciejRewera, line 172: This feature flag is not the one that allows for 'exports' and 'sfus' values. This one can only be 'enabled' or 'disabled'.
It's `SecureMessagingInbox` that accepts 3 values.

So the test description is incorrect.

test/views/DeclarationInformationViewSpec.scala
```diff
@@ -155,12 +168,36 @@ class DeclarationInformationViewSpec extends UnitViewSpec with Injector {
...
+      }
+    }
+
+    "contain page header" in {
 
       viewWithFeatures.getElementsByTag("h1") must containMessageForElements("submissions.declarationInformation")
     }
 
-    "contains references table with correct labels" in {
+    "contain references table with correct labels" in {
```
MaciejRewera, line 200: 👍

app/views/project/build.properties
udinobi: Mind that this file was under `views`. Probable was added here by error.

test/views/SubmissionsViewSpec.scala
```diff
@@ -37,15 +42,26 @@ import views.tags.ViewTest
...
+  private def createViewForNavigationBannerTest(enableSecureMessaging: Boolean): HtmlFormat.Appendable = {
+    val secureMessagingConfig = mock[SecureMessagingConfig]
+    when(secureMessagingConfig.isSecureMessagingEnabled).thenReturn(enableSecureMessaging)
+
+    val injector = new OverridableInjector(bind[SecureMessagingConfig].toInstance(secureMessagingConfig))
+    val page = injector.instanceOf[submissions]
+    page(SubmissionsPagesElements(Seq.empty)(mock[PaginationConfig]))(request, messages)
+  }
+
+  private val zone: ZoneId = ZoneId.of("UTC")
```
MaciejRewera, line 63: I'd expect that you can replace the "old" way of creating the view with the one you did here. Is there a reason why there can't be a single way of creating the view?

test/views/SubmissionsViewSpec.scala
```diff
@@ -99,6 +115,29 @@ class SubmissionsViewSpec extends UnitViewSpec with ExportsTestData with Stubs w
       messages must haveTranslationFor("submissions.status.header")
     }
 
+    "contain the navigation banner" when {
+      "the Secure Messaging flag is set to 'exports'" in {
```
MaciejRewera, line 119: Same case here as above with test description.

test/views/messaging/InboxWrapperSpec.scala
```diff
@@ -16,26 +16,62 @@
...
+  }
+
   "Inbox Wrapper page" should {
 
     "display page title" in {
       view.getElementsByTag("title").first must containMessage("inbox.heading")
     }
 
+    "contain the navigation banner" when {
+      "the Secure Messaging flag is set to 'exports'" in {
```
MaciejRewera, line 54: Same case here as above with test description.

MaciejRewera (changes requested):
Implementation looks good. I've left a few comments around tests and most importantly - I don't know why you'd like to delete `build.properties`?

MaciejRewera (dismissed):
Thank you for your comment on the SBT version.

test/views/SubmissionsViewSpec.scala
udinobi: The difference is that `SubmissionsViewSpec` contains many tests and having a common `createView` you would have and handle a mock in all of them that is not in use. In addition to test the navigation banner you do not need all submission used by the top `createView`.

glidester (approved):

test/views/SubmissionsViewSpec.scala
MaciejRewera: I am not convinced, TBH 😄 I'd rather go the way of having a single view creation method, the mock can have a default return value set as `false` in `beforeEach` method.
Right now there might be navigation banner or not, depending on what is defined in configuration. So effectively you have 2 instances of this config to handle, but one is kind of 'implicit'.

MaciejRewera (approved):

PR #1163, opened 2021-03-29
[CEDS-2879] Make address validation more permissive

glidester:
Before the address form validation restricted values
to be alphanumeric and latin characters only.

This is in contrast to the DMS Schema rules that allow
any characters and only limits on the field lengths
allowed.

These changes relax the implemented validation rules
to become inline with the DMS Schema rules.

### DIFF
--- a/conf/messages.en
+++ b/conf/messages.en
@@ -498,11 +498,11 @@ declaration.representative.representationType.indirect = Indirect representative
 declaration.representative.representationType.indirect.hint = The representative is sharing legal responsibility with the exporter.
 
 declaration.address.fullName = Full name
-declaration.address.fullName.error = Enter a name, up to 35 characters
-declaration.address.fullName.empty = Enter a name, up to 35 characters
+declaration.address.fullName.error = Enter a name, up to 70 characters.
+declaration.address.fullName.empty = Enter a name, up to 70 characters.
 declaration.address.addressLine = Address line 1
-declaration.address.addressLine.error = Enter an address line 1, up to 35 characters
-declaration.address.addressLine.empty = Enter an address line 1, up to 35 characters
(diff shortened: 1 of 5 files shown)


conf/messages.en
```diff
@@ -498,11 +498,11 @@ declaration.representative.representationType.indirect = Indirect representative
...
 declaration.address.fullName = Full name
-declaration.address.fullName.error = Enter a name, up to 35 characters
-declaration.address.fullName.empty = Enter a name, up to 35 characters
+declaration.address.fullName.error = Enter a name, up to 70 characters.
+declaration.address.fullName.empty = Enter a name, up to 70 characters.
 declaration.address.addressLine = Address line 1
-declaration.address.addressLine.error = Enter an address line 1, up to 35 characters
-declaration.address.addressLine.empty = Enter an address line 1, up to 35 characters
+declaration.address.addressLine.error = Enter an address line 1, up to 70 characters.
+declaration.address.addressLine.empty = Enter an address line 1, up to 70 characters.
```
MaciejRewera, line 505: 👍

test/forms/common/AddressSpec.scala
```diff
@@ -203,16 +167,21 @@ object AddressSpec {
 
   import play.api.libs.json._
 
+  val seventyOneChars = "01234567890123456789012345678901234567890123456789012345678901234567890"
+  val thirtySixChars = "0123456789012345678901234567890123456"
+  val tenChars = "01234567890"
```
MaciejRewera, line 172: Nit: There is a `TestHelper` class that contains `createRandomString` method. You can simply provide the length you want, no need to type it yourself ;)

MaciejRewera (approved):
LGTM

PR #1166, opened 2021-04-08
CEDS-3058 View individual message & Reply Success

janeturner:
Create screen to view individual message and submit a reply.
Include back button and link to declaration dashboard.


udinobi (approved):
LGTM

PR #1167, opened 2021-04-08
Ceds 3079 text changes to commodity code screen

glidester:


app/views/components/gds/exportsInsetText.scala.html
```diff
@@ -0,0 +1,31 @@
+@*
```
udinobi, line 1: Do you really need a new component? Do you expect that it will also be used in other templates?

glidester: Yeah, I think Liz has plans to use them in other pages too

udinobi (approved):

PR #1168, opened 2021-04-08
CEDS-3075 Amend valitation of goods location

udinobi:


MaciejRewera (approved):
I'm approving this, but I think that validation should be tested in `GoodsLocationFormSpec`. It is not a responsibility of the View spec to test all validation scenarios. Only to check if any error is displayed correctly.

But I know it's been like this already and you only changed existing tests.

PR #1174, opened 2021-04-16
CEDS-3076 Add inset text to 'Destination Country' page

udinobi:


app/views/declaration/destinationCountries/destination_country.scala.html
```diff
@@ -18,30 +18,37 @@
...
+    appConfig: AppConfig
 )
 
 @(mode: Mode, form: Form[Country])(implicit request: JourneyRequest[_], messages: Messages)
 
+@insetLink = {
+    <a class="govuk-link govuk-link--no-visited-state" target="_blank" href=@appConfig.govUkPageForTypeCO>
+        @messages("declaration.destinationCountry.inset.link")
+    </a>
+}
```
glidester, line 50: Nit: Why not use the link component views.components.gds.link ?

glidester (approved):

app/views/declaration/destinationCountries/destination_country.scala.html
udinobi: Yes but I still think that `insetLink` is more readable than
```

        @insetText(content = HtmlContent(messages(
            "declaration.destinationCountry.inset.text",
            link(
                text = messages("declaration.destinationCountry.inset.link"),
                call = Call("GET", appConfig.govUkPageForTypeCO),
                target = "_blank"
           )
        )))
```

PR #1176, opened 2021-04-20
Audit an explicit event for successful retrieval of Message inbox

udinobi:


app/services/audit/AuditService.scala
```diff
@@ -69,33 +69,56 @@ class AuditService @Inject()(connector: AuditConnector, appConfig: AppConfig)(im
...
   }
 
-  private def getAuditDetails(userInput: JsObject)(implicit hc: HeaderCarrier) = {
+  private def getAuditDetails(userInput: JsObject)(implicit hc: HeaderCarrier): JsObject = {
     val hcAuditDetails = Json.toJson(AuditExtensions.auditHeaderCarrier(hc).toAuditDetails()).as[JsObject]
     hcAuditDetails.deepMerge(userInput)
   }
+
+  private def detailsForMessageInboxPartialRetrieved(eori: String): JsValue =
+    Json.obj("enrolment" -> "HMRC-CUS-ORG", "eoriNumber" -> eori, "tags" -> Json.obj("notificationType" -> "CDS-EXPORTS"))
```
glidester, line 109: I think the values 'enrolment' and 'notificationType' should be passed into the AuditService from the connector where they are defined and used. Hard coding them here means if someone changes these values in the connector they might not know to change them here also.

app/services/audit/AuditService.scala
```diff
@@ -69,33 +69,56 @@ class AuditService @Inject()(connector: AuditConnector, appConfig: AppConfig)(im
...
+
+  private def tagsForMessageInboxPartialRetrieved(implicit hc: HeaderCarrier): Map[String, String] =
+    AuditExtensions
+      .auditHeaderCarrier(hc)
+      .toAuditTags(transactionName = "callExportPartial", path = secureMessagingConfig.fetchInbox)
```
glidester, line 114: Interestingly, I chose to hard code this path part for SFUS as I (maybe wrongly) assumed that this was a value that CIP would be explicitly looking for in their queries (and so should not really change). Maybe we should check with Ian & CIP about this :thinking:

glidester (changes requested):

app/services/audit/AuditService.scala
udinobi: I asked yesterday to Ian and he told me to use the path specific to Exports. The data in the Confluence page is just an example.

glidester (approved):

PR #1177, opened 2021-04-21
CEDS-3129 Remove Dispatch Location page

janeturner:
Redirect Declaration Choice to Additional Declaration Type page.
Update back button logic.
Remove references to Dispatch Location.


udinobi (approved):
LGTM

app/connectors/exchange/ExportsDeclarationExchange.scala
```diff
@@ -39,7 +39,6 @@ case class ExportsDeclarationExchange(
   updatedDateTime: Instant,
   sourceId: Option[String],
   `type`: DeclarationType,
-  dispatchLocation: Option[DispatchLocation] = None,
```
MaciejRewera, line 42: Since you are changing the contract between services, please update the other service (customs-declare-exports) as well. This field is optional on both ends so I don't expect it to cause problems if they are deployed independently.

MaciejRewera (approved):
LGTM

PR #1178, opened 2021-04-22
[CEDS-3138] National Additional Code

glidester:
Add a new tariff guide link specific for the actual commodity code
the user will have already entered on the previous page.


MaciejRewera (approved):
LGTM

PR #1179, opened 2021-04-22
CEDS-3134 Amendments to the 'TARIC additional code' page

udinobi:
Amendments, based on the commodity code specified or not, to the 'TARIC additional code' page


app/forms/declaration/TaricCodeFirst.scala
```diff
@@ -35,12 +35,13 @@ object TaricCodeFirst extends DeclarationPage {
   implicit val format = Json.format[TaricCodeFirst]
 
   val hasTaricCodeKey = "hasTaric"
+  val none = TaricCodeFirst(None)
```
glidester, line 38: Nit: not sure what value this serves, looked cleaner with the literal value in the pattern match below as before

udinobi: It's a sort of static value and it's used in a few places, app and tests.

glidester (approved):
Looks good, just one minor comment.

Our implementations of basically the same thing (just on different pages) are very similar but they do differ slightly in execution in one or two places. Not a big deal and neither is 'better' than the other but it does mean there is now inconsistency in the service. Maybe a topic to cover in the next dev catchup! :grin:

app/forms/declaration/TaricCodeFirst.scala
glidester: :+1:

PR #1188, opened 2021-04-29
BAU tests refactor

MaciejRewera:


glidester (dismissed):
Middle mouse finger has RSI now!

PR #1197, opened 2021-05-15
CEDS-3143 Add 2 new pages, 'Link DUCR to a MUCR' and 'Enter a MUCR'

udinobi:


conf/application.conf
```diff
@@ -185,6 +185,8 @@ urls {
   locationCodesForCsePremises = "https://www.gov.uk/government/publications/location-codes-to-declare-goods-for-export-at-cse-premises-for-data-element-523-of-the-customs-declaration-service"
   goodsLocationCodesForDataElement = "https://www.gov.uk/government/collections/goods-location-codes-for-data-element-523-of-the-customs-declaration-service"
   tariffCdsChiefSupplement = "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/956071/Tariff_Volume_3_Great_Britain_Supplement_CDS_and_CHIEF_version_5.3__1_.odt"
+  ducrToMucrHint1 = "https://www.gov.uk/government/publications/uk-trade-tariff-cds-volume-3-export-declaration-completion-guide/group-2-references-of-messages-document-certificates-and-authorisations#de-21-simplified-declaration-previous-documents-box-40-declaration-previous-document"
+  ducrToMucrHint2 = "https://www.gov.uk"  // To replace. It will be provided through another ticket
```
MaciejRewera, line 189: IMHO these two fields should be named based on where they point to. The name `ducrToMucrHintX` is good, but maybe somewhere deeper in the code, for e.g. fields in `AppConfig`.

app/utils/validators/forms/FieldValidator.scala
```diff
@@ -79,6 +79,8 @@ object FieldValidator {
 
   val isValidAddressField: String => Boolean = (input: String) => isAlphanumericWithSpecialCharacters(Set(' ', '\'', ',', '-'))(input)
 
+  val isValidMucr: String => Boolean = (input: String) => isAlphanumericWithSpecialCharacters(Set('\\', '-'))(input)
```
MaciejRewera, line 82: This seems incorrect to me. For e.g. in Movements we validate MUCR against a regex: `GB/[0-9A-Z]{3,4}-[0-9A-Z]{5,28}|GB/[0-9A-Z]{9,12}-[0-9A-Z]{1,23}|A:[0-9A-Z]{3}[0-9]{8}|C:[A-Z]{3}[0-9A-Z]{3,30}`. Why is it different here?

app/views/declaration/link_ducr_to_mucr.scala.html
```diff
@@ -0,0 +1,102 @@
...
+        ))
+    )
+}
+
+@buttons = @{
+  request.declarationType match {
+    case DeclarationType.CLEARANCE => saveButtons()
+    case _                         => saveAndContinue()
+  }
+}
```
MaciejRewera, line 72: Could you tell me why only on Clearance user is presented with 'Save for later' link/button?

app/controllers/declaration/LinkDucrToMucrController.scala
```diff
@@ -0,0 +1,77 @@
...
+  authenticate: AuthAction,
+  journeyType: JourneyAction,
+  override val exportsCacheService: ExportsCacheService,
+  navigator: Navigator,
+  mcc: MessagesControllerComponents,
+  linkDucrToMucrPage: link_ducr_to_mucr
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
+
+  private val yesNoForm = form()
```
MaciejRewera, line 45: Not sure is this is correct, as the same instance is shared between all calls. Probably the only reason we are getting away with this is because both `withSubmissionErrors()` and `bindFromRequest()` return a new instance.
Was very confusing to me and I had to check manually if it poses any danger.

app/controllers/navigation/Navigator.scala
```diff
@@ -83,7 +84,8 @@ object Navigator {
   val standard: PartialFunction[DeclarationPage, Mode => Call] = {
     case DeclarantDetails            => routes.AdditionalDeclarationTypeController.displayPage
     case ConsignmentReferences       => routes.DeclarantDetailsController.displayPage
-    case DeclarantIsExporter         => routes.ConsignmentReferencesController.displayPage
+    case LinkDucrToMucr              => routes.ConsignmentReferencesController.displayPage
+    case Mucr                        => routes.LinkDucrToMucrController.displayPage
```
MaciejRewera, line 88: Could all these be moved to `common` partial function?

docs/Customs Declare Exports AutoComplete.js
```diff
@@ -241,6 +241,20 @@ function consignmentReferences(){
...
+    }
+}
+
+function enterMucr() {
+    if (currentPageIs('/customs-declare-exports/declaration/enter-a-mucr')) {
+        document.getElementById('MUCR').value = 'CZYX123A';
+        document.getElementById('submit').click();
+    }
+}
+
```
MaciejRewera, line 257: Big thanks 👍

MaciejRewera (commented):
Not sure about the MUCR validation. Left some other comments as well.

app/controllers/navigation/Navigator.scala
udinobi: Good spot. Confused :confused:

app/controllers/declaration/LinkDucrToMucrController.scala
udinobi: :+1:

app/utils/validators/forms/FieldValidator.scala
udinobi: Wasn't aware. Ticket says nothing on this. Still, not sure the Movements regex is entirely correct (the colon IMO is wrong)

daveygit2050:
rebuild

app/controllers/navigation/Navigator.scala
MaciejRewera: May need to rework Navigator class to be more readable. It was fine at the beginning when journeys were mostly the same, because the specific parts were rather small. Now it's the opposite :/

MaciejRewera (approved):

glidester (approved):

PR #1199, opened 2021-05-19
[CEDS-3182] New transport payment option

glidester:
Making the express transport payment screen mandatory and
adding a new radio option of 'Payment information is
not available'

### DIFF
--- a/app/forms/declaration/TransportPayment.scala
+++ b/app/forms/declaration/TransportPayment.scala
@@ -16,14 +16,16 @@
 
 package forms.declaration
 import forms.DeclarationPage
+import forms.Mapping.requiredRadio
 import models.DeclarationType.DeclarationType
 import models.viewmodels.TariffContentKey
 import play.api.data.{Form, Mapping}
 import play.api.data.Forms.{mapping, optional, text}
+import play.api.data.validation.Constraints
 import play.api.libs.json.{Json, OFormat}
 import utils.validators.forms.FieldValidator.isContainedIn
 
-case class TransportPayment(paymentMethod: Option[String] = None)
+case class TransportPayment(paymentMethod: String)
 
 object TransportPayment extends DeclarationPage {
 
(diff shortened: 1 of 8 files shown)


app/forms/declaration/TransportPayment.scala
```diff
@@ -36,14 +38,14 @@ object TransportPayment extends DeclarationPage {
...
+  val validPaymentMethods = Set(cash, creditCard, cheque, other, eFunds, accHolder, notPrePaid, notAvailable)
 
   val formMapping: Mapping[TransportPayment] = mapping(
-    "paymentMethod" -> optional(
-      text()
-        .verifying("standard.transportDetails.paymentMethod.error", isContainedIn(validPaymentMethods))
-    )
+    "paymentMethod" ->
+      requiredRadio("declaration.transportInformation.transportPayment.paymentMethod.error.empty")
+        .verifying("declaration.transportInformation.transportPayment.paymentMethod.error.empty", isContainedIn(validPaymentMethods))
```
MaciejRewera, line 48: Are the error msg keys correct? They're both the same.

app/forms/declaration/TransportPayment.scala
```diff
@@ -36,14 +38,14 @@ object TransportPayment extends DeclarationPage {
   val eFunds = "H"
   val accHolder = "Y"
   val notPrePaid = "Z"
+  val notAvailable = "_"
```
MaciejRewera, line 41: `_` might cause problems when used in html forms. Have checked it is working and correct value is stored in DB?

glidester (commented):
ignore this comment

app/forms/declaration/TransportPayment.scala
glidester: Its a standard ASCII char so I don't see an issue. I just didn't want to use a A->Z character in case in future they decide to add another option with that value

glidester: AC's have been updated to store the '_' value and they all pass: https://github.com/hmrc/cds-exports-acceptance/pull/698/files#diff-d489df75c5973ab0eadec1a8b8c52e66a37ab7a21f6f24432f0ce9595fd84ed5

MaciejRewera (approved):

PR #1200, opened 2021-05-21
CEDS-3143 Amend url of Arrive-Depart config property

udinobi:


conf/application.conf
```diff
@@ -186,7 +186,7 @@ urls {
   goodsLocationCodesForDataElement = "https://www.gov.uk/government/collections/goods-location-codes-for-data-element-523-of-the-customs-declaration-service"
   tariffCdsChiefSupplement = "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/956071/Tariff_Volume_3_Great_Britain_Supplement_CDS_and_CHIEF_version_5.3__1_.odt"
   notesForMucrConsolidation = "https://www.gov.uk/government/publications/uk-trade-tariff-cds-volume-3-export-declaration-completion-guide/group-2-references-of-messages-document-certificates-and-authorisations#de-21-simplified-declaration-previous-documents-box-40-declaration-previous-document"
-  arriveOrDepartExportsService = "https://www.gov.uk"  // To replace. It will be provided through another ticket
+  arriveOrDepartExportsService = "/customs-movements"
```
MaciejRewera, line 189: Are you sure this will work? I'd expect it to be in a similar way to the line 161 and then overridden on environments.

udinobi: Ok. I'll use that one, without `/start` btw

udinobi: Yes it works, but I'll rename it as the property at line 161

conf/application.conf
```diff
@@ -158,7 +158,7 @@ urls {
   nactCodes = "https://www.gov.uk/guidance/national-additional-codes-for-data-element-617-of-cds"
   relevantLicenses = "https://www.gov.uk/starting-to-export/licences"
   serviceAvailability = "https://www.gov.uk/guidance/customs-declaration-service-service-availability-and-issues"
-  customsMovementsFrontend = "http://localhost:6796/customs-movements/start"
+  customsMovementsFrontend = "/customs-movements"
```
glidester, line 161: Think this link does needs to be as before (an absolute URL). When running the service locally, if you use this relative link it will not resolve because the movements service is listening on a different port number (6796). That is why for the local app.conf we have to make it an absolute url and then override the value in app-conf-base to make it a relative url for the MDTP environments.

udinobi: :+1:

glidester (approved):

PR #1201, opened 2021-05-21
[CEDS-3225]_Add_json_procedure_code_data_file

glidester:
Added new json file containing the code and
descriptions in each supported language.

Plus new connector class to 'pull' code lists
for required locale.

Also make missing config values fail fast

### DIFF
--- a/app/connectors/CodeListConnector.scala
+++ b/app/connectors/CodeListConnector.scala
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package connectors
+
+import akka.util.Helpers.Requiring
+import config.AppConfig
+import models.codes.{CommonCode, ProcedureCode}
+import play.api.libs.json.{Json, OFormat}
+import utils.JsonFile
+
+import java.util.Locale
+import java.util.Locale._
+import javax.inject.{Inject, Singleton}
+
+case class CodeItem(code: String, en: String, cy: String) {
+  def getDescriptionByLocale(locale: Locale): String =
+    locale.getLanguage match {
+      case "cy" => cy
+      case _    => en
+    }
+}
+
+object CodeItem {
+  implicit val formats: OFormat[CodeItem] = Json.format[CodeItem]
+}
+
+trait CodeListConnector {
+  def getProcedureCodes(locale: Locale): Seq[ProcedureCode]
+
+  val WELSH = new Locale("cy", "GB", "");
+}
+
+@Singleton
+class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeListConnector {
+
+  private implicit val supportedLanguages = Seq(ENGLISH, WELSH)
+
+  private val procedureCodeListsByLang = readCodesFromFile(
+    appConfig.procedureCodeListFile,
+    (codeItem: CodeItem, locale: Locale) => ProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
+  )
+
+  private val procedureCodeForC21ListsByLang = readCodesFromFile(
+    appConfig.procedureCodeForC21ListFile,
+    (codeItem: CodeItem, locale: Locale) => ProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
+  )
+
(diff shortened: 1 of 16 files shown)


app/connectors/CodeListConnector.scala
```diff
@@ -0,0 +1,76 @@
...
+class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeListConnector {
+
+  private implicit val supportedLanguages = Seq(ENGLISH, WELSH)
+
+  private val procedureCodeListsByLang = readCodesFromFile(
+    appConfig.procedureCodeListFile,
+    (codeItem: CodeItem, locale: Locale) => ProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
+  )
+
+  private val procedureCodeForC21ListsByLang = readCodesFromFile(
```
udinobi, line 57: This `val` is not used. Do you really need it?

app/connectors/CodeListConnector.scala
```diff
@@ -0,0 +1,76 @@
...
+    appConfig.procedureCodeListFile,
+    (codeItem: CodeItem, locale: Locale) => ProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
+  )
+
+  private def readCodesFromFile[T <: CommonCode](srcFile: String, factory: (CodeItem, Locale) => T)(implicit locals: Seq[Locale]) = {
+    val codeList = JsonFile.getJsonArrayFromFile(srcFile, CodeItem.formats)
+
+    locals.map { locale =>
```
udinobi, line 65: Why do not use `supportedLanguages` directly? This way you would not need to pass the implicit `locals`.

app/connectors/CodeListConnector.scala
```diff
@@ -0,0 +1,76 @@
...
+trait CodeListConnector {
+  def getProcedureCodes(locale: Locale): Seq[ProcedureCode]
+
+  val WELSH = new Locale("cy", "GB", "");
+}
+
+@Singleton
+class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeListConnector {
+
+  private implicit val supportedLanguages = Seq(ENGLISH, WELSH)
```
udinobi, line 50: Does `supportedLanguages` really need to be implicit?

app/connectors/CodeListConnector.scala
```diff
@@ -0,0 +1,76 @@
...
+      val procedureCodes = codeList.map(factory(_, locale))
+      (locale -> procedureCodes)
+    }.toMap
+  }
+
+  def getProcedureCodes(locale: Locale): Seq[ProcedureCode] =
+    procedureCodeListsByLang.getOrElse(locale, procedureCodeListsByLang.value.head._2)
+
+  def getProcedureCodesForC21(locale: Locale): Seq[ProcedureCode] =
+    procedureCodeListsByLang.getOrElse(locale, procedureCodeListsByLang.value.head._2)
```
udinobi, line 75: Maybe `procedureCodeForC21ListsByLang` instead of `procedureCodeListsByLang`?

app/connectors/CodeListConnector.scala
glidester: Ah, good spot. I do need it but the code that should use it was left pointing to the other val! Thanks

app/utils/JsonFile.scala
```diff
@@ -16,19 +16,44 @@
...
+    val jsonInputStream = getClass.getResourceAsStream(file)
+
+    Try(Json.parse(jsonInputStream)) match {
+      case Success(JsArray(jsValues)) =>
+        val items = jsValues.toList.map { jsValue =>
+          reader.reads(jsValue).asOpt
+        }
+
+        if (items.contains(None)) {
+          throw new IllegalArgumentException(s"One or more entries in JSON file: '$file' could not be parsed!")
```
udinobi, line 47: The error message is not clear (but it's just me :smile:). Why cannot the file be parsed?

app/connectors/CodeListConnector.scala
glidester: Yeah, this was why 'procedureCodeForC21ListsByLang' was not being used :)

app/connectors/CodeListConnector.scala
glidester: Just to make `readCodesFromFile` a pure function (no external influences)

app/utils/JsonFile.scala
glidester: Well each element in the array needs to be parsed and each element could have a different parsing issue. I did not think it worth while to try and collect and list every pare error as its an 'all or nothing' operation (either the whole file can be parsed or none of it will be taken).

app/connectors/CodeListConnector.scala
udinobi: But `supportedLanguages` is a constant value. It has no influence on what `readCodesFromFile` returns for the same set of the input parameters. I mean... the 'purity', in terms of functional programming, does not depend on constant values, but on the fact that the function should always returns the same value for the same input parameters.

app/utils/JsonFile.scala
udinobi: tbh I was just asking to make the error message a bit more readable. Something like
`'$file' could not be parsed due to one or more invalid JSON attributes` or similar

udinobi (approved):

PR #1206, opened 2021-05-26
[CEDS-3228] Add additional procedure codes list

glidester:
Add list of all additional procedure codes (both
for standard and C21 journeys) loaded from json
files.

Also add association link logic files to link
certain procedure codes with a list of valid
additional procedure codes.

Adding in a service class to contain the business
logic of what procedure codes lists and what
additional procedure code lists should be used in
certain circumstances.


app/connectors/CodeLinkConnector.scala
```diff
@@ -0,0 +1,56 @@
...
+
+trait CodeLinkConnector {
+  def getValidAdditionalProcedureCodesForProcedureCode(procedureCode: String): Option[Seq[String]]
+  def getValidAdditionalProcedureCodesForProcedureCodeC21(procedureCode: String): Option[Seq[String]]
+}
+
+@Singleton
+class FileBasedCodeLinkConnector @Inject()(appConfig: AppConfig) extends CodeLinkConnector { //codeListConnector: CodeListConnector,
+
+  private def readCodeLinksFromFile[T <: CommonCode](srcFile: String) = {
```
udinobi, line 40: I would love to see what type this `def` is returning. Even to have a better understanding at the call-sites for this `def`.

udinobi (approved):

PR #1214, opened 2021-06-10
[CEDS-3198] Additional Procedure Codes page development

glidester:
Swap free text enty of Additional Procedure Code to a drop-down
selection box of only valid codes. Plus content changes

### DIFF
--- a/app/connectors/CodeListConnector.scala
+++ b/app/connectors/CodeListConnector.scala
@@ -26,6 +26,7 @@ import utils.JsonFile
 import java.util.Locale
 import java.util.Locale._
 import javax.inject.{Inject, Singleton}
+import scala.collection.immutable.ListMap
 
 case class CodeItem(code: String, en: String, cy: String) {
   def getDescriptionByLocale(locale: Locale): String =
@@ -41,10 +42,10 @@ object CodeItem {
 
 @ImplementedBy(classOf[FileBasedCodeListConnector])
 trait CodeListConnector {
-  def getProcedureCodes(locale: Locale): Seq[ProcedureCode]
-  def getProcedureCodesForC21(locale: Locale): Seq[ProcedureCode]
-  def getAdditionalProcedureCodes(locale: Locale): Seq[AdditionalProcedureCode]
-  def getAdditionalProcedureCodesForC21(locale: Locale): Seq[AdditionalProcedureCode]
+  def getProcedureCodes(locale: Locale): Map[String, ProcedureCode]
+  def getProcedureCodesForC21(locale: Locale): Map[String, ProcedureCode]
+  def getAdditionalProcedureCodesMap(locale: Locale): Map[String, AdditionalProcedureCode]
+  def getAdditionalProcedureCodesMapForC21(locale: Locale): Map[String, AdditionalProcedureCode]
 
   val WELSH = new Locale("cy", "GB", "");
 }
@@ -54,44 +55,49 @@ class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeLis
 
   private implicit val supportedLanguages = Seq(ENGLISH, WELSH)
 
-  private def readCodesFromFile[T <: CommonCode](srcFile: String, factory: (CodeItem, Locale) => T)(implicit locals: Seq[Locale]) = {
+  private def readCodesMapFromFile[T <: CommonCode](srcFile: String, factory: (CodeItem, Locale) => T) = {
     val codeList = JsonFile.getJsonArrayFromFile(srcFile, CodeItem.formats)
 
-    locals.map { locale =>
-      val procedureCodes = codeList.map(factory(_, locale))
-      (locale -> procedureCodes)
-    }.toMap
+    val langCodes = supportedLanguages.map { locale =>
+      val procedureCodes = codeList
+        .map(factory(_, locale))
+        .map(cc => (cc.code, cc))
+
+      (locale.getLanguage -> ListMap(procedureCodes: _*))
+    }
+
+    ListMap(langCodes: _*)
   }
 
-  private val procedureCodeListsByLang = readCodesFromFile(
+  private val procedureCodeListsByLang = readCodesMapFromFile(
     appConfig.procedureCodesListFile,
     (codeItem: CodeItem, locale: Locale) => ProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
   )
 
-  private val procedureCodeForC21ListsByLang = readCodesFromFile(
+  private val procedureCodeForC21ListsByLang = readCodesMapFromFile(
     appConfig.procedureCodesForC21ListFile,
     (codeItem: CodeItem, locale: Locale) => ProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
   )
 
-  private val additionalProcedureCodeListsByLang = readCodesFromFile(
+  private val additionalProcedureCodeMapsByLang = readCodesMapFromFile(
     appConfig.additionalProcedureCodes,
     (codeItem: CodeItem, locale: Locale) => AdditionalProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
   )
 
-  private val additionalProcedureCodeForC21ListsByLang = readCodesFromFile(
+  private val additionalProcedureCodeForC21MapsByLang = readCodesMapFromFile(
     appConfig.additionalProcedureCodesForC21,
     (codeItem: CodeItem, locale: Locale) => AdditionalProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
   )
 
-  def getProcedureCodes(locale: Locale): Seq[ProcedureCode] =
-    procedureCodeListsByLang.getOrElse(locale, procedureCodeListsByLang.value.head._2)
+  def getProcedureCodes(locale: Locale): Map[String, ProcedureCode] =
+    procedureCodeListsByLang.getOrElse(locale.getLanguage, procedureCodeListsByLang.value.head._2)
 
-  def getProcedureCodesForC21(locale: Locale): Seq[ProcedureCode] =
-    procedureCodeForC21ListsByLang.getOrElse(locale, procedureCodeForC21ListsByLang.value.head._2)
+  def getProcedureCodesForC21(locale: Locale): Map[String, ProcedureCode] =
+    procedureCodeForC21ListsByLang.getOrElse(locale.getLanguage, procedureCodeForC21ListsByLang.value.head._2)
 
-  def getAdditionalProcedureCodes(locale: Locale): Seq[AdditionalProcedureCode] =
-    additionalProcedureCodeListsByLang.getOrElse(locale, additionalProcedureCodeListsByLang.value.head._2)
+  def getAdditionalProcedureCodesMap(locale: Locale): Map[String, AdditionalProcedureCode] =
+    additionalProcedureCodeMapsByLang.getOrElse(locale.getLanguage, additionalProcedureCodeMapsByLang.value.head._2)
 
-  def getAdditionalProcedureCodesForC21(locale: Locale): Seq[AdditionalProcedureCode] =
-    additionalProcedureCodeForC21ListsByLang.getOrElse(locale, additionalProcedureCodeForC21ListsByLang.value.head._2)
+  def getAdditionalProcedureCodesMapForC21(locale: Locale): Map[String, AdditionalProcedureCode] =
+    additionalProcedureCodeForC21MapsByLang.getOrElse(locale.getLanguage, additionalProcedureCodeForC21MapsByLang.value.head._2)
 }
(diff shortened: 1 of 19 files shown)


app/connectors/CodeListConnector.scala
```diff
@@ -59,7 +59,20 @@ class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeLis
 
     locals.map { locale =>
       val procedureCodes = codeList.map(factory(_, locale))
-      (locale -> procedureCodes)
+      (locale.getLanguage -> procedureCodes)
+    }.toMap
+  }
+
+  private def readCodesMapFromFile[T <: CommonCode](srcFile: String, factory: (CodeItem, Locale) => T)(implicit locals: Seq[Locale]) = {
```
udinobi, line 66: I hope you will consider to replace `readCodesFromFile` and `readCodesMapFromFile` with `loadCodesAsOrderedMap` (in my PR) as it satisfies our requirements, ordering and lookup.

udinobi, line 66: IMHO `implicit locals: Seq[Locale]` is not useful at all. Why don't you directly use `supportedLanguages`?

app/connectors/CodeListConnector.scala
```diff
@@ -73,25 +86,25 @@ class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeLis
...
+    procedureCodeListsByLang.getOrElse(locale.getLanguage, procedureCodeListsByLang.value.head._2)
 
   def getProcedureCodesForC21(locale: Locale): Seq[ProcedureCode] =
-    procedureCodeForC21ListsByLang.getOrElse(locale, procedureCodeForC21ListsByLang.value.head._2)
+    procedureCodeForC21ListsByLang.getOrElse(locale.getLanguage, procedureCodeForC21ListsByLang.value.head._2)
 
-  def getAdditionalProcedureCodes(locale: Locale): Seq[AdditionalProcedureCode] =
-    additionalProcedureCodeListsByLang.getOrElse(locale, additionalProcedureCodeListsByLang.value.head._2)
+  def getAdditionalProcedureCodesMap(locale: Locale): Map[String, AdditionalProcedureCode] =
+    additionalProcedureCodeMapsByLang.getOrElse(locale.getLanguage, additionalProcedureCodeMapsByLang.value.head._2)
```
udinobi, line 106: The `def` could potentially return not what you expect if the provided `locale` is not found. I think we should return the English map in that case and being this a map we cannot be sure the `head` is the English one and not the Welsh for instance. Am I wrong?

app/connectors/CodeListConnector.scala
```diff
@@ -73,25 +86,25 @@ class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeLis
...
 
-  def getAdditionalProcedureCodesForC21(locale: Locale): Seq[AdditionalProcedureCode] =
-    additionalProcedureCodeForC21ListsByLang.getOrElse(locale, additionalProcedureCodeForC21ListsByLang.value.head._2)
+  def getAdditionalProcedureCodesMapForC21(locale: Locale): Map[String, AdditionalProcedureCode] =
+    additionalProcedureCodeForC21MapsByLang.getOrElse(locale.getLanguage, additionalProcedureCodeForC21MapsByLang.value.head._2)
```
udinobi, line 109: Same here

app/connectors/CodeListConnector.scala
glidester: Yeah, I suppose as they are private methods we don't need to pass them as a param. Changed

app/connectors/CodeListConnector.scala
```diff
@@ -73,25 +86,25 @@ class FileBasedCodeListConnector @Inject()(appConfig: AppConfig) extends CodeLis
...
     (codeItem: CodeItem, locale: Locale) => AdditionalProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
   )
 
-  private val additionalProcedureCodeForC21ListsByLang = readCodesFromFile(
+  private val additionalProcedureCodeForC21MapsByLang = readCodesMapFromFile(
     appConfig.additionalProcedureCodesForC21,
     (codeItem: CodeItem, locale: Locale) => AdditionalProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
   )
 
   def getProcedureCodes(locale: Locale): Seq[ProcedureCode] =
```
udinobi, line 99: Personal opinion... :smiley_cat: 
`procedureCodeListsByLang`, `procedureCodeForC21ListsByLang`,  `getProcedureCodes` and `getProcedureCodesForC21` should be placed in `ProcedureCodeService`

app/connectors/CodeListConnector.scala
glidester: Yes good spot. Ok altered readCodesMapFromFile to return a ListMap to ensure the order is same as supportedLanguages :+1:

app/connectors/CodeListConnector.scala
glidester: As above

app/services/ProcedureCodeService.scala
```diff
@@ -19,33 +19,49 @@ package services
...
 
   def getProcedureCodesFor(journey: DeclarationType, isEidr: Boolean, locale: Locale): Seq[ProcedureCode] =
     journey match {
       case CLEARANCE if !isEidr => codeListConnector.getProcedureCodesForC21(locale)
       case _                    => codeListConnector.getProcedureCodes(locale)
     }
 
+  def getProcedureCodeFor(procedureCode: String, journey: DeclarationType, isEidr: Boolean, locale: Locale): Option[ProcedureCode] =
+    (journey match {
+      case CLEARANCE if !isEidr => codeListConnector.getProcedureCodesForC21(locale).filter(_.code == procedureCode)
```
udinobi, line 37: Why not using a Map for the lookup?

app/services/ProcedureCodeService.scala
```diff
@@ -19,33 +19,49 @@ package services
...
+      case _                    => codeListConnector.getProcedureCodes(locale).filter(_.code == procedureCode)
```
udinobi, line 38: Same here

app/connectors/CodeListConnector.scala
glidester: The CodeListConnector was designed to mimic the type of functionality you would get from a connector that looked up values via a DB or a REST service. The maps are just an efficient way of preventing continually going back to read the json files for each call. I do not see any benefit in creating a service class for each code type, there is no 'business logic' that needs encapsulating in such a class. The whole idea is that in future we could simply replace the `FileBasedCodeListConnector` with an implementation that retrieved the codes from a REST service. If we created these service classes to hold the maps we would have to alter these service classes too (as we probably then would not want to cache the values, or at least not in perpetuity).

app/services/view/AutoCompleteItem.scala
```diff
@@ -16,14 +16,16 @@
...
-import models.codes.ProcedureCode
+import models.codes.{AdditionalProcedureCode, ProcedureCode}
 import services.model.{Country, CustomsOffice, OfficeOfExit, PackageType}
 import services.{DocumentType, HolderOfAuthorisationCode}
 
 case class AutoCompleteItem(label: String, value: String)
 
 object AutoCompleteItem {
 
+  def formatProcedureCode(pc: ProcedureCode) = s"${pc.code} - ${pc.description}"
```
udinobi, line 27: Personal opinion :smiling_face_with_three_hearts: 
`formatProcedureCode` should be in `ProcedureCodeService`.

app/services/view/AutoCompleteItem.scala
```diff
@@ -43,5 +45,8 @@ object AutoCompleteItem {
     offices map (d => AutoCompleteItem(s"${d.description} - ${d.code}", d.code))
 
   def fromProcedureCodes(procedureCodes: List[ProcedureCode]): List[AutoCompleteItem] =
-    procedureCodes map (pc => AutoCompleteItem(s"${pc.code} - ${pc.description}", pc.code))
+    procedureCodes map (pc => AutoCompleteItem(formatProcedureCode(pc), pc.code))
+
+  def fromAdditionalProcedureCodes(additionalProcedureCodes: List[AdditionalProcedureCode]): List[AutoCompleteItem] =
```
udinobi, line 50: Personal opinion :smiling_face_with_three_hearts: 
`fromProcedureCodes` and `fromAdditionalProcedureCodes` should be in their respective service class.

app/views/components/inputs/input_accessible_autocomplete.scala.html
```diff
@@ -40,14 +45,22 @@
...
-    <select class="govuk-form-group" id="@(fieldNameWithoutBrackets)" name="@field.name" value="@field.value" data-gov-select-autocomplete="true">
-        <option value="">@emptySelectValue</option>
-        @items.map { i =>
-            <option value="@i.value" @if(selectedValue == i.value) {selected="selected"}>@i.label</option>
-        }
-    </select>
+<div class="govuk-grid-row">
+    <div class="@rightElement.map(_ => "govuk-grid-column-two-thirds").getOrElse("govuk-grid-column-full")">
+        <div id="@(fieldNameWithoutBrackets)-container" class="autocomplete__wrapper @widthClass">
+            <select class="govuk-form-group" id="@(fieldNameWithoutBrackets)" name="@field.name" value="@field.value" data-gov-select-autocomplete="true">
```
udinobi, line 51: Not sure about having the same `id` for two different elements, even if they are parent-child. But maybe it does not harm.

app/services/ProcedureCodeService.scala
glidester: Because we are dealing with a small list of values so I did not see the need for a Map

app/services/ProcedureCodeService.scala
glidester: As above

app/services/view/AutoCompleteItem.scala
glidester: I strongly disagree. This format is a UI level responsibility, not a data retrieval level one.

app/services/view/AutoCompleteItem.scala
glidester: As already stated, service classes in this use case adds no value

app/views/components/inputs/input_accessible_autocomplete.scala.html
glidester: They are not the same, one is "@(fieldNameWithoutBrackets)-container" and the inner element is just "@(fieldNameWithoutBrackets)"

app/connectors/CodeListConnector.scala
glidester: Not sure how this is relevant to this PR?

app/connectors/CodeListConnector.scala
udinobi: I do not understand. The `Map`s should be placed in the service classes, not in the connector (irrespective if the loading take place by REST or from the DB). No need to call the connector any more once the codes are retrieved. The service classes will not require any change if we change the connector's implementation, because the service class uses the interface provided by the trait.

udinobi: In my PR I call `CodeListConnector` only once.

glidester: We should not be discussing what is in your PR as part of this PR.

app/services/ProcedureCodeService.scala
udinobi: My suggestion to use a `Map` instead of a `List` was to simplify `CodeListConnector` trait, which could have then only one signature. And using a `Map` would not harm. O(1) vs O(n) for lookup.

app/services/view/AutoCompleteItem.scala
udinobi: But `ProcedureCodeService` is not a "data retrieval" class. It's a **service** class only used for UI purpose. And in fact it should be then placed under /services/view.

app/views/components/inputs/input_accessible_autocomplete.scala.html
udinobi: You are right. I missed `-container` :+1:

app/connectors/CodeListConnector.scala
udinobi: I was using my PR just as an example simply to refute your claim
>  If we created these service classes to hold the maps we would have to alter these service classes too

app/services/ProcedureCodeService.scala
glidester: Ok, fair enough. As it also removes the need for a separate `readCodesFromFile` method in the connector I'll make this change :+1:

udinobi (approved):
Approved, even if I still think that `CodeListConnector` as a trait should provide a single signature only, loading the data (from file or from the network), while all code-list handling should be moved to specific "**UI service**" classes.

app/services/ProcedureCodeService.scala
udinobi: Indeed. That was what I meant :+1:

app/connectors/CodeListConnector.scala
glidester: If this connector was not a file based connector (and was instead a DB or REST based connector) then you absolutely would have to call the connector more than once (otherwise what happens if the code list changes!). By caching the values in the service class you are making assumptions about the behavior/implementation of the connector class in the service class (which you should not do). The implementation of the connector class should have no bearing/influence on the service class. The service class should only hold the business logic (like do I need the C21 or standard list).

udinobi (approved):

PR #1215, opened 2021-06-11
CEDS-3207 Refactor Authorisation codes

udinobi:
I really think we are using a bad design pattern for handling our "codes", where all of them are handled by a single big class.

I would go instead with different **small** classes, connectors or file loaders, called when needed by **UI** service classes for each object, because "Holder of Authorisation codes" are not "Procedure codes", and so on for all other "codes", even if they have the same structure.

I really dislike this idea of a class handling multiple business objects that have nothing in common but structure.
I think this `CommonCode` is a really bad idea, it places constraints and decreases flexibility.


app/connectors/CodeListConnector.scala
```diff
@@ -42,62 +42,73 @@ object CodeItem {
...
+  type CodeMap[T <: CommonCode] = Map[String, ListMap[String, T]]
+
+  def getHolderOfAuthorisationCodes(locale: Locale): ListMap[String, HolderOfAuthorisationCode]
+  def getProcedureCodes(locale: Locale): ListMap[String, ProcedureCode]
+  def getProcedureCodesForC21(locale: Locale): ListMap[String, ProcedureCode]
+  def getAdditionalProcedureCodesMap(locale: Locale): ListMap[String, AdditionalProcedureCode]
+  def getAdditionalProcedureCodesMapForC21(locale: Locale): ListMap[String, AdditionalProcedureCode]
 
   val WELSH = new Locale("cy", "GB", "");
+  val supportedLanguages = Seq(ENGLISH, WELSH)
```
glidester, line 55: Which languages an implementation of CodeListConnector supports is an implementation detail, does not belong in the trait IMO

app/connectors/CodeListConnector.scala
```diff
@@ -42,62 +42,73 @@ object CodeItem {
...
   )
 
-  private val additionalProcedureCodeMapsByLang = readCodesMapFromFile(
+  private lazy val additionalProcedureCodeMapsByLang = loadCommonCodesAsOrderedMap(
     appConfig.additionalProcedureCodes,
     (codeItem: CodeItem, locale: Locale) => AdditionalProcedureCode(codeItem.code, codeItem.getDescriptionByLocale(locale))
   )
 
-  private val additionalProcedureCodeForC21MapsByLang = readCodesMapFromFile(
+  private lazy val additionalProcedureCodeForC21MapsByLang = loadCommonCodesAsOrderedMap(
```
glidester, line 95: Don't use lazy keyword. These maps need to be loaded on app startup so that they fail fast if the data files (that get read to populate them) are not found for any reason.

app/services/view/HolderOfAuthorisationCodes.scala
```diff
@@ -0,0 +1,35 @@
...
+@Singleton
+class HolderOfAuthorisationCodes @Inject()(codeListConnector: CodeListConnector) {
+
+  private def description(h: HolderOfAuthorisationCode): String = s"${h.code} - ${h.description}"
+
+  def getCodeDescription(locale: Locale, code: String): String =
+    codeListConnector.getHolderOfAuthorisationCodes(locale).get(code).fold("")(description)
+
+  def asListOfAutoCompleteItems(locale: Locale): List[AutoCompleteItem] =
+    codeListConnector.getHolderOfAuthorisationCodes(locale).values.map(h => AutoCompleteItem(description(h), h.code)).toList
```
glidester, line 34: I don't agree with moving the responsibility of creating `AutoCompleteItem`s out of the existing `AutoCompleteItem` class. The method `asListOfAutoCompleteItems` is unnecessarily coupling this service class to a UI form element IMO.

test/connectors/CodeListConnectorSpec.scala
```diff
@@ -30,107 +31,157 @@ class CodeListConnectorSpec extends ConnectorSpec {
...
 
-  private val sampleAPCsEnglish = Map(
+  private val sampleAPCsEnglish = ListMap(
     "001" -> AdditionalProcedureCode("001", "English"),
     "002" -> AdditionalProcedureCode("002", "English"),
     "003" -> AdditionalProcedureCode("003", "English")
   )
 
-  private val sampleAPCsWelsh = Map(
+  private val sampleAPCsWelsh = ListMap(
```
glidester, line 170: Good catch, thank you :+1:

app/connectors/CodeListConnector.scala
udinobi: I don't agree. We are loading the codes according to a specific order (by using a `ListMap`), and this is why I also changed the return value from `Map` to `ListMap`, so it does not harm to let clients know which is the order. In addition we expect a `Locale` as input parameter so I do not see why we should not tell to clients what languages we support. As well as if we want to test the generic interface (the trait) and not the implementation.

app/connectors/CodeListConnector.scala
udinobi: I'll remove the lazy but that's another reason why I strongly disagree with this implementation. I cannot test "Holder of Authorisation codes" in isolation because if the vals are not lazy they get all loaded when the class is instantiated.

I would go instead with different **small** classes, connectors or file loaders, called when needed by **UI** service classes for each object, because "Holder of Authorisation codes" are not "Procedure codes", and so on for all other "codes", even if they have the same structure.

I really dislike this idea of a class handling multiple business objects that have nothing in common but structure.
I think this `CommonCode` is a really bad idea, it places constraints and decreases flexibility.

app/services/view/HolderOfAuthorisationCodes.scala
udinobi: In turn, I disagree. Please, read my comment. This service class is a **UI** service class like `AutoCompleteItems` but it only has the responsability for `HolderOfAuthorisationCodes` and not for any other codes like instead has `AutoCompleteItems` which follows in my opinion a really bad design.

test/services/view/HolderOfAuthorisationCodesSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+  private lazy val codeListConnector = new FileBasedCodeListConnector(appConfig)
+
+  private lazy val holderOfAuthorisationCodes = new HolderOfAuthorisationCodes(codeListConnector)
+
+  override def beforeEach(): Unit = {
+    super.beforeEach()
+
+    reset(appConfig)
+    when(appConfig.holderOfAuthorisationCodes).thenReturn("/code-lists/holderOfAuthorisationCodes.json")
+    when(appConfig.procedureCodesListFile).thenReturn("/code-lists/manyCodes.json")
```
udinobi, line 40: Why am I forced here to also provide the config for "Procedure codes" and "Additional Procedure codes" when what I want is just to test in isolation "Holder of Authorisation codes" ?

glidester: Well if you mocked FileBasedCodeListConnector rather than use the real connector class you could test this service class in isolation. It is the same deal as we have with the AppConfigSpec. As we want them to fail fast then we must provide them with all their requirements when harnessing them for tests.

app/connectors/CodeListConnector.scala
glidester: But the loading of codes according to a specific language order of is an implementation detail of the `FileBasedCodeListConnector`. Other implementations of this trait may not be so constrained and should not concern themselves with it. The trait is defining a contract of 'if I ask for these codes in a specified language, give me an ordered index of these code entities'.  I won't ask you to modify it but I disagree with this change.

test/services/view/HolderOfAuthorisationCodesSpec.scala
udinobi: With "in isolation" I meant not coupled with the other "codes", not testing `HolderOfAuthorisationCodesSpec` in isolation. In this test I want to use the real connector because I want to test if the "Holder of Authorisation" codes are loaded ordered as per requirements, and not using fake but the real values. AppConfigSpec is not useful at all all from this point of view because it is only using fake values.

glidester: I'd suggest not using the real codes file to test that the order is maintained, otherwise if the real codes file changes order in the future the tests will break (even though the code is actually still fine).

udinobi: But that's what I want, I want the test fails if the actual result is different from what is expected from the loader. Especially with the current file-based implementation, which in case of changes will however require a new deployment. With a future, if it will ever happen :smirk: , network-based implementation the real values will be actually not so relevant any more.

glidester (approved):
Approved with our differences documented in comments :slightly_smiling_face:

PR #1216, opened 2021-06-11
[CEDS-2961] Content update of the Additional Document Codes page

glidester:
Adding two new expanders, new content and some dynamic
content (with dynamic links to tariff guide).


app/controllers/declaration/CusCodeController.scala
```diff
@@ -78,3 +78,10 @@ class CusCodeController @Inject()(
       controllers.declaration.routes.TaricCodeSummaryController.displayPage(_, itemId)
     }
 }
+
+object CusCodeController {
```
udinobi, line 82: IMO this object should be placed under `controllers/utils`

glidester: Ok, I've created a new generic `ExportsDecModelHelper` controller util class to hold this method.

udinobi (approved):

PR #1219, opened 2021-06-16
CEDS-3244 Upgrade to Play 2.8

udinobi:


test/connectors/CodeLinkConnectorISpec.scala
```diff
@@ -19,7 +19,7 @@ package connectors
 import config.AppConfig
 import org.mockito.Mockito.{reset, when}
 
-class CodeLinkConnectorSpec extends ConnectorSpec {
+class CodeLinkConnectorISpec extends ConnectorISpec {
```
glidester, line 22: Why rename to ISpec? This isn't really an integration test, its part of the unit tests.

udinobi: But it extends ConnectorISpec which in turn extends `GuiceOneAppPerSuite` and `WiremockTestServer` which are used for ITs

glidester, line 22: This is not really an integration test (no use of wiremock) so I'd leave out the "ISpec" postfix.

test/connectors/CodeListConnectorISpec.scala
```diff
@@ -23,7 +23,7 @@ import java.util.Locale.{ENGLISH, JAPANESE}
 
 import scala.collection.immutable.ListMap
 
-class CodeListConnectorSpec extends ConnectorSpec {
+class CodeListConnectorISpec extends ConnectorISpec {
```
glidester, line 26: This is not really an integration test (no use of wiremock) so I'd leave out the "ISpec" postfix.

test/connectors/exchange/ExportsDeclarationExchangeSpec.scala
```diff
@@ -23,12 +23,12 @@ import forms.declaration.additionaldeclarationtype.AdditionalDeclarationType.Add
...
 import models.declaration._
 import models.{DeclarationStatus, DeclarationType, ExportsDeclaration}
-import org.scalatest.{Matchers, WordSpec}
+import base.UnitSpec
 import org.scalatestplus.mockito.MockitoSugar
 import play.api.libs.json.{JsSuccess, Json}
 import services.cache.ExportsDeclarationBuilder
 
-class ExportsDeclarationExchangeSpec extends WordSpec with Matchers with ExportsDeclarationBuilder with MockitoSugar {
+class ExportsDeclarationExchangeSpec extends UnitSpec with ExportsDeclarationBuilder with MockitoSugar {
```
glidester, line 31: Shouldn't this also be UnitWithMocksSpec

udinobi: Good spot :+1:

test/connectors/CodeLinkConnectorISpec.scala
udinobi: I'll change then the extended trait.

test/connectors/CodeListConnectorISpec.scala
udinobi: I'll change then the extended trait.

glidester (approved):

PR #1222, opened 2021-06-17
[CEDS-3203] Fix validation not applying on save and continue

glidester:
000 validation was only being applied when the user pressed the
"Add" button. This was due to two different validation routines
depending on which button was used to submit the form.

Now implemented single validation routine used for all form submit
buttons on page. Also updated tests to include the "Save for later"
form action.

### DIFF
--- a/app/controllers/declaration/AdditionalProcedureCodesController.scala
+++ b/app/controllers/declaration/AdditionalProcedureCodesController.scala
@@ -16,21 +16,18 @@
 
 package controllers.declaration
 
-import scala.concurrent.{ExecutionContext, Future}
-
 import controllers.actions.{AuthAction, JourneyAction}
 import controllers.navigation.Navigator
-import controllers.util.MultipleItemsHelper.remove
 import controllers.util._
+import controllers.util.MultipleItemsHelper.remove
 import forms.declaration.procedurecodes.AdditionalProcedureCode
 import forms.declaration.procedurecodes.AdditionalProcedureCode._
-import javax.inject.Inject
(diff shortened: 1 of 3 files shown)


app/controllers/declaration/AdditionalProcedureCodesController.scala
```diff
@@ -76,10 +76,15 @@ class AdditionalProcedureCodesController @Inject()(
...
       formAction match {
-        case Add if !boundForm.hasErrors =>
-          addAnotherCodeHandler(mode, itemId, boundForm.get, procedureCode, cachedData, validAdditionalProcedureCodes)
-        case SaveAndContinue | SaveAndReturn if !boundForm.hasErrors =>
-          saveAndContinueHandler(mode, itemId, boundForm.get, procedureCode, cachedData, validAdditionalProcedureCodes)
+        case Add | SaveAndContinue | SaveAndReturn if !boundForm.hasErrors =>
+          validateForm(
+            formAction,
+            mode,
+            returnErrorPage(mode, itemId, boundForm.get, procedureCode, cachedData.additionalProcedureCodes, validAdditionalProcedureCodes),
```
udinobi, line 83: Nit: assign `returnErrorPage` to val for better readibility

udinobi (approved):

app/controllers/declaration/AdditionalProcedureCodesController.scala
glidester: Good idea :+1:

udinobi (approved):

PR #1226, opened 2021-06-22
CEDS-3246 Conditionally skip page at /supervising-customs-office

udinobi:
Skip page at /supervising-customs-office if all declaration's items have '1040' as Procedure code and '000' as unique APC

### DIFF
--- a/app/controllers/declaration/AdditionalProcedureCodesController.scala
+++ b/app/controllers/declaration/AdditionalProcedureCodesController.scala
@@ -54,16 +54,16 @@ class AdditionalProcedureCodesController @Inject()(
   private val emptyProcedureCodesData = ProcedureCodesData(None, Seq())
 
   def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
-    val frm = form().withSubmissionErrors()
-
     val (maybeCachedProcedureCode, cachedData) = getCachedData(itemId)
     val availableAdditionalProcedureCodes = getAvailableAdditionalProcedureCodes(maybeCachedProcedureCode)
 
     (maybeCachedProcedureCode, availableAdditionalProcedureCodes) match {
       case (Some(procedureCode), Some(validAdditionalProcedureCodes)) if !validAdditionalProcedureCodes.isEmpty =>
+        val frm = form().withSubmissionErrors()
         Ok(additionalProcedureCodesPage(mode, itemId, frm, procedureCode, validAdditionalProcedureCodes, cachedData.additionalProcedureCodes))
+
       case _ =>
-        Redirect(controllers.declaration.routes.ProcedureCodesController.displayPage(mode, itemId))
+        Redirect(routes.ProcedureCodesController.displayPage(mode, itemId))
     }
   }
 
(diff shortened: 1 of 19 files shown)


app/controllers/declaration/AdditionalProcedureCodesController.scala
```diff
@@ -183,14 +191,22 @@ class AdditionalProcedureCodesController @Inject()(
...
-        navigator.continueTo(mode, routes.FiscalInformationController.displayPage(_, itemId))
-      case _ => navigator.continueTo(mode, routes.CommodityDetailsController.displayPage(_, itemId))
+        Future.successful(navigator.continueTo(mode, routes.FiscalInformationController.displayPage(_, itemId)))
+
+      case _ =>
+        val continueToCommodityDetails = navigator.continueTo(mode, routes.CommodityDetailsController.displayPage(_, itemId))
+
+        val model = maybeModel.fold(request.cacheModel)(identity)
+        if (SupervisingCustomsOfficeHelper.isConditionForAllProcedureCodesNotVerified(model)) Future.successful(continueToCommodityDetails)
+        else SupervisingCustomsOfficeHelper.resetCache(this).map(_ => continueToCommodityDetails)
```
glidester, line 209: Would this logic not be better placed inside the `updateCache` method (it has no effect on which page we continue to)?

app/controllers/util/SupervisingCustomsOfficeHelper.scala
```diff
@@ -0,0 +1,58 @@
...
+  def nextPage(implicit request: JourneyRequest[_]): Mode => Call =
+    request.declarationType match {
+      case DeclarationType.SUPPLEMENTARY | DeclarationType.STANDARD => routes.InlandTransportDetailsController.displayPage
+      case DeclarationType.SIMPLIFIED | DeclarationType.OCCASIONAL  => routes.ExpressConsignmentController.displayPage
+      case DeclarationType.CLEARANCE                                => routes.DepartureTransportController.displayPage
+    }
+
+  def resetCache(modelCacheable: ModelCacheable)(implicit hc: HeaderCarrier, request: JourneyRequest[_]): Future[Option[ExportsDeclaration]] =
+    modelCacheable.updateExportsDeclarationSyncDirect { model =>
+      model.copy(locations = model.locations.copy(supervisingCustomsOffice = None))
```
glidester, line 56: Think this would be better placed inside the `ExportsDeclaration` case class (similar to the existing `removeCountryOfRouting` method).

app/controllers/declaration/AdditionalProcedureCodesController.scala
udinobi: `updateCache` is not the right place because it is called for every `ExportItem`.
The condition for resetting `supervisingCustomsOffice` is instead that **ALL** `ExportItem`s have 1040-000 as PC and APC.
I could end-up resetting the value just after the 1st `ExportItem` with 1040-000. A 2nd one could not have 1040-000 as PC-APC and in that case the condition will not be verified.

app/controllers/util/SupervisingCustomsOfficeHelper.scala
udinobi: :+1:

app/controllers/declaration/AdditionalProcedureCodesController.scala
glidester: Ah yes, you mean it is called for every new APC that gets added/removed (before continuing to the next page) :+1:

glidester (approved):

PR #1231, opened 2021-06-25
BAU Fix caching bug

MaciejRewera:
The bug was appearing when user viewed their
submitted Declaration and then moved back and
started a new Declaration. The viewed one was
still in the cache and they were actually
trying to change a Declaration that was already
submitted. Since this is not possible, user
would see "There is a problem with service" page.

The bug was present because of the way the data
for declaration viewing page was taken from cache.
Declaration ID was added to session and based on
that cache data was retrieved. But this ID was
never cleared from session and therefore users
could get stuck, not being able to create a new
declaration.

This fix changes the way cache is being retrieved
by passing the declaration ID in URL parameter
instead of session. The difference to the user
is that this declaration id is visible in the
browser. But this should not be a massive problem
since we have this id present in some other pages.


app/views/declaration/summary/item_section.scala.html
```diff
@@ -257,7 +257,8 @@
             ))
         )
     ),
-    if(journeyRequest.isType(DeclarationType.STANDARD, DeclarationType.SUPPLEMENTARY, DeclarationType.SIMPLIFIED, DeclarationType.OCCASIONAL))
+    if(Seq(DeclarationType.STANDARD, DeclarationType.SUPPLEMENTARY, DeclarationType.SIMPLIFIED, DeclarationType.OCCASIONAL).contains(declarationType))
+//    if(journeyRequest.isType(DeclarationType.STANDARD, DeclarationType.SUPPLEMENTARY, DeclarationType.SIMPLIFIED, DeclarationType.OCCASIONAL))
```
glidester, line 261: Comment left in

app/controllers/SubmissionsController.scala
```diff
@@ -86,8 +89,15 @@ class SubmissionsController @Inject()(
     }
   }
 
-  def viewDeclaration(id: String): Action[AnyContent] = (authenticate andThen verifyEmail) { implicit request =>
-    Redirect(controllers.declaration.routes.SubmittedDeclarationController.displayPage()).addingToSession(ExportsSessionKeys.declarationId -> id)
+  def viewDeclaration(id: String): Action[AnyContent] = (authenticate andThen verifyEmail).async { implicit request =>
+    customsDeclareExportsConnector.findDeclaration(id).flatMap {
+      case Some(declaration) =>
+        customsDeclareExportsConnector.findNotifications(id).map { notifications =>
+          Ok(submittedDeclarationPage(notifications, declaration))
```
glidester, line 96: Could we clear the `declarationId` session val here to avoid the confusion of having a different Id in the cache than the dec you are viewing?

app/views/declaration/summary/items_section.scala.html
```diff
@@ -14,18 +14,17 @@
...
 
 @if(declarationData.items.nonEmpty) {
 
     <h2 class="govuk-heading-m">@messages("declaration.summary.items")</h2>
-    @declarationData.items.toSeq.sortBy(_.sequenceId).map { item =>
-        @item_section(mode, item, actionsEnabled)
+    @declarationData.items.sortBy(_.sequenceId).map { item =>
+        @item_section(mode, item, declarationData.`type`, actionsEnabled)
     }
 }
```
glidester, line 30: NIT: empty line missing

app/views/declaration/summary/submitted_declaration_page.scala.html
```diff
@@ -63,21 +58,21 @@
...
+    @items_section(Mode.Normal, declarationData, false)
 
-    <div class="govuk-section-break">@warehouse_section(Mode.Normal, request.cacheModel, false)</div>
+    <div class="govuk-section-break">@warehouse_section(Mode.Normal, declarationData, false)</div>
 
-    <div class="govuk-section-break">@transport_section(Mode.Normal, request.cacheModel, false)</div>
+    <div class="govuk-section-break">@transport_section(Mode.Normal, declarationData, false)</div>
 
     @printThisPage
 }
```
glidester, line 78: NIT: empty line missing

glidester (approved):
Code is fine, just not sure if this is the right direction to fix the underlying issue of session not being cleared when user expects/assumes it is.

app/controllers/SubmissionsController.scala
MaciejRewera: I've decided to clear this from session on Choice page for the time being. We'll see if we get a decision to move away from storing this data in session and then clearing will not be required anymore.

glidester (approved):

PR #1233, opened 2021-06-28
[CEDS-3224] Update to supplementary units

glidester:
Adding new checkbox to Commodity Measurements screen plus
new expander and some other content changes.


app/controllers/declaration/CommodityMeasureController.scala
```diff
@@ -41,20 +41,21 @@ class CommodityMeasureController @Inject()(
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
 
-  private def form()(implicit request: JourneyRequest[_]) = CommodityMeasure.form(request.declarationType).withSubmissionErrors()
-
   def displayPage(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    val commodityCode = request.cacheModel.itemBy(itemId).flatMap(_.commodityDetails).flatMap(_.combinedNomenclatureCode)
```
udinobi, line 45: Cannot use `getCommodityCode(itemId)` def ?

glidester: unused val, must of missed removing it :+1:

app/controllers/declaration/CommodityMeasureController.scala
```diff
@@ -63,6 +63,15 @@ class CommodityMeasureController @Inject()(
...
   }
 
+  private def form()(implicit request: JourneyRequest[_]) = CommodityMeasure.form(request.declarationType).withSubmissionErrors()
+
+  private def getCommodityCode(itemId: String)(implicit request: JourneyRequest[_]) =
+    for {
+      exportItem <- request.cacheModel.itemBy(itemId)
+      commodityDetails <- exportItem.commodityDetails
+      commodityCode <- commodityDetails.combinedNomenclatureCode
+    } yield commodityCode
```
MaciejRewera, line 73: I'd suggest placing this method in `ExportsDeclaration`. There are more similar to this one already.

app/forms/declaration/CommodityMeasure.scala
```diff
@@ -28,30 +28,62 @@ import utils.validators.forms.FieldValidator._
 
 import scala.util.Try
 
-case class CommodityMeasure(supplementaryUnits: Option[String], grossMass: Option[String], netMass: Option[String])
+case class CommodityMeasure(
+  supplementaryUnits: Option[String],
+  supplementaryUnitsNotRequired: Option[Boolean],
```
MaciejRewera, line 33: Is there any situation when this equals `Some(false)` ? I'm suggesting, that maybe unwrapped `Boolean` would be enough?

app/forms/declaration/CommodityMeasure.scala
```diff
@@ -28,30 +28,62 @@ import utils.validators.forms.FieldValidator._
...
   def unapplyClearance(value: CommodityMeasure): Option[(Option[String], Option[String])] =
     Some((value.grossMass, value.netMass))
 
   private val massFormatValidation: String => Boolean = str => validateDecimalGreaterThanZero(16)(6)(str) and containsNotOnlyZeros(str)
 
+  private val checkboxTrueValue = "true"
+
   private val mappingDefault = Forms.mapping(
     "supplementaryUnits" -> optional(text().verifying("declaration.commodityMeasure.supplementaryUnits.error", massFormatValidation)),
+    "supplementaryUnitsNotRequired" -> of(
```
udinobi, line 72: Not easy to understand what's going on here :woozy_face:

app/views/declaration/summary/item_section.scala.html
```diff
@@ -265,7 +265,7 @@
                     content = Text(messages("declaration.summary.items.item.supplementaryUnits"))
                 ),
                 value = Value(
-                    content = Text(commodityMeasure.supplementaryUnits.getOrElse(""))
+                    content = Text(commodityMeasure.supplementaryUnits.getOrElse("No"))
```
MaciejRewera, line 268: Not sure, but this might require it's own `messages` key-value pair (considering Welsh).

app/views/declaration/commodityMeasure/commodity_measure_default.scala.html
```diff
@@ -16,13 +16,44 @@
...
+        govukHint: GovukHint,
+        link: link,
+        appConfig: AppConfig
 )
 
-@(form: Form[CommodityMeasure])(implicit messages: Messages)
+@(form: Form[CommodityMeasure], maybeCommodityCode: Option[String])(implicit messages: Messages)
+
+@expanderContent = {
+    @maybeCommodityCode.map { commodityCode =>
```
udinobi, line 35: Could you maybe calling `getCommodityCode` passing `request.cacheModel` avoid to pass `maybeCommodityCode` from the controller?

app/controllers/declaration/CommodityMeasureController.scala
glidester: :+1:

app/forms/declaration/CommodityMeasure.scala
MaciejRewera: Agree. If I may suggest, changing the second validation to:
`!(checkbox == checkboxTrueValue and nonEmpty(supUnits))`
is easier for me to understand.

app/forms/declaration/CommodityMeasure.scala
glidester: Yes, if the user does not tick the supplementaryUnitsNotRequired checbox. I think its is better that this value is optional to reflect that the user has not yet made the yes/no determination on this field.

app/views/declaration/summary/item_section.scala.html
glidester: Yeah, good spot. :+1:

app/forms/declaration/CommodityMeasure.scala
glidester: No, play form validation never is! :grin:

app/views/declaration/commodityMeasure/commodity_measure_default.scala.html
glidester: I prefer passing the smallest amount of info required (i.e the optional commodity code, rather than the whole ExportDeclaration class), make testing much easier.

app/views/declaration/commodityMeasure/commodity_measure.scala.html
```diff
@@ -38,7 +38,7 @@
         appConfig: AppConfig
 )
 
-@(mode: Mode, itemId: String, form: Form[CommodityMeasure])(implicit request: JourneyRequest[_], messages: Messages)
+@(mode: Mode, itemId: String, form: Form[CommodityMeasure], commodityCode: Option[String])(implicit request: JourneyRequest[_], messages: Messages)
```
udinobi, line 41: Same here. Instead of passing `commodityCode: Option[String]` you could have something like..
```
request.cacheModel.getCommodityCodeOfItem(itemId)
```
btw not a big deal, not need to change it if you prefer to pass the argument.

app/views/declaration/commodityMeasure/commodity_measure_default.scala.html
glidester: Its interesting. I don't actually have a implicit `request` arg in this view class (but I do in the `commodity_measure.scala.html` that calls this view as you point out below). I have played with removing it as you suggest and while it does simplify the controller (that I like) it does complicates the tests quite a bit (as instead of simply passing the `commodityCode` to the view, we now have to craft a request instance and a whole `ExportsDeclaration` model. 

I'd prefer to leave as is for this PR, though I'll keep this in mind for future :+1:

MaciejRewera (approved):

PR #1240, opened 2021-07-01
[CEDS-3301] Skipping CUS code page

glidester:
Only show user the cus code page if commodityCode
starts with the chemical prefix digits, else skip
to taric page.

### DIFF
--- a/app/models/ExportsDeclaration.scala
+++ b/app/models/ExportsDeclaration.scala
@@ -96,6 +96,13 @@ case class ExportsDeclaration(
       commodityCode <- commodityDetails.combinedNomenclatureCode
     } yield commodityCode
 
+  def isCommodityCodeOfItemPrefixedWith(itemId: String, prefix: Seq[Int]): Boolean =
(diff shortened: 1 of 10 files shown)


app/models/ExportsDeclaration.scala
```diff
@@ -96,6 +96,14 @@ case class ExportsDeclaration(
       commodityCode <- commodityDetails.combinedNomenclatureCode
     } yield commodityCode
 
+  def isCommodityCodeOfItemPrefixedWith(itemId: String, prefix: Seq[Int]): Boolean =
+    getCommodityCodeOfItem(itemId) match {
```
udinobi, line 100:
```
    getCommodityCodeOfItem(itemId) match {
      case Some(commodityCode) if (commodityCode.trim.length > 0) =>
        prefix.exists(digits => commodityCode.startsWith(digits.toString))

      case _ => false
    }
```

udinobi (dismissed):

udinobi (approved):

PR #1242, opened 2021-07-01
CEDS-3263 Update Submissions page content

MaciejRewera:
Changes are related to handling DMSQRY notifications
and status.


app/views/submissions.scala.html
```diff
@@ -46,6 +47,14 @@
     }
 }
 
+@warningText = @{
+    if (submissionsPagesElements.actionSubmissions.currentPageElements.exists(_._2.exists(_.status == SubmissionStatus.QUERY_NOTIFICATION_MESSAGE))) {
```
udinobi, line 51: Just to simplify a little bit, maybe something like
```
val element = submissionsPagesElements.actionSubmissions.currentPageElements
if (element.exists(_._2.exists(_.status == QUERY_NOTIFICATION_MESSAGE))) {
```
importing `QUERY_NOTIFICATION_MESSAGE` ?

glidester: Yes, it is a bit of hard line to parse. Perhaps this test should be done in the controller and just pass a `isQueryNotificationPresent` boolean to the view.
Could expand this logic in the view to something more readable like:
` val isQueryNotificationPresent = submissionsPagesElements.actionSubmissions.currentPageElements
      .exists{ submissionNotificationsPair =>
        submissionNotificationsPair._2.exists( notification => notification.status == SubmissionStatus.QUERY_NOTIFICATION_MESSAGE)
      }`

udinobi (approved):

PR #1244, opened 2021-07-01
CEDS-3254 new 'Additional Documents Required' page

udinobi:
Here and there a few `TODO`s. I'm keeping them in anticipation of the next ticket, 3255. In case of complaints I can remove them btw :smiley_cat:


app/controllers/declaration/TaricCodeAddController.scala
```diff
@@ -46,21 +45,23 @@ class TaricCodeAddController @Inject()(
...
 
       case _ =>
         val form = TaricCodeFirst.form().withSubmissionErrors()
-        Ok(taricCodeAddFirstPage(mode, itemId, commodityCode(exportItem), form))
+        Ok(taricCodeAddFirstPage(mode, itemId, commodityCode(itemId), form))
     }
   }
 
   def submitForm(mode: Mode, itemId: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    def formWithErrors(formWithErrors: Form[TaricCodeFirst]): Future[Result] =
```
MaciejRewera, line 62: Nit: Could you change the name for this method to be in a verb form?

app/controllers/navigation/Navigator.scala
```diff
@@ -393,9 +394,12 @@ object Navigator {
 
   private def documentsProducedPreviousPage(cacheModel: ExportsDeclaration, mode: Mode, itemId: String): Call =
```
MaciejRewera, line 395: Since you've changed 'DocumentsProduced' to 'AdditionalDocuments', please go through the whole service and change all places where this term was used.

app/forms/common/YesNoAnswer.scala
```diff
@@ -34,7 +34,10 @@ object YesNoAnswer {
 
   import YesNoAnswers._
 
-  val allowedValues: Seq[String] = Seq(yes, no)
+  lazy val allowedValues: Seq[String] = Seq(yes, no)
+
+  lazy val Yes = YesNoAnswer(yes)
+  lazy val No = YesNoAnswer(no)
```
MaciejRewera, line 40: What is the reason for making them lazy? These fields use almost no computational and space resources.

app/models/ExportsDeclaration.scala
```diff
@@ -52,6 +53,18 @@ case class ExportsDeclaration(
   lazy val lrn: Option[String] = consignmentReferences.map(_.lrn.value)
   lazy val ducr: Option[String] = consignmentReferences.map(_.ducr.ducr)
 
+  def documentsProducedDataIfAny(itemId: String): Option[DocumentsProducedData] =
```
MaciejRewera, line 56: Here, again the old terminology is in use.

app/models/ExportsDeclaration.scala
```diff
@@ -63,6 +76,9 @@ case class ExportsDeclaration(
   def clearRoutingCountries(): ExportsDeclaration =
     copy(locations = locations.copy(hasRoutingCountries = Some(false), routingCountries = Seq.empty))
 
+  def commodityCode(itemId: String): Option[String] =
+    itemBy(itemId).flatMap(_.commodityDetails.flatMap(_.combinedNomenclatureCode))
```
MaciejRewera, line 80: 👍

app/controllers/declaration/AdditionalDocumentsRequiredController.scala
```diff
@@ -0,0 +1,74 @@
...
+  authenticate: AuthAction,
+  journeyType: JourneyAction,
+  override val exportsCacheService: ExportsCacheService,
+  navigator: Navigator,
+  mcc: MessagesControllerComponents,
+  additionalDocumentsRequired: additional_documents_required
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
+
+  private lazy val emptyKey = "declaration.additionalDocumentsRequired.empty"
```
MaciejRewera, line 45: Is it necessary to be `lazy` ?

app/views/declaration/additionalDocuments/additional_documents_required.scala.html
```diff
@@ -0,0 +1,104 @@
...
+        target = "_blank"
+    )&#46;
+}
+
+@insetSection = {
+    <p class="govuk-label--s">@messages("declaration.additionalDocumentsRequired.inset.text1")</p>
+
+    <p class="govuk-body">@Html(messages("declaration.additionalDocumentsRequired.inset.text2", commodityCodeLink))</p>
+
+    <p class="govuk-body">@licenseAndCertificatesParagraph</p>
```
MaciejRewera, line 76: There is a component `paragraphBody` made just for such situation.

test/controllers/declaration/AdditionalDocumentsRequiredControllerSpec.scala
```diff
@@ -0,0 +1,139 @@
...
+
+      "return 303 (SEE_OTHER) and redirect to the 'Additional Documents' page" when {
+        "answer is 'yes'" in {
+          verifyRedirect(Some(YesNoAnswers.yes), routes.AdditionalDocumentsAddController.displayPage(Mode.Normal, itemId))
+        }
+      }
+
+      "return 303 (SEE_OTHER) and redirect to the 'Items Summary' page" when {
+        "answer is 'no'" in {
+          verifyRedirect(Some(YesNoAnswers.no), routes.ItemsSummaryController.displayItemsSummaryPage())
```
MaciejRewera, line 106: Using this method makes it very unclear about what this test is actually testing. Especially, which method is it testing? Because in this method both controller endpoints are used.

test/controllers/declaration/AdditionalDocumentsRequiredControllerSpec.scala
```diff
@@ -0,0 +1,139 @@
...
+        "display page method is invoked and cache is not empty" in {
+          val documentsProducedData = DocumentsProducedData(Some(YesNoAnswer.Yes), Seq.empty)
+          withNewCaching(
+            aDeclarationAfter(request.cacheModel, withItem(anItem(withItemId(itemId), withDocumentsProducedData(documentsProducedData))))
+          )
+
+          val result = controller.displayPage(Mode.Normal, itemId)(getRequest())
+          status(result) must be(OK)
+          verifyPageInvoked
+        }
```
MaciejRewera, line 95: I am not sure about the purpose of this test case. It would make more sense if you would also assert if cached data was passed to the view template. Otherwise it is only (implicitly) checking if having cached data, the controller will not throw exceptions.

MaciejRewera (commented):
LGTM but requires a few improvements in my opinion.

Also, I guess that leaving some code commented made sense in this case, but I think this something we should avoid. I've seen similar situation even in this project and because the planned work was not picked up right after, the comments were left in the code for quite a long time. On the other hand bear in mind that sometimes you may plan forward this way, but if someone else picks up the next piece of work, they may have completely different view on how it should be implemented. And might even not be aware of your comments.

test/controllers/declaration/AdditionalDocumentsRequiredControllerSpec.scala
udinobi: > It would make more sense if you would also assert...
That was the idea, but I just forgot to verify if the yes radio is actually checked :upside_down_face:

app/models/ExportsDeclaration.scala
glidester: This method already exists `getCommodityCodeOfItem` Line 108

test/views/declaration/AdditionalDocumentRemoveViewSpec.scala
```diff
@@ -0,0 +1,146 @@
...
+        view.select("dl>div:nth-child(4)>dd").text() mustBe correctAdditionalDocument.documentStatusReason.get
+        view.select("dl>div:nth-child(5)>dt").text() mustBe messages("declaration.additionalDocument.remove.issuingAuthorityName")
+        view.select("dl>div:nth-child(5)>dd").text() mustBe correctAdditionalDocument.issuingAuthorityName.get
+        view.select("dl>div:nth-child(6)>dt").text() mustBe messages("declaration.additionalDocument.remove.dateOfValidity")
+        view.select("dl>div:nth-child(6)>dd").text() mustBe correctAdditionalDocument.dateOfValidity.map(_.toDisplayFormat).get
+        view.select("dl>div:nth-child(7)>dt").text() mustBe messages("declaration.additionalDocument.remove.measurementUnitAndQualifier")
+        view.select("dl>div:nth-child(7)>dd").text() mustBe correctAdditionalDocument.documentWriteOff.map(_.measurementUnitDisplay).get
+        view.select("dl>div:nth-child(8)>dt").text() mustBe messages("declaration.additionalDocument.remove.documentQuantity")
+        view.select("dl>div:nth-child(8)>dd").text() mustBe correctAdditionalDocument.documentWriteOff.flatMap(_.documentQuantity).map(_.toString).get
```
MaciejRewera, line 93: Is it not possible to find these elements by class or even id?

MaciejRewera (dismissed):
LGTM, but I've left a question about tests.
Also, there are changes to urls. Usually, PAs wanted to know about it, so please let Alex know ;)

MaciejRewera (approved):
LGTM

PR #1256, opened 2021-07-21
CEDS-3261 Redesign of the 'Declaration Details' page

udinobi:


app/views/helpers/TimelineEvents.scala
```diff
@@ -0,0 +1,46 @@
...
+
+package views.helpers
+
+import java.time.ZonedDateTime
+
+import models.declaration.notifications.Notification
+import play.api.i18n.Messages
+import play.twirl.api.Html
+
+case class TimelineEvent(title: String, dateTime: ZonedDateTime, dateTimeAsShown: String, content: Option[Html]) extends Ordered[TimelineEvent] {
```
glidester, line 25: Nit: Do we want two versions of this event's dateTime (a real date and a string version of it)? Seems a violation of the 'single source of truth' rule.

udinobi: Point is that `Timeline` is a reusable generic partial. In a different page we could want a different format of `dateTime`. In any case, the `dateTime` field in `TimelineEvent` is not important at all (I would say even optional). It's only used for the html attribute `datetime` of tag `time` and it does not matter even if it is different from `dateTimeAsShown`. It was my choice to add it to the tag for completeness, but it is of no use for us.

test/views/DeclarationDetailsViewSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+import views.helpers.TimelineEvents.dateTimeAsShown
+import views.html.declaration_details
+
+class DeclarationDetailsViewSpec extends UnitViewSpec with Injector {
+
+  private val submission = Submission("id", "eori", "lrn", Some("mrn"), Some("ducr"), Seq.empty)
+
+  private val now = ZonedDateTime.now
+
+  private val notifications = List(
```
glidester, line 42: Would be good to test for the scenario where there are no Notifications for a given submission (how would the page handle that?)

glidester (commented):
Would be good to add test for when there are no notifications for a submission

test/views/DeclarationDetailsViewSpec.scala
udinobi: I thought about that but it seemed to me that we always have at least one notification if we are in that page... but I might be wrong :upside_down_face:

MaciejRewera: According to what business wants, you're right @udinobi . But IMHO we should be covering for unhappy-path scenarios.

app/views/helpers/TimelineEvents.scala
MaciejRewera: I can see `dateTime` field is used to populate `datetime` attribute in the html. Although, I am not sure if you really need this field. If you're saying it is not really used, then I'd remove that from the case class. I am aware that if you don't remove it now, we'll get stuck with it forever.

app/views/components/gds/timeline.scala.html
```diff
@@ -0,0 +1,26 @@
...
+            <div class="hmrc-timeline__event-content">
+                @content
+            </div>
+        }
+    </li>
+}
+
+<ol class="hmrc-timeline govuk-!-margin-top-5">
+    @timelineEvents.map(htmlForTimelineEvent)
+</ol>
```
MaciejRewera, line 26: Have you checked if there is a component available in govuk/hmrc frontend libs?

app/views/components/gds/timeline.scala.html
```diff
@@ -0,0 +1,26 @@
...
+@import views.helpers.TimelineEvent
+
+@this()
+
+@(timelineEvents: Seq[TimelineEvent])
+
+@htmlForTimelineEvent(timelineEvent: TimelineEvent) = {
+    <li class="hmrc-timeline__event">
+        <h2 class="hmrc-timeline__event-title">@timelineEvent.title</h2>
+        <time class="hmrc-timeline__event-meta" datetime="@{ISO_OFFSET_DATE_TIME.format(timelineEvent.dateTime)}">
```
MaciejRewera, line 13: Could this be formatted the same way the displayed dateTime is formatted?

test/controllers/DeclarationDetailsControllerSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+          responsesToReturn(isQueryNotificationMessageEnabled = true)
+
+          val result = controller.displayPage(actionId)(getRequest())
+          status(result) mustBe OK
+
+          verifyNoInteractions(declarationInformationPage)
+
+          verifyPage { (submissionCaptor: ArgumentCaptor[Submission], notificationsCaptor: ArgumentCaptor[Seq[Notification]]) =>
+            verify(declarationDetailsPage).apply(submissionCaptor.capture(), notificationsCaptor.capture())(any(), any())
+          }
```
MaciejRewera, line 108: Nit: I am not convinced with this new approach. We used to handle this situation by having a method returning proper `Captor` instance. And then it was invoked for every verify I am just not comfortable with introducing more and more new approaches to handle the same situation.
Maybe instead of passing this function parameter, you could simply pass the template/page? You'd avoid some of the boiler-plate code.

test/controllers/DeclarationDetailsControllerSpec.scala
```diff
@@ -0,0 +1,143 @@
...
+        val notificationsCaptor: ArgumentCaptor[Seq[Notification]] = ArgumentCaptor.forClass(classOf[Seq[Notification]])
+        callsiteToVerify(submissionCaptor, notificationsCaptor)
+        submissionCaptor.getValue mustBe submission
+        notificationsCaptor.getValue mustBe Seq(notification)
+      }
+    }
+
+    "return 303 (SEE_OTHER)" when {
+
+      "there is no submission during display Declaration with notification method" in {
```
MaciejRewera, line 129: It might be useful to add a test scenario when there is a `Submission`, but no `Notification` related to it.

app/views/components/gds/timeline.scala.html
udinobi: The timeline only consists of CSS styling, it's not defined as a component.
If you think about it, different services might have different requirements in term of what to display in the timeline (even if the design guidelines ask for at least title, date and/or time) and how to style the timeline's items, so IMO there is not point to have it as a component from the library POV. There are only two other hmrc services in github using the timeline CSS and they are also defining their own timeline component.

app/views/components/gds/timeline.scala.html
udinobi: No. It adds a semantic **timestamp** to the element, and it should be readable by screen readers, search engines, JavaScript code for apps that want to set for instance appointments, calendar remainders or something like that.

test/controllers/DeclarationDetailsControllerSpec.scala
udinobi: I tried just passing the template/page, but what's the common type to be used for the page parameter?
If you have an alternative implementation I'll be more than happy to use that. I really do not care about how we implement this verification. If not I can even just use the same verifying code for each of two pages under test.

app/views/helpers/TimelineEvents.scala
udinobi: Actually, it was a nice suggestion because it forced me to find a better solution :slightly_smiling_face:

test/controllers/DeclarationDetailsControllerSpec.scala
udinobi: Refactored

app/views/components/gds/timeline.scala.html
MaciejRewera: You are right. And according to what you're saying, there are not much adoption for this pattern.
But I think it is feasible to do such a generic component that could serve all services.

Anyway, if it is not available ATM, then we have no other choice.

MaciejRewera (approved):
LGTM

PR #1258, opened 2021-07-23
CEDS-3249 Update conditionally the content of the 'Additional Documents' page

udinobi:


app/controllers/declaration/AdditionalDocumentChangeController.scala
```diff
@@ -75,6 +75,9 @@ class AdditionalDocumentChangeController @Inject()(
     }
   }
 
+  private def isAuthCodeRequiringAdditionalDocuments(implicit request: JourneyRequest[_]): Boolean =
+    request.cacheModel.isAuthCodeRequiringAdditionalDocuments
+
```
MaciejRewera, line 80: Maybe instead of having this method in two classes, you could change the `form` method to accept `ExportsDeclaration`as a parameter? Then you simply write `form(request.cacheModel)`. This is concise and readable and you don't need to create same methods.

MaciejRewera: Also, I think it is better to `ExportsDeclaration` as a parameter, because right now you rely on client of this mehod to know what the `Boolean` param means.

app/views/helpers/AdditionalDocument.scala
```diff
@@ -0,0 +1,62 @@
...
+      val html = new Html(
+        List(
+          new paragraphBody().apply(messages("declaration.additionalDocument.documentIdentifier.inset.fromAuthCode.paragraph1")),
+          new paragraphBody().apply(messages("declaration.additionalDocument.documentIdentifier.inset.fromAuthCode.paragraph2"))
+        )
+      )
+
+      val insets = new govukInsetText().apply(InsetText(content = HtmlContent(html)))
+      Some(insets)
+    }
```
MaciejRewera, line 46: Could this be changed not to use negation (`!`) ? You could have the whole logic inside the positive case then.

test/forms/declaration/AdditionalDocumentSpec.scala
```diff
@@ -37,14 +37,23 @@ class AdditionalDocumentSpec extends UnitSpec {
 
       "provided with Document Type Code" which {
 
-        "is empty" in {
+        "is empty when it is instead required" in {
```
MaciejRewera, line 40: Nit: Could this be re-worded? It is not clear to me what it means...

test/views/declaration/AdditionalDocumentEditViewSpec.scala
```diff
@@ -136,19 +145,63 @@ class AdditionalDocumentEditViewSpec extends UnitViewSpec with CommonMessages wi
     }
   }
 
+  "additional_document_edit view on empty page" when {
+
+    "the entered authorisation code requires additional documents" should {
```
MaciejRewera, line 150: I can't find symetrical case for this one. Is it treated as a default? Since we have two different versions of the page I'd expect that to be reflected in two sections in tests.

MaciejRewera (approved):
LGTM, but left some nit comments.

test/views/declaration/AdditionalDocumentEditViewSpec.scala
udinobi: Yes. The page's default content is still there, how it was before this ticket. Only some parts (title and a couple of field hints) have to be changed when the auth code requires additional documents

MaciejRewera (approved):
LGTM

PR #1261, opened 2021-07-28
[CEDS-3131] Trim white spaces

glidester:
Auto trim strings on certain fields so that users do not need
to worry if they accidentally paste a white space char at the
beginning or end of a value

### DIFF
--- a/test/forms/CancelDeclarationSpec.scala
+++ b/test/forms/CancelDeclarationSpec.scala
@@ -17,25 +17,30 @@
 package forms
 
 import base.UnitSpec
+import play.api.data.Form
 import play.api.libs.json.{JsObject, JsString}
 
 class CancelDeclarationSpec extends UnitSpec {
 
-  def formData(lrn: String = "lrn", mrn: String = "123456789012345678", description: String = "description", reason: String = "1") =
-    JsObject(
-      Map(
-        "functionalReferenceId" -> JsString(lrn),
-        "mrn" -> JsString(mrn),
-        "statementDescription" -> JsString(description),
-        "changeReason" -> JsString(reason)
-      )
+  private val validMrn = "123456789012345678"
+  def getBoundedForm(lrn: String = "lrn", mrn: String = validMrn, description: String = "description", reason: String = "1") =
+    CancelDeclaration.form.bind(
+      JsObject(
+        Map(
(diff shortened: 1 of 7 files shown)


test/forms/declaration/GoodsLocationFormSpec.scala
```diff
@@ -16,25 +16,103 @@
...
 
 class GoodsLocationFormSpec extends DeclarationPageBaseSpec {
 
+  private val validCode = "GBAUFXTFXTFXT"
+
   "GoodsLocation form" should {
 
+    "return form with errors" when {
+
+      "provided with a Code" which {
```
MaciejRewera, line 32: 👍

MaciejRewera (approved):
LGTM

test/forms/CancelDeclarationSpec.scala
```diff
@@ -17,105 +17,110 @@
...
-      Map(
-        "functionalReferenceId" -> JsString(lrn),
-        "mrn" -> JsString(mrn),
-        "statementDescription" -> JsString(description),
-        "changeReason" -> JsString(reason)
-      )
+  private val validMrn = "123456789012345678"
+  def getBoundedForm(lrn: String = "lrn", mrn: String = validMrn, description: String = "description", reason: String = "1") =
+    CancelDeclaration.form.bind(
+      JsObject(
```
udinobi, line 28: `bind(JsValue)` is deprecated. You can directly use here `bind(Map)`. No need to convert the map to json.

test/forms/CancelDeclarationSpec.scala
```diff
@@ -17,105 +17,110 @@
...
+        Map(
+          "functionalReferenceId" -> JsString(lrn),
```
udinobi, line 30: No need to convert to json. Pass directly the value. You can use `.toString` when it hasn't `String` type.

test/forms/CancelDeclarationSpec.scala
glidester: Not the version I used (that includes the `MaxChars` param) :slightly_smiling_face:

udinobi: OK, but when you already have a Map when not directly use it? Why the need to convert it to json?

PR #1264, opened 2021-07-29
CEDS-3237 Fix Change link for 'Authorisations for this declaration'

MaciejRewera:
This implementation relies on /authorisation-choice GET endpoint
redirecting to /is-authorisation-required endpoint, according to
the journey flow.
This also allows all pages pages that previously contained this
logic to simply redirect to /authorisation-choice and not even
be aware of possible routing.


app/controllers/declaration/AdditionalActorsSummaryController.scala
```diff
@@ -58,7 +59,7 @@ class AdditionalActorsSummaryController @Inject()(
         validYesNo =>
           validYesNo.answer match {
             case YesNoAnswers.yes => navigator.continueTo(mode, controllers.declaration.routes.AdditionalActorsAddController.displayPage)
-            case YesNoAnswers.no  => navigator.continueTo(mode, nextPage)
+            case YesNoAnswers.no  => navigator.continueTo(mode, controllers.declaration.routes.AuthorisationProcedureCodeChoiceController.displayPage)
```
udinobi, line 62: Nit: `controllers.declaration.` can be removed. Personally I would use this as a general rule for a better readability

app/controllers/declaration/AuthorisationProcedureCodeChoiceController.scala
```diff
@@ -40,29 +43,46 @@ class AuthorisationProcedureCodeChoiceController @Inject()(
...
-
-  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType(validTypes)) { implicit request =>
-    val form = AuthorisationProcedureCodeChoice.form().withSubmissionErrors()
-    request.cacheModel.parties.authorisationProcedureCodeChoice match {
-      case Some(data) => Ok(authorisationProcedureCodeChoice(form.fill(data), mode))
-      case _          => Ok(authorisationProcedureCodeChoice(form, mode))
+  def displayPage(mode: Mode): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
+    request.declarationType match {
+      case CLEARANCE if request.cacheModel.isNotEntryIntoDeclarantsRecords =>
+        navigator.continueTo(mode, controllers.declaration.routes.DeclarationHolderRequiredController.displayPage)
```
udinobi, line 49: Same here

app/controllers/declaration/AuthorisationProcedureCodeChoiceController.scala
```diff
@@ -40,29 +43,46 @@ class AuthorisationProcedureCodeChoiceController @Inject()(
...
+      case OCCASIONAL =>
+        navigator.continueTo(mode, controllers.declaration.routes.DeclarationHolderRequiredController.displayPage)
```
udinobi, line 51: Same here

app/controllers/declaration/AuthorisationProcedureCodeChoiceController.scala
```diff
@@ -40,29 +43,46 @@ class AuthorisationProcedureCodeChoiceController @Inject()(
...
     choice: AuthorisationProcedureCodeChoice
   )(implicit request: JourneyRequest[AnyContent]): Future[Option[ExportsDeclaration]] =
     updateExportsDeclarationSyncDirect(_.updateAuthorisationProcedureCodeChoice(choice))
+
+  private def nextPage(exportsDeclarationOpt: Option[ExportsDeclaration]): Mode => Call =
+    (for {
+      exportsDeclaration <- exportsDeclarationOpt
+      authProcedureCodeChoice <- exportsDeclaration.parties.authorisationProcedureCodeChoice.map(_.code)
+
+    } yield (exportsDeclaration.`type`, exportsDeclaration.additionalDeclarationType, authProcedureCodeChoice)) match {
```
udinobi, line 83: Nit: I would use `def someFun(exportsDeclaration, authProcedureCodeChoice)` for the `yield` block.

app/controllers/declaration/ConsigneeDetailsController.scala
```diff
@@ -65,11 +63,9 @@ class ConsigneeDetailsController @Inject()(
   }
 
   private def nextPage()(implicit request: JourneyRequest[AnyContent]): Mode => Call =
-    (request.declarationType, request.cacheModel.isEntryIntoDeclarantsRecords) match {
-      case (DeclarationType.CLEARANCE, true) =>
+    request.declarationType match {
+      case CLEARANCE =>
         controllers.declaration.routes.AuthorisationProcedureCodeChoiceController.displayPage
```
udinobi, line 68: Same here

app/controllers/declaration/ConsigneeDetailsController.scala
```diff
@@ -65,11 +63,9 @@ class ConsigneeDetailsController @Inject()(
...
-      case (DeclarationType.CLEARANCE, false) =>
-        controllers.declaration.routes.DeclarationHolderSummaryController.displayPage
       case _ =>
         controllers.declaration.routes.AdditionalActorsSummaryController.displayPage
```
udinobi, line 70: Same here

app/views/declaration/summary/parties_section_holders.scala.html
```diff
@@ -51,13 +51,17 @@
             changeLink(
                 messages("site.change"),
                 Some(messages("declaration.summary.parties.holders.change", holder.authorisationTypeCode.getOrElse(""), holderEori(holder))),
-                controllers.declaration.routes.DeclarationHolderSummaryController.displayPage(mode)
+                controllers.declaration.routes.AuthorisationProcedureCodeChoiceController.displayPage(mode)
```
udinobi, line 54: Same here

app/views/declaration/summary/parties_section_holders.scala.html
```diff
@@ -71,7 +75,7 @@
             actions = actions(Actions(
                 items = Seq(
                     actionItem(
-                        href = controllers.declaration.routes.DeclarationHolderSummaryController.displayPage(mode).url,
+                        href = controllers.declaration.routes.AuthorisationProcedureCodeChoiceController.displayPage(mode).url,
```
udinobi, line 78: Same here

udinobi (dismissed):
A few up-to-you comments

udinobi (approved):
LGTM

app/controllers/declaration/AuthorisationProcedureCodeChoiceController.scala
MaciejRewera: Tried that, but the `yield` block ends right before `match` statement. So the pattern matching actually works with the output of this for-comprehension.
Therefore, if I were to extract it into a function, then I would still end up with `Option` type as a result and would need to unwrap it providing default value. Which makes pattern matching redundant.

PR #1265, opened 2021-07-30
CEDS-3271 'Documents required' html section on new Timeline page

udinobi:


app/views/helpers/TimelineEvents.scala
```diff
@@ -18,22 +18,38 @@ package views.helpers
...
-  def compare(that: TimelineEvent): Int =
-    if (dateTime == that.dateTime) 0
-    else if (dateTime.isAfter(that.dateTime)) -1
-    else 1
-}
+class TimelineEvents @Inject()(sfusConfig: SfusConfig, uploadFilesPartialForTimeline: upload_files_partial_for_timeline) {
+  def apply(submission: Submission, notifications: Seq[Notification])(implicit messages: Messages): Seq[TimelineEvent] = {
+    val sortedNotifications =
+      notifications
+        .map(notification => notification.copy(dateTimeIssued = notification.dateTimeIssuedInUK))
```
glidester, line 35: Not sure what the purpose of this reassignment is, the ordering would not be affected in any way :thinking:

udinobi: From what I read the normalisation we are doing by using `ZonedDateTime.withZoneSameInstant` could potentially affect the order if the source ZonedDateTime instances in `Notification` have different time zones.

But tbh, I wasn't sure, so I went for this solution just to be safe.

glidester (dismissed):

glidester (approved):

PR #1267, opened 2021-08-03
[CEDS-3292] Supplementary Content update and validation

glidester:
Adding extra MRN field to the /consignment-references page
for supplementary journeys only (plus some content
changes)


app/forms/declaration/ConsignmentReferences.scala
```diff
@@ -16,24 +16,53 @@
...
+
+    def form2Model: (Ducr, Lrn, Option[Mrn]) => ConsignmentReferences = {
+      case (ducr, lrn, mrn) =>
+        mrn match {
+          case Some(x) => ConsignmentReferences(ducr, lrn, mrn)
+          case None    => ConsignmentReferences(ducr, lrn, None)
+        }
+    }
+
+    def model2Form: ConsignmentReferences => Option[(Ducr, Lrn, Option[Mrn])] =
```
udinobi, line 43: Maybe worth to have a proper `case class(ducr, lrn, mrn)` ?

udinobi (approved):
One small comment

app/forms/declaration/ConsignmentReferences.scala
glidester: Not sure what you mean, this is the unapply function for the mapping (so it must return an Option[Tuple3])

PR #1276, opened 2021-08-06
CEDS-3373 Add a DUCR validation to the Consignment References page

udinobi:
An error will be displayed on the Consignment References page if the user enters a DUCR already submitted.


test/controllers/declaration/ConsignmentReferencesControllerSpec.scala
```diff
@@ -97,7 +101,7 @@ class ConsignmentReferencesControllerSpec extends ControllerSpec {
     }
   }
 
-  "ConsignmentReferencesController on displayPage" should {
```
glidester, line 100: Are there no tests then for the displayPage method of the controller?

glidester (approved):

PR #1283, opened 2021-08-20
CEDS-3368 Content amendments to the 'Goods Location' page

udinobi:


app/views/components/gds/paragraph.scala.html
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@this()
+
+@(html: Html, classes: String = "govuk-body", id: Option[String] = None)
+
+<p class="@classes" @id.map{id => id="@id" }>@html</p>
```
glidester, line 21: Nit: might be worth updating the original paragraphBody template to call this new paragraph template to avoid duplication of the definition of the '<p>' tag

glidester (approved):

app/views/components/gds/paragraph.scala.html
udinobi: Tbh... the duplication of tags is less worrying to me than the nesting of partials, which makes more problematic using them in pure scala-code files, due to the dependency injection requirement.

PR #1285, opened 2021-08-24
CEDS-3379 - Ro-Ro guidance page

oddbunchmedia:
Just waiting on Liz to check the page but any other feedback welcome in the interim.

The tests are just a copy/paste/edit from the previous guidance page - if there is a better way to do this or it won't work as is (wasn't sure how to test other than local sbt test) let me know.

### DIFF
--- a/app/views/guidance/send_by_roro.scala.html
+++ b/app/views/guidance/send_by_roro.scala.html
@@ -26,7 +26,7 @@
 
 @()(implicit request: Request[_], messages: Messages)
 
-@govukLayout(title = Title("guidance.error.explanation.title"), useCustomContentWidth = true) {
+@govukLayout(title = Title("guidance.roro.title"), useCustomContentWidth = true) {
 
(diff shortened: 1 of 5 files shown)


app/views/guidance/send_by_roro.scala.html
```diff
@@ -0,0 +1,122 @@
...
+    govukLayout: gdsMainTemplate,
+    pageTitle: pageTitle,
+    paragraphBody: paragraphBody
+)
+
+@()(implicit request: Request[_], messages: Messages)
+
+@govukLayout(title = Title("guidance.roro.title"), useCustomContentWidth = true) {
+
+    <span class="govuk-caption-xl">Guidance</span>
```
udinobi, line 31: @sectionHeader("Guidance")

oddbunchmedia: Changed this to use the component, will update this on the error codes page as well.

PR #1286, opened 2021-08-25
BAU Change a few template helpers to singleton classes

udinobi:
Change a few template helpers to singleton classes in order to inject the component/partial instances used, instead of creating new ones at every call.

### DIFF
--- a/app/views/declaration/additionalDocuments/additional_document_edit.scala.html
+++ b/app/views/declaration/additionalDocuments/additional_document_edit.scala.html
@@ -40,6 +40,7 @@
     tariffExpander: tariffExpander,
     saveButtons: saveButtons,
     link: link,
+    additionalDocument: AD,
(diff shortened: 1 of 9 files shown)


app/views/declaration/additionalDocuments/additional_document_edit.scala.html
```diff
@@ -40,6 +40,7 @@
     tariffExpander: tariffExpander,
     saveButtons: saveButtons,
     link: link,
+    additionalDocument: AD,
```
glidester, line 43: Nit: the pseudonym "AD" looks a little odd, maybe the helper class could be renamed "AdditionalDocHelper" to avoid this?

glidester (approved):
Looks much better :+1:

glidester (approved):

PR #1290, opened 2021-08-27
CEDS-3381 - Start page

oddbunchmedia:
Not sure what to do about URLs in the conf file - there are other URLs that use localhost:6791 but I feel like that's going to make testing these URLs difficult in other env?


app/views/guidance/start.scala.html
```diff
@@ -0,0 +1,97 @@
...
+    numberedList: numberedList,
+    sectionHeader: sectionHeader,
+    appConfig: AppConfig
+)
+
+@()(implicit request: Request[_], messages: Messages)
+
+@govukLayout(title = Title("guidance.start.title")) {
+
+    @pageTitle(messages("guidance.start.title"), "govuk-heading-xl")
```
udinobi, line 39: `, "govuk-heading-xl"` can be removed

app/views/components/gds/button.scala.html
```diff
@@ -0,0 +1,35 @@
+@*
```
udinobi, line 1: Do we really need this partial? It seems to me that it does not help compared to the direct use of `@govukButton`.

conf/application.conf
```diff
@@ -153,6 +153,7 @@ assets {
 urls {
   login = "http://localhost:9949/auth-login-stub/gg-sign-in"
   loginContinue = "http://localhost:6791/customs-declare-exports/start"
+  firstPageInService = "/customs-declare-exports/choice"
```
MaciejRewera, line 156: Since these two urls are related to this service, there is a better way to define them. You can simply put `controllers.routes.ChoiceController.displayPage().url` in any place you need to redirect to the Choice Page.

conf/application.conf
```diff
@@ -194,6 +195,7 @@ urls {
   additionalDocumentsReferenceCodes = "https://www.gov.uk/guidance/data-element-23-documents-and-other-reference-codes-national-of-the-customs-declaration-service-cds#table-2-national-document-codes"
   additionalDocumentsUnitCodes = "https://www.gov.uk/government/publications/measurement-unit-codes-for-data-elements-23-and-44-of-the-customs-declaration-service"
   licensesForExportingGoods = "https://www.gov.uk/export-goods"
+  exportDeclarationGuidance = "/customs-declare-exports/complete-declaration"
```
MaciejRewera, line 198: Added a better alternative above, but it seems that there is no such endpoint at the moment in this service.

test/views/guidance/StartViewSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  private val startPage = instanceOf[start]
+  private val view = startPage()(request, messages)
+
+  "'Start' view" should {
+
+    "display page header" in {
+      view.getElementsByTag("h1").first must containMessage("guidance.start.title")
+    }
+  }
+}
```
MaciejRewera, line 39: Don't know if there was any other convention agreed for these pages, but normally we tend to test in more detail.

MaciejRewera (commented):
LGTM apart from the way the urls to the service are made. It is better to use `routes` as they are less prone to errors.

PR #1292, opened 2021-09-06
CEDS-3383 - Added entry page for guidance

oddbunchmedia:


app/views/guidance/entry.scala.html
```diff
@@ -0,0 +1,95 @@
...
+    @paragraphBody(messages("guidance.entry.section.1.paragraph.4", link(
+        text = messages("guidance.entry.section.1.paragraph.4.link.1"),
+        call = Call("GET", controllers.routes.GuidanceController.sendByRoro.url)
+    )))
+
+
+@* Section 2 *@
+
+    @heading(messages("guidance.entry.section.2.header.1"), "govuk-heading-l", "h2")
+    @heading(messages("guidance.entry.section.2.header.2"), "govuk-heading-m", "h3")
```
udinobi, line 75: My only doubt is if `sub-header`s should have as class `govuk-heading-m` or `govuk-heading-s`, because a part being in bold, in the text prototype they seem to have the same size of normal text.

We should agreed on this so to have uniformity there.

oddbunchmedia: I'm happy either way - I went with what was in the prototype:

        <h2 class="govuk-heading-l">Make the declaration yourself
          <h3 class="govuk-heading-m">Make a declaration on the new Customs Declaration Service (CDS) system</h3>
        </h2>

To be honest I don't like orphan subheaders - so if an improvement can be made here then great.

udinobi: I'll leave the decision to you :+1: 

I have used `govuk-heading-s` as class and `paragraphBody` as partial for the `complete_declaration` page.

If you decide for `heading(..., "govuk-heading-m")` please also amend it in `complete_declaration` (it's just one line) by using this ticket.

oddbunchmedia: Changed to -s for consistency.

udinobi (dismissed):

udinobi (approved):

PR #1297, opened 2021-09-08
CEDS 3394 new authorizations page with automatic eori population

glidester:


app/controllers/declaration/DeclarationHolderChangeController.scala
```diff
@@ -40,34 +41,44 @@ class DeclarationHolderChangeController @Inject()(
...
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
 
-  def displayPage(mode: Mode, id: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
-    val holder = DeclarationHolderAdd.fromId(id)
-    Ok(declarationHolderChangePage(mode, id, form.fill(holder).withSubmissionErrors()))
+  def displayPage(mode: Mode, id: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    val maybeExistingHolder = cachedHolders.find(_.id.equals(id))
+
+    maybeExistingHolder.fold(errorHandler.displayErrorPage()) { holder =>
```
udinobi, line 53: Just wondering, when there's no an existing declaration holder, what information will be shown to the user when landing on this page.

udinobi: Will the default error page enough to explain what's happening?

app/controllers/declaration/DeclarationHolderRemoveController.scala
```diff
@@ -40,38 +42,45 @@ class DeclarationHolderRemoveController @Inject()(
...
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with ModelCacheable with SubmissionErrors {
 
-  def displayPage(mode: Mode, id: String): Action[AnyContent] = (authenticate andThen journeyType) { implicit request =>
-    val holder = DeclarationHolderAdd.fromId(id)
-    Ok(holderRemovePage(mode, holder, removeYesNoForm.withSubmissionErrors()))
+  def displayPage(mode: Mode, id: String): Action[AnyContent] = (authenticate andThen journeyType).async { implicit request =>
+    val maybeExistingHolder = cachedHolders.find(_.id.equals(id))
+
+    maybeExistingHolder.fold(errorHandler.displayErrorPage()) { holder =>
```
udinobi, line 54: Same here as above

app/controllers/declaration/DeclarationHolderChangeController.scala
glidester: If the id passed does not match a holder in the cache then `errorHandler.displayErrorPage()` is called which is a generic error page for the service.

udinobi: I got this :+1: . Just wondering if we need to show a more specific error message than the generic "There is a problem. Try again later".

Not a big deal btw.

glidester: Yeah, its an odd one as this scenario should never be experienced by a user apart from in some very edge cases (like user editing URL param, user bookmarking old page or someone concurrently removing this holder and then trying to edit/remove it on another tab).

app/models/declaration/EoriSource.scala
```diff
@@ -0,0 +1,33 @@
...
+
+sealed trait EoriSource
+
+object EoriSource {
+  case object UserEori extends EoriSource
+  case object OtherEori extends EoriSource
+
+  val values = Seq(UserEori, OtherEori)
+
+  lazy val lookupByValue: Map[String, EoriSource] = values.map(entry => entry.toString -> entry).toMap
```
MaciejRewera, line 29: Similar case to the one in the backend repository. Not sure how `collect` method works, but if it is an eager implementation, then having `lookupByValue` lazy, does not make any difference.

MaciejRewera (approved):
LGTM despite one small comment

PR #1300, opened 2021-09-10
[CEDS-3403] Update of content to additional-documentation

glidester:
Adding support to form field components to enable
label -> bodyText -> hintText -> form field
combination (bodyText not supported before).


app/views/components/gds/exportsDateInput.scala.html
```diff
@@ -36,15 +42,30 @@
   if(errors.nonEmpty) Some(ErrorMessage(content = HtmlContent(errors))) else None
 }
 
+@bodyOrFieldsetTuple = @{bodyHtml
```
udinobi, line 45: tbh it's not really clear to me what's going on here. Maybe a simplification by using helper methods whose names could indicate what they are for? Just my 2 cents

udinobi:
As a general comment, these `bodyHtml.map` ops within the components seem to me quite obscure to figure out how to use them. Maybe not today, but in some time. Helper methods could "**perhaps**" help.

app/views/components/gds/exportsDateInput.scala.html
glidester: I've modified the template to use two options rather than a tuple (which does add to the complexity I agree). Any better?

glidester: I've edited the template to use two options rather than a tuple (that does make things more complex I agree). Is this improvement any better?

app/views/components/gds/exportsInputText.scala.html
```diff
@@ -26,41 +30,45 @@
     hintKey: Option[String] = None,
     hintKeyLinks: Seq[Html] = Seq.empty,
     hintHtml: Option[Html] = None,
+    bodyHtml: Option[Html] = None,
```
udinobi, line 33: Could this be renamed to something like `bodyUnderLabel`, `paragraphUnderLabel` or similar so to make clear what's for?

app/views/components/gds/exportsInputTextArea.scala.html
```diff
@@ -14,19 +14,25 @@
...
-        hintKey: Option[String] = None,
-        isPageHeading: Boolean = false,
-        headingClasses: String = gdsPageLabel,
-        labelClasses: String = "govuk-label--m",
-        inputClasses: Option[String] = None
+    field: Field,
+    labelKey: String,
+    labelArg: String = "",
+    hintKey: Option[String] = None,
+    bodyHtml: Option[Html] = None,
```
udinobi, line 31: Could this be renamed to something like `bodyUnderLabel`, `paragraphUnderLabel` or similar so to make clear what's for?

app/views/components/gds/exportsInputText.scala.html
glidester: Don't really want to include positional information in a param name (as it might not always be in that position). I think the concept of each field component having a form, label and then an optional body and hint element is clear enough no? Could rename 'body' to be 'description' I suppose if that is any better for you?

udinobi (approved):

PR #1301, opened 2021-09-10
CEDS-3390 - Rejected declaration links

oddbunchmedia:


udinobi:
It would maybe worth to test the correctness of the new links in `rejected_notification_errors.scala.html`

oddbunchmedia:
> It would maybe worth to test the correctness of the new links in `rejected_notification_errors.scala.html`

Added in some additional link tests based on what I found in other tests.

udinobi (approved):

PR #1302, opened 2021-09-11
CEDS-3405 Add currency guidance to 'statistical_value' page

udinobi:
Ticket's purpose was in truth really simple. It just required me to make a few modifications to
- configuration
- statistical_value.scala.html
- StatisticalValueViewSpec.scala
 
However, it has been a while since I wanted to move, in the parameters of the `link` component, `textHidden`, since its 2nd position in the parameter set was quite annoying, forcing you to pass `None`, or to name the other passed params, when not required. Note that this parameter is mostly used only in the partials of the summary page.

Unfortunately (sorry :grimacing:, Friday afternoon) I forgot to use different commits for the 2 change sets, which would have made easier to review the PR.


glidester (approved):
LGTM

PR #1306, opened 2021-09-14
CEDS-3393 Content amendments to the page at '/additional-taric-code'

udinobi:


test/views/declaration/TaricCodeAddFirstViewSpec.scala
```diff
@@ -98,7 +102,7 @@ class TaricCodeAddFirstViewSpec extends UnitViewSpec with ExportsTestData with S
       view must haveGovukGlobalErrorSummary
       view must containErrorElementWithTagAndHref("a", "#taricCode")
 
-      view must containErrorElementWithMessageKey("declaration.taricAdditionalCodes.error.invalid")
+      view must containErrorElementWithMessageKey(s"$prefix.error.invalid")
```
glidester, line 105: Nit: Making the i18n key dynamically generated with `$prefix` in the test just makes it more difficult to find uses of this key in the IDE for future refactoring work. I know it is DRYer, but in this instance I think it is counter productive (IMO :slightly_smiling_face: )

glidester (dismissed):

test/views/declaration/TaricCodeAddFirstViewSpec.scala
udinobi: :+1:

glidester (approved):
:+1:

PR #1309, opened 2021-09-15
CEDS-3410 LRN validation on ConsignmentReferences page

MaciejRewera:
The new validation checks whether the same LRN
has not been used in the past 48 hours.

### DIFF
--- a/app/connectors/CustomsDeclareExportsConnector.scala
+++ b/app/connectors/CustomsDeclareExportsConnector.scala
@@ -117,24 +117,25 @@ class CustomsDeclareExportsConnector @Inject()(appConfig: AppConfig, httpClient:
 
   def submitDeclaration(id: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Submission] =
     httpClient
-      .POSTEmpty[Submission](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.declarations}/$id/submission")
-
-  def findSubmission(id: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Option[Submission]] =
-    httpClient
-      .GET[Option[Submission]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.declarations}/$id/submission")
-
-  def findNotifications(id: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Seq[Notification]] =
-    httpClient
-      .GET[Seq[Notification]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.declarations}/$id/submission/notifications")
-
-  def fetchNotifications()(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Seq[Notification]] =
-    httpClient.GET[Seq[Notification]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.fetchNotifications}")
+      .POSTEmpty[Submission](s"${appConfig.customsDeclareExportsBaseUrl}/submission/$id")
 
   def fetchSubmissions()(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Seq[Submission]] =
-    httpClient.GET[Seq[Submission]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.fetchSubmissions}").map { response =>
+    httpClient.GET[Seq[Submission]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.submissions}").map { response =>
       logger.debug(s"CUSTOMS_DECLARE_EXPORTS fetch submission response is --> ${response.toString}")
       response
     }
+
+  def findSubmission(id: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Option[Submission]] =
+    httpClient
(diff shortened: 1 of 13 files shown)


app/connectors/CustomsDeclareExportsConnector.scala
```diff
@@ -117,25 +117,30 @@ class CustomsDeclareExportsConnector @Inject()(appConfig: AppConfig, httpClient:
 
   def submitDeclaration(id: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Submission] =
     httpClient
-      .POSTEmpty[Submission](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.declarations}/$id/submission")
+      .POSTEmpty[Submission](s"${appConfig.customsDeclareExportsBaseUrl}/submission/$id")
+
+  def fetchSubmissions()(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Seq[Submission]] =
+    httpClient.GET[Seq[Submission]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.submissions}").map { response =>
```
udinobi, line 123: Why using here `${appConfig.submissions}` and using instead `/submission` at line 120 ? Or viceversa.
I get they are different EPs. Only why the different syntax?

app/connectors/CustomsDeclareExportsConnector.scala
```diff
@@ -117,25 +117,30 @@ class CustomsDeclareExportsConnector @Inject()(appConfig: AppConfig, httpClient:
...
+      .map(_.headOption)
+
+  def findSubmissionsByLrn(lrn: Lrn)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Seq[Submission]] =
+    httpClient
+      .GET[Seq[Submission]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.submissions}", Seq("lrn" -> lrn.value))
 
   def findNotifications(id: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Seq[Notification]] =
     httpClient
-      .GET[Seq[Notification]](s"${appConfig.customsDeclareExportsBaseUrl}${appConfig.declarations}/$id/submission/notifications")
+      .GET[Seq[Notification]](s"${appConfig.customsDeclareExportsBaseUrl}/submission/notifications/$id")
```
udinobi, line 139: Same here, in relation to lines 130 and 135

app/forms/LrnValidator.scala
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.hmrc.http.HeaderCarrier
+
+import java.time.ZonedDateTime
+import javax.inject.Inject
+import scala.concurrent.{ExecutionContext, Future}
+
+class LrnValidator @Inject()(customsDeclareExportsConnector: CustomsDeclareExportsConnector) {
+
+  def hasBeenSubmittedInThePast48Hours(lrn: Lrn)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Boolean] = {
+    val now = ZonedDateTime.now(Action.defaultDateTimeZone)
```
udinobi, line 30: Shouldn't this statement, to be precise, be moved inside method `isSubmissionOlderThan48Hours` ?

udinobi:
Personal opinion... with the implicit class and directly calling the connector from the `form` package, the solution seems a bit over-engineered, not straightforward. But maybe it's just me.

app/forms/LrnValidator.scala
MaciejRewera: This way all Submissions are compared to the same moment in time. If it was as you're suggesting, then every Submission would be compared against a slightly different moment.
The difference is small, but this how I felt would be more appropriate.

MaciejRewera:
I was looking at it from responsibility perspective - should it be controller or form class responsible for validation? In my opinion it should be the form. It is (too) easy to put too many things into controller classes, maybe because it seems simpler. But I was aiming more towards modular structure and having single-responsiblity principle in mind, than towards being straightforward.

app/connectors/CustomsDeclareExportsConnector.scala
MaciejRewera: The only reason is legacy. Initially, EPs were configurable, because we assumed we may want or need to change them in different environments. I don't think this is the case anymore, but the code stays.
If you think it is a good idea to stop using them via `AppConfig`, then I am happy to refactor this part.

udinobi: What I would suggest is to choose one single approach for all EPs and not a mixed one.

Personally I like to have in `application.conf` `customsDeclareExportsBaseUrl` and the rest of the path directly specified in the connector, because to have it in `application.conf` does not add any value, is less readable and make the impl more complicated. And in any case you have to redeploy if paths need to be modified. I would have paths in `application.conf` only if they were used in multiple places.

But, as long as we follow for all EPs the same approach, I'm fine even to have paths in `application.conf`.

app/forms/LrnValidator.scala
udinobi: You're right. I forgot the requirement... search between all submissions with that LRN if one of them is older then 48h from now

udinobi:
I got that. But for the same reason, following your logic, my question then is... should it be the controller or the form class responsible for calling the connector? I do not think should be the form class.

udinobi (approved):

PR #1311, opened 2021-09-16
[CEDS-3469] Validation for missing status reason for specific codes

glidester:
The 'status reason' field is now a mandatory field if the user selects
certain 'document code's or certain 'status code's. Otherwise it
remains an optional field with same content validation rules.


app/forms/AdditionalConstraintsMapping.scala
```diff
@@ -0,0 +1,61 @@
...
+import uk.gov.voa.play.form.Condition
+
+case class ConditionalConstraint[T](shouldConstraintBeApplied: Condition, errorKey: String, constraint: (T => Boolean))
+
+/**
+  * A mapping that extends another mapping with additional constraints that are only added if their
+  * conditions are satisfied.
+  *
+  * @param baseMappings the wrapped mapping that should always be applied.
+  * @param conditionalAdditionalConstraints a Seq of additional constraints to be applied, only the first one who's
```
udinobi, line 30: Nit: the explanation is not clear. Maybe `...only **if** the first one condition...`

app/forms/AdditionalConstraintsMapping.scala
```diff
@@ -0,0 +1,61 @@
...
+  *                                         condition is satisfied is added (the others are ignored).
+  */
+case class AdditionalConstraintsMapping[T](
```
udinobi, line 33: Would not a `class` (not a `case class`) be a better choice here, considering its complexity?

glidester: I'm just following the other established implementation pattern such as play.api.data.OptionalMapping

app/forms/AdditionalConstraintsMapping.scala
glidester: I've updated it to: "a Seq of additional constraints to be applied. Only the first constraints in the sequence who's condition is satisfied is added (the others are ignored)." Any clearer?

udinobi (approved):

PR #1315, opened 2021-09-20
[CEDS-3402] New_content_at_base_of_CYA_page

glidester:
Adding warning text and changing 'Continue' button to
be a link.

Also refactoring summary page tests to have one per
'mode'.

### DIFF
--- a/app/views/declaration/summary/amend_summary_page.scala.html
+++ b/app/views/declaration/summary/amend_summary_page.scala.html
@@ -21,9 +21,11 @@
 @import views.helpers.{BackButton, Title}
 @import views.html.components.gds.{linkButton, _}
 @import views.html.declaration.summary._
+@import uk.gov.hmrc.govukfrontend.views.html.components._
 
 @this(
         govukLayout: gdsMainTemplate,
+        govukWarningText: GovukWarningText,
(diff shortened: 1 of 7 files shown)


app/views/declaration/summary/amend_summary_page.scala.html
```diff
@@ -35,7 +37,8 @@
         items_section: items_section,
         warehouse_section: warehouse_section,
         transport_section: transport_section,
-        linkButton: linkButton
+        label: label,
+        link: link
```
MaciejRewera, line 41: I don't understand this change. Why the `linkButton` component is not sufficient? It's purpose was to have a link that looks like a button.

glidester: The AC on the ticket is "Change the 'Continue' buttons to a link on both pages", so we are replacing the button looking link with a link looking link :wink:

MaciejRewera (approved):

app/views/declaration/summary/amend_summary_page.scala.html
MaciejRewera: Okay, didn't think about that. No probs for me then.

PR #1317, opened 2021-09-20
CEDS-3456 Fix tabnabbing

MaciejRewera:


glidester (approved):
LGTM

udinobi:
`detailsContentLink` template also has `target="_blank"`

app/views/components/gds/link.scala.html
```diff
@@ -27,7 +27,7 @@
 
 @hasHint = @{textHidden.exists(_.nonEmpty)}
 
-<a class="@classes" href="@call" target=@target @id.map{id => id="@id" }>
- <span @if(hasHint){aria-hidden="true"}>@text</span>
- @if(hasHint){<span class="govuk-visually-hidden">@{textHidden.getOrElse("")}</span>}
-</a>
```
glidester, line 33: Maybe this new extra line is the cause of the space?

app/views/components/gds/tariffLink.scala.html
```diff
@@ -14,6 +14,14 @@
  * limitations under the License.
  *@
 
+@import views.html.components.gds.link
+
+@this(link: link)
+
 @(linkText: String, linkUrl: String)
 
-<a class="govuk-link govuk-link--no-visited-state" target="_blank" href="@linkUrl">@linkText</a>
```
glidester, line 19: ... and here too

app/views/declaration/not_declarant.scala.html
```diff
@@ -23,15 +23,18 @@
...
 )
 
 @()(implicit request: Request[_], messages: Messages)
 
 @paragraph1Link = {<a href=@controllers.routes.RootController.displayPage.url>@messages("notDeclarent.paragraph1Link")</a>}
 @paragraph2Link = {<a href="https://www.gov.uk/guidance/dispatching-your-goods-within-the-eu">@messages("notDeclarent.paragraph2Link")</a>}
 @supportPhone = {<b>0300 200 3700</b>}
-@enquiriesLink = {<a href="https://www.gov.uk/government/organisations/hm-revenue-customs/contact/customs-international-trade-and-excise-enquiries" target="_blank" rel="noopener noreferrer">@messages("general.inquiries.help.link")</a>}
+@enquiriesLink = {
+  @link(text = messages("general.inquiries.help.link"), call = Call("GET", "https://www.gov.uk/government/organisations/hm-revenue-customs/contact/customs-international-trade-and-excise-enquiries"), target = "_blank")
```
udinobi, line 36: It's not that I'm against to have the url hardcoded here, but it seems to me at odds with the logic we usually follow... the urls are in `application.conf`. And this is the same for all the other urls I'm seeing hardcoded in the templates in this PR (I do not want to repeat the same comment)

udinobi (approved):

app/views/components/gds/link.scala.html
```diff
@@ -27,7 +27,5 @@
...
\ No newline at end of file
+<a class="@classes" href="@call" target=@target @id.map{id => id="@id" } @if(target == "_blank"){rel="noopener noreferrer"}
```
glidester, line 30: Would be wise to add a comment here informing future devs not to split the `<a>` tag into multiple lines. The effect on messages with links is far from obvious so its likely they would not realise (and the problem will then return) :zombie:

udinobi: :muscle:

MaciejRewera: I'll add that, but if you try to format this piece of code, then you'd have a few tests failing. So you'll definitely notice ;-)

glidester (approved):

PR #1320, opened 2021-09-22
CEDS-3484 - Removed duplicate tracking consent reference

oddbunchmedia:
Removing duplicate tracking consent script from the layout:

Screenshot from 2021-09-22 09-43-02

Screenshot from 2021-09-22 09-55-32

This is called in the HmrcHead now, so no need for a standalone reference unless user is no longer signed in.


glidester (approved):
Looks good. Could you though perhaps rename the val `useTimeoutDialog` to something more generic like `userIsSignedIn`, would make the if-else statement easier to comprehend :slightly_smiling_face:

glidester (approved):

PR #1326, opened 2021-09-23
CEDS-3486 - Change layout of multiple para in hints

oddbunchmedia:
Changing how hint render multiple paras in the frontend. Instead of multiple paras with the same ID I have wrapped the hint in a div with the ID on.

This is closer to the GDS standard as shown in the design guide:
https://design-system.service.gov.uk/components/text-input/

They do also say "Do not use long paragraphs and lists in hint text. Screen readers read out the entire text when users interact with the form element. This could frustrate users if the text is long."


udinobi (approved):

app/views/components/fields/field_accessible_autocomplete_with_header.scala.html
```diff
@@ -63,8 +63,12 @@
 
     @allErrors.map { error => <span id="error-message-@{field.name}-input" class="govuk-error-message">@error</span> }
 
-    @hintParagraphs.map { hint =>
-        <p class="govuk-hint" id="@(fieldNameWithoutBrackets)-hint">@hint</p>
+    @if(hintParagraphs) {
```
MaciejRewera, line 66: I have to admit that I am surprised this is working, since Scala requires `Boolean` type inside `if` statement. I'm guessing this is due to having Twirl here rather than Scala.

MaciejRewera (dismissed):
LGTM

oddbunchmedia:
Just looking at this conflict...

app/views/components/fields/field_accessible_autocomplete_with_header.scala.html
```diff
@@ -66,8 +66,12 @@
 
     @bodyHtml.map(identity)
 
-    @hintParagraphs.map { hint =>
-        <p class="govuk-hint" id="@(fieldNameWithoutBrackets)-hint">@hint</p>
+    @if(hintParagraphs.nonEmpty) {
```
udinobi, line 69: Why not simply...
```
   @hintParagraphs.map { hint =>
        <div id="@(fieldNameWithoutBrackets)-hint">
            <p class="govuk-hint">@hint</p>
        </div>
    }
```

oddbunchmedia: Where there are multiple paragraphs I'd just end up with each paragraph wrapped in a div with the same ID - I'm trying to wrap all the paragraphs in one div.

MaciejRewera (approved):

app/views/components/fields/field_accessible_autocomplete_with_header.scala.html
udinobi: Yeah, sorry my bad. I supposed `hintParagraphs` was an `Option`, and not a `List`.

PR #1327, opened 2021-09-23
[CEDS-3478] Validation update to total-numbers-of-items

glidester:
Allow commas and periods in numerical values and limit
total number of digits and decimal places for each
field.

### DIFF
--- a/app/forms/declaration/TotalNumberOfItems.scala
+++ b/app/forms/declaration/TotalNumberOfItems.scala
@@ -30,18 +30,37 @@ object TotalNumberOfItems extends DeclarationPage {
   implicit val format = Json.format[TotalNumberOfItems]
 
   val formId = "TotalNumberOfItems"
+  val exchangeRate = "exchangeRate"
+  val totalAmountInvoiced = "totalAmountInvoiced"
 
-  val totalAmountInvoicedPattern = "[0-9]{1,14}|[0-9]{1,14}[.][0-9]{1,2}"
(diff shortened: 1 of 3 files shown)


app/forms/declaration/TotalNumberOfItems.scala
```diff
@@ -30,18 +30,34 @@ object TotalNumberOfItems extends DeclarationPage {
...
+
+  val removeCommasFirst = (validator: String => Boolean) => (input: String) => validator(input.replaceAll(",", ""))
+  val notJustCommas = (input: String) => !input.forall(_.equals(','))
 
   val mapping = Forms.mapping(
-    "exchangeRate" -> optional(
+    exchangeRate -> optional(
       text()
-        .verifying("declaration.exchangeRate.error", isEmpty or ofPattern(exchangeRatePattern))
+        .verifying(rateFieldErrorKey, isEmpty or (notJustCommas and removeCommasFirst(ofPattern(exchangeRatePattern))))
```
udinobi, line 56: Maybe a comment to explain how `notJustCommas` and `removeCommasFirst` combine logically? (At least for me) it's not easy to grasp. As an example of me not understanding... why not checking if the resulting string is empty once `removeCommasFirst` is applied?

glidester: The reason is that this field allows empty values (its an optional field). If a user types ",,," as the value and then we use `removeCommasFirst(isEmpty)` the validation would allow the nonsensical numeric value of ",,," through. So instead I check if the field is truly empty (that is allowed) or else that the field isn't just commas and is a numeric value in range. I'll try and add a comment explaining this

udinobi (approved):

PR #1331, opened 2021-09-27
CEDS-3438 Add new validation check to page at '/add-authorisation-required'

udinobi:
1st commit is just renaming package `controllers.util` to `controllers.helpers` for uniformity with `views.helpers`

2nd commit adds a new validation check (EXRR authorisation code with pre-lodged decls is not allowed) to the page at  `/add-authorisation-required`

### DIFF
--- a/app/controllers/helpers/DeclarationHolderHelper.scala
+++ b/app/controllers/helpers/DeclarationHolderHelper.scala
@@ -16,43 +16,29 @@
 
 package controllers.helpers
 
-import forms.declaration.additionaldeclarationtype.AdditionalDeclarationType._
 import forms.declaration.declarationHolder.DeclarationHolder
 import models.requests.JourneyRequest
-import play.api.data.FormError
+import play.api.data.{Form, FormError}
 
 object DeclarationHolderHelper {
 
   val DeclarationHolderFormGroupId: String = "declarationHolder"
 
-  def cachedHolders(implicit request: JourneyRequest[_]): Seq[DeclarationHolder] =
-    request.cacheModel.parties.declarationHoldersData.map(_.holders).getOrElse(Seq.empty)
+  def declarationHolders(implicit request: JourneyRequest[_]): Seq[DeclarationHolder] = request.cacheModel.declarationHolders
 
-  def validateAuthCode(maybeDeclarationHolder: Option[DeclarationHolder])(implicit r: JourneyRequest[_]): Option[FormError] =
-    maybeDeclarationHolder match {
-      case Some(DeclarationHolder(Some(authorisationCode), _, _)) =>
-        if (isExrrSelectedForPrelodgedDecl(authorisationCode, r.cacheModel.additionalDeclarationType))
-          Some(FormError(DeclarationHolderFormGroupId, "declaration.declarationHolder.EXRR.error.prelodged"))
-        else
-          validateMutuallyExclusiveAuthCodes(authorisationCode, cachedHolders)
-
-      case _ => None
-    }
-
-  private val nonExrrAdditionalDeclarationTypes =
-    List(STANDARD_PRE_LODGED, SIMPLIFIED_PRE_LODGED, OCCASIONAL_PRE_LODGED, CLEARANCE_PRE_LODGED)
-
-  private def isExrrSelectedForPrelodgedDecl(authorisationCode: String, declarationType: Option[AdditionalDeclarationType]): Boolean =
-    declarationType.fold(false)(authorisationCode == "EXRR" && nonExrrAdditionalDeclarationTypes.contains(_))
+  def form(implicit request: JourneyRequest[_]): Form[DeclarationHolder] =
+    DeclarationHolder.form(request.eori, request.cacheModel.additionalDeclarationType)
 
   private val mutuallyExclusiveAuthorisationCodes = List("CSE", "EXRR")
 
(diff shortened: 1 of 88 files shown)


app/controllers/helpers/DeclarationHolderHelper.scala
```diff
@@ -16,13 +16,43 @@
...
+  private val mutuallyExclusiveAuthorisationCodes = List("CSE", "EXRR")
+
+  private def validateMutuallyExclusiveAuthCodes(authorisationCode: String, holders: Seq[DeclarationHolder]): Option[FormError] =
+    if (!mutuallyExclusiveAuthorisationCodes.contains(authorisationCode)) None
+    else {
+      val mustNotAlreadyContainCodes = mutuallyExclusiveAuthorisationCodes.filter(_ != authorisationCode)
+
+      if (!holders.map(_.authorisationTypeCode.getOrElse("")).containsSlice(mustNotAlreadyContainCodes)) None
+      else Some(FormError(DeclarationHolderFormGroupId, s"declaration.declarationHolder.${authorisationCode}.error.exclusive"))
+    }
```
MaciejRewera, line 57: This is a kind of validation, so I'd place it in the `Form` class and this is what we used to do in past.
This way there is no need to call it explicitly in addition to regular form binding.

test/controllers/declaration/DeclarationHolderChangeControllerSpec.scala
```diff
@@ -133,25 +133,56 @@ class DeclarationHolderChangeControllerSpec extends ControllerSpec with OptionVa
...
+            "eori" -> declarationHolder2.eori.map(_.value).get,
+            "eoriSource" -> declarationHolder2.eoriSource.map(_.toString).get
+          )
           val result = controller.submitForm(Mode.Normal, declarationHolder1.id)(postRequestAsFormUrlEncoded(requestBody: _*))
 
           status(result) mustBe BAD_REQUEST
           verifyChangePageInvoked()
         }
+
+        "user adds mutually exclusive data" when {
```
MaciejRewera, line 147: IMHO these tests should be done for the `DeclarationHolderHelper` class or `Form` class if you decide to move the validation there.
It is not controller's  responsibility to validate the input, only to invoke that validation. Besides, right now the test cases are duplicated.

app/controllers/helpers/DeclarationHolderHelper.scala
```diff
@@ -0,0 +1,44 @@
...
+  def validateMutuallyExclusiveAuthCodes(maybeHolder: Option[DeclarationHolder], holders: Seq[DeclarationHolder]): Option[FormError] =
+    maybeHolder match {
+      case Some(DeclarationHolder(Some(code), _, _)) if mutuallyExclusiveAuthorisationCodes.contains(code) =>
+        val mustNotAlreadyContainCodes = mutuallyExclusiveAuthorisationCodes.filter(_ != code)
+
+        if (!holders.map(_.authorisationTypeCode.getOrElse("")).containsSlice(mustNotAlreadyContainCodes)) None
+        else Some(FormError(DeclarationHolderFormGroupId, s"declaration.declarationHolder.${code}.error.exclusive"))
+
+      case _ => None
+    }
```
glidester, line 43: Could this method be moved into the `DeclarationHolder` companion object? Then at least all the validation logic would be in one place :slightly_smiling_face:

udinobi: Actually no. The CSE-EXRR validation must only take place when adding an authorisation, and not when changing one. The `form` can only include validations supporting both pages (add and change).

Liz's words...
> I think that I would allow this to happen. The user will get a DMSREJ. We cannot protect them from every possible thing they want to do wrong

glidester: Can't see a unit test for this method, think it would be useful to test this is isolation (rather than relying on the controller's unit tests)

app/controllers/helpers/DeclarationHolderHelper.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  private val mutuallyExclusiveAuthorisationCodes = List("CSE", "EXRR")
+
+  def validateMutuallyExclusiveAuthCodes(maybeHolder: Option[DeclarationHolder], holders: Seq[DeclarationHolder]): Option[FormError] =
+    maybeHolder match {
+      case Some(DeclarationHolder(Some(code), _, _)) if mutuallyExclusiveAuthorisationCodes.contains(code) =>
+        val mustNotAlreadyContainCodes = mutuallyExclusiveAuthorisationCodes.filter(_ != code)
+
+        if (!holders.map(_.authorisationTypeCode.getOrElse("")).containsSlice(mustNotAlreadyContainCodes)) None
+        else Some(FormError(DeclarationHolderFormGroupId, s"declaration.declarationHolder.${code}.error.exclusive"))
```
glidester, line 40: Could this logic be simplified (by removing the `filter` and the `containsSlice` parts)?: 
`case Some(DeclarationHolder(Some(code), _, _)) if mutuallyExclusiveAuthorisationCodes.contains(code) =>

        if (holders.map(_.authorisationTypeCode.getOrElse("")).contains(code))
          Some(FormError(DeclarationHolderFormGroupId, s"declaration.declarationHolder.${code}.error.exclusive"))
        else
          None`

glidester (approved):
Looks fine, just would be good to put all validation in one class/file though :slightly_smiling_face:

glidester (approved):
:+1:

PR #1332, opened 2021-09-27
CEDS-3490 - Changes to content for A11y

oddbunchmedia:


app/views/components/goodsLocationExpander.scala.html
```diff
@@ -34,106 +35,94 @@
...
-            )
+        content = HtmlContent(link(
+            text = text("paragraph01.link"),
+            call = Call("GET", appConfig.rollOnRollOffPorts),
+            target = "_blank"
         ))
     ))
 
-    @paragraphBody(text("paragraph04.title"), "govuk-label--s")
+    @heading(text("paragraph02.title"), "govuk-heading-s govuk-!-margin-top-4 govuk-!-margin-bottom-1", "h2")
```
glidester, line 47: Just wondering why we need all these styles for these headings (just never seen them used anywhere else for headings so slightly curios) :thinking:

oddbunchmedia: The content looked a bit weird with just the heading and a link:
Screenshot from 2021-09-27 12-00-03
So Liz asked if I could play about with the spacing a bit:
Screenshot from 2021-09-27 12-55-49

Not usually needed but in this case I think that was just down to the sparseness of the content that it needed tweeking.

glidester (approved):

PR #1336, opened 2021-09-28
[CEDS-3339] Change content and url of total-numbers-of-items page

glidester:


app/connectors/exchange/ExportsDeclarationExchange.scala
```diff
@@ -68,9 +73,10 @@ case class ExportsDeclarationExchange(
     locations = this.locations,
     items = this.items,
     totalNumberOfItems = this.totalNumberOfItems.flatMap { exchange =>
-      (exchange.totalAmountInvoiced, exchange.exchangeRate) match {
-        case (None, None)                        => None
-        case (totalAmountInvoiced, exchangeRate) => Some(TotalNumberOfItems(exchangeRate, totalAmountInvoiced))
+      (exchange.totalAmountInvoiced, exchange.totalAmountInvoicedCurrency, exchange.exchangeRate) match {
```
udinobi, line 76: Probably I miss something, but why do you need pattern matching here. Can't you directly be doing something like...
```
Some(TotalNumberOfItems(
    exchange.exchangeRate,
    exchange.totalAmountInvoiced,
    exchange.totalAmountInvoicedCurrency
))
```

app/forms/declaration/TotalNumberOfItems.scala
```diff
@@ -50,6 +55,14 @@ object TotalNumberOfItems extends DeclarationPage {
   val removeCommasFirst = (validator: String => Boolean) => (input: String) => validator(input.replaceAll(",", ""))
   val notJustCommas = (input: String) => !input.forall(_.equals(','))
 
+  val equalsIgnoreCaseOptionString = (value: String) => (input: Option[String]) => input.map(_.equalsIgnoreCase(value)).getOrElse(false)
+  val isEmptyOptionString = (input: Option[String]) => isEmpty(input.getOrElse(""))
+  val isAlphabeticOptionString = (input: Option[String]) => isAlphabetic(input.getOrElse(""))
+  val lengthInRangeOptionString = (min: Int) => (max: Int) => (input: Option[String]) => lengthInRange(min)(max)(input.getOrElse(""))
+
+  def isFieldEmpty(field: String): Condition = _.get(field).map(_.isEmpty()).getOrElse(true)
+  def isFieldNotEmpty(field: String): Condition = _.get(field).map(!_.isEmpty()).getOrElse(false)
```
udinobi, line 64: Would it be worth to move these validation methods (lines 58-64) to `FieldValidator.scala` ?

app/views/declaration/total_number_of_items.scala.html
```diff
@@ -26,19 +29,117 @@
...
+        )
+    ))
+
+    <span class="govuk-hint" id="totalAmountInvoiced-hint">
+        @Html(messages("declaration.totalAmountInvoiced.hint"))
+    </span>
+
+
+    @if(totalAmountInvoicedErrors.nonEmpty){
+        <span class="govuk-error-message">
```
udinobi, line 88: Just a confirmation. If `totalAmountInvoicedErrors` is not empty , will this `span` tag and the one at line 82 be shown inlined ? And, if yes, is this what we want?

app/views/declaration/total_number_of_items.scala.html
```diff
@@ -26,19 +29,117 @@
...
+@currencyCodeAndAmount = {
+    <div class="govuk-form-group @if(totalAmountInvoicedErrors.nonEmpty){govuk-form-group--error}">
+
+    @govukFieldset(Fieldset(
+        legend = Some(Legend(
+            content = Text(messages("declaration.totalAmountInvoiced")),
+            classes = "govuk-fieldset__legend--m"
+        )),
+        attributes = Map("id" -> "currencyCodeAndAmount"),
+        html = HtmlFormat.fill(immutable.Seq(currencyCodeAndAmountContent))
```
udinobi, line 121: Could we start using `List` instead of `immutable.Seq`?

app/connectors/exchange/ExportsDeclarationExchange.scala
glidester: Good question. On closer inspection it appears I've broken the original intent of the code to not populate the field with a `Some(TotalNumberOfItems(None, None, None))`. I'll correct this :+1:

app/views/declaration/total_number_of_items.scala.html
glidester: TBH I'm not sure. This pattern was taken from the existing /additional-documentation page which has been through a DAC assessment so I'm assuming the HTML structure is correct. :crossed_fingers: :slightly_smiling_face:

udinobi (approved):

PR #1346, opened 2021-10-05
[CEDS-3480] Consignment references LRN text update

glidester:


test/views/declaration/summary/ReferencesSectionViewSpec.scala
```diff
@@ -75,15 +76,32 @@ class ReferencesSectionViewSpec extends UnitViewSpec with ExportsTestData with I
...
 
-      row must haveSummaryActionsHref(routes.ConsignmentReferencesController.displayPage(Mode.Change))
+        row must haveSummaryActionsHref(routes.ConsignmentReferencesController.displayPage(Mode.Change))
+      }
+    }
+
+    onJourney(SUPPLEMENTARY) { implicit request =>
+      "have LRN with change button" in {
+
+        val view = section(Mode.Change, data.copy(`type` = SUPPLEMENTARY))(messages)
```
MaciejRewera, line 95: Nit: Since you are using `onJourney` method, there is no point in explicitly specifying it in the test. You can simply take the journey type from the request.

test/views/declaration/summary/ReferencesSectionViewSpec.scala
```diff
@@ -75,15 +76,32 @@ class ReferencesSectionViewSpec extends UnitViewSpec with ExportsTestData with I
...
+        row must haveSummaryKey(messages("declaration.summary.references.lrn"))
+        row must haveSummaryValue("LRN")
 
-      row must haveSummaryActionsTexts("site.change", "declaration.summary.references.lrn.change")
+        row must haveSummaryActionsTexts("site.change", "declaration.summary.references.lrn.change")
 
-      row must haveSummaryActionsHref(routes.ConsignmentReferencesController.displayPage(Mode.Change))
+        row must haveSummaryActionsHref(routes.ConsignmentReferencesController.displayPage(Mode.Change))
+      }
+    }
```
MaciejRewera, line 90: Actually, in this test case you don't have different journeys, because you are using the `view` value. Which is defined once with Standard journey data

MaciejRewera (commented):
Tests may need a bit of care. They don't test what you'd expect them to.

MaciejRewera (approved):

PR #1349, opened 2021-10-06
CEDS-3353 - Moving hint to para

oddbunchmedia:


app/views/declaration/link_ducr_to_mucr.scala.html
```diff
@@ -71,16 +72,19 @@
...
+        @paragraph(messages("declaration.linkDucrToMucr.paragraph"))
+
         @yesNoRadios(
             form = form,
             heading = Some(Fieldset(
                 legend = Some(Legend(
                     content = Text(messages("declaration.linkDucrToMucr.title")),
-                    isPageHeading = true,
-                    classes = gdsPageLegend
+                    isPageHeading = false,
```
udinobi, line 84: Nit: this is the default, if I'm not wrong, so I would remove it.

PR #1350, opened 2021-10-06
CEDS-3354 - Change hint for para

oddbunchmedia:


app/views/declaration/mucr_code.scala.html
```diff
@@ -47,17 +49,15 @@
...
             labelKey = "declaration.mucr.title",
-            hintKey = Some("declaration.mucr.hint"),
-            hintKeyLinks = List(link(
-                text = messages("declaration.mucr.hint.link"),
-                call = Call("GET", appConfig.notesForMucrConsolidationUrl),
-                target = "_blank"
-            )),
-            isPageHeading = true,
-            headingClasses = gdsPageLabel,
+            isPageHeading = false,
```
udinobi, line 59: Nit: this is the default, if I'm not wrong, so I would remove it.

PR #1359, opened 2021-10-12
CEDS-3311 - Radio option orientation

oddbunchmedia:


app/views/declaration/declarant_details.scala.html
```diff
@@ -67,6 +67,7 @@
 
         @govukRadios(Radios(
             name = isEoriKey,
+            classes = "govuk-radios--inline",
```
glidester, line 70: Could this use be replaced with the `yesNoRadios` component?

PR #1366, opened 2021-10-19
CEDS-3464 Confirmation page's content according to the notification status

udinobi:


app/controllers/declaration/ConfirmationController.scala
```diff
@@ -16,31 +16,101 @@
...
 
-  val displayHoldingConfirmation: Action[AnyContent] = authenticate { implicit request =>
-    Ok(holdingConfirmationPage())
+  val displayHoldingConfirmation: Action[AnyContent] = authenticate.async { implicit request =>
+    request.getQueryString(js) match {
+
+      // Show page at /holding and wait a few secs.
+      case None =>
+        val holdingUrl = routes.ConfirmationController.displayHoldingConfirmation.url
+        Future.successful(Ok(holdingConfirmationPage(s"$holdingUrl?$js=$Disabled", s"$holdingUrl?$js=$Enabled")))
```
glidester, line 56: Nit: feels a little odd for the controller to pass these two values to the view, couldn't the view just be passes the `holdingUrl` and specialise it for its own needs?

app/controllers/declaration/ConfirmationController.scala
```diff
@@ -16,31 +16,101 @@
...
   mcc: MessagesControllerComponents,
+  errorHandler: ErrorHandler,
   draftConfirmationPage: draft_confirmation_page,
   holdingConfirmationPage: holding_confirmation_page,
   submissionConfirmationPage: submission_confirmation_page
-) extends FrontendController(mcc) with I18nSupport {
+)(implicit ec: ExecutionContext) extends FrontendController(mcc) with I18nSupport with Logging {
 
   val displayDraftConfirmation: Action[AnyContent] = authenticate { implicit request =>
     Ok(draftConfirmationPage())
```
glidester, line 47: Just wondering if we should now separate out the draft confirmation journey from the submitted declaration as they are two very different things (and I think the draft page just adds noise to this controller's main purpose IMO)

PR #1367, opened 2021-10-19
[CEDS-3532] Refactor list of countries loading to support internation…

glidester:
…alisation

Made 'Country' entity a 'CommonCode' and updated its source file to be a well formatted json file.

Removed the EU and special country lists as they were not used anywhere.


app/services/Countries.scala
```diff
@@ -16,46 +16,32 @@
...
 package services
 
-import play.api.libs.json._
-import services.model.Country
-import utils.FileReader
+import connectors.CodeListConnector
+import models.codes.Country
+import play.api.i18n.Messages
 
 object Countries {
```
udinobi, line 23: Maybe it's a wrong idea (I did not check if it's really possible) but couldn't `Countries` changed to be a singleton class so that you can directly inject `CodeListConnector`, and maybe `Messages` too, so that you do not need to pass it everywhere?

PR #1376, opened 2021-11-02
CEDS3540 - Error string prepended to title

oddbunchmedia:


app/views/helpers/Title.scala
```diff
@@ -18,14 +18,28 @@ package views.helpers
...
+  headingArg: String = "",
+  headingArgs: Option[Seq[String]] = None,
+  hasErrors: Boolean = false
+) {
 
   def toString(implicit messages: Messages): String = {
     def args = headingArgs.getOrElse(Seq(headingArg))
     if (sectionKey.isEmpty) {
-      messages("title.format", messages(headingKey, args: _*), messages("service.name"))
+      if (hasErrors) {
```
udinobi, line 32: What about something a bit more DRY...
```
val key = if (hasErrors) "title.hasErrors.format" else "title.format"
messages(key, messages(headingKey, args: _*), messages("service.name"))
```

app/views/helpers/Title.scala
```diff
@@ -18,14 +18,28 @@ package views.helpers
...
+        messages("title.hasErrors.format", messages(headingKey, args: _*), messages("service.name"))
+      } else {
+        messages("title.format", messages(headingKey, args: _*), messages("service.name"))
+      }
     } else {
-      messages("title.withSection.format", messages(headingKey, args: _*), messages(sectionKey), messages("service.name"))
+      if (hasErrors) {
```
udinobi, line 38: Even better...
```
val key = if (hasErrors) ".hasErrors" else "" // This could be used for both `if` test
messages(s"title$key.withSection.format", messages(headingKey, args: _*), messages("service.name"))
```

app/views/helpers/Title.scala
oddbunchmedia: Nice - changed the code :)

udinobi (approved):

PR #1386, opened 2021-11-09
[CEDS-3510] Rewrite headings with labels so they are semantically correct

glidester:
Removed the ability of a Label to contain a heading (as they are two
distinctly different things)


udinobi (approved):
Completely agree with the ticket's goal. :+1:

oddbunchmedia (approved):
LGTM

PR #1393, opened 2021-11-15
CEDS-3459 new configuration of drop down type ahead

glidester:
Adding non-visible country aliases to the country drop down
form field element.

Also cleaning up remnant country code CSV files.


app/views/components/fields/field_accessible_location_autocomplete_with_header.scala.html
```diff
@@ -0,0 +1,83 @@
...
+@import views.html.helper._
+@(field: Field,
+  label: String,
+  labelClass: Option[String],
+  section: Option[String] = None,
+  bodyHtml: Option[Html] = None,
+  hintParagraphs: List[String] = List.empty,
+  emptySelectValue: String,
+  items: List[AutoCompleteItem],
+  isPageHeading: Boolean,
```
udinobi, line 29: Shouldn't we remove pageHeading from components?

app/views/components/fields/field_accessible_location_autocomplete_with_header.scala.html
```diff
@@ -0,0 +1,83 @@
...
+}
+
+<div id="@fieldNameWithoutBrackets-autocomplete-outer" class="govuk-form-group @if(allErrors.nonEmpty) {govuk-form-group--error}">
+
+    @section.map { key =>
+        <span id="section-header" class="govuk-caption-xl">@messages(key)</span>
+    }
+
+    @if(isPageHeading) {<h1 class="govuk-label-wrapper">}
+    <label id="@(fieldNameWithoutBrackets)-label" for="@fieldNameWithoutBrackets" class="govuk-label@if(labelClass.nonEmpty){ @labelClass}" >
```
udinobi, line 60: `isPageHeading` for a `<label>` tag?

app/views/helpers/CountryAutoCompleteHelper.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import javax.inject.{Inject, Singleton}
+
+@Singleton
+class CountryAutoCompleteHelper @Inject()(codeListConnector: CodeListConnector, codeLinkConnector: CodeLinkConnector) {
+  def generateAutocompleteEnhancementJson()(implicit messages: Messages) = {
+    val jsObjects = for {
+      country <- codeListConnector.getCountryCodes(messages.lang.toLocale).values
+    } yield {
+      val aliases = codeLinkConnector.getAliasesForCountryCode(country.countryCode).getOrElse(Seq())
```
udinobi, line 33: Nit: `Seq()` -> `Seq.empty`

app/views/components/fields/field_accessible_location_autocomplete_with_header.scala.html
glidester: I spoke to Tony about this when playing CEDS-3510. He said it is ok for a heading to contain a label, but not for a label to contain a heading. So this aligns with that policy.

app/views/components/fields/field_accessible_location_autocomplete_with_header.scala.html
glidester: Should we? This is used on those pages that only have a single form country element with no body text (so just heading then form field component) such as country of routing.

udinobi: I have seen amendments by Tony where he was removing the pageHeading from the components.

Ideally I would not group components, but that's just my personal opinion. I would not argue if you choose to keep it.

udinobi (approved):

app/views/components/fields/field_accessible_location_autocomplete_with_header.scala.html
glidester: As its not the primary purpose of this ticket I won't attempt to remove it now (would also need to update the original field_accessible_autocomplete_with_header.scala.html template at the same time). Maybe a separate UI ticket could be created for this work.

oddbunchmedia (approved):

PR #1398, opened 2021-11-17
CEDS-3526 Dynamic content for the page at '/supplementary-units'

udinobi:
Dynamic content, retrieved from the Tariff API after the given commodity code, for the page at '/supplementary-units'.

If the expected content cannot be extracted from the Tariff API's Json payload, the default static page is shown instead.


app/controllers/declaration/SupplementaryUnitsController.scala
```diff
@@ -18,77 +18,112 @@ package controllers.declaration
...
+    SupplementaryUnits.form(yesNoPage).withSubmissionErrors
 
   private def hasSupplementaryUnits(commodityMeasure: CommodityMeasure): Boolean =
     commodityMeasure.supplementaryUnits.isDefined || commodityMeasure.supplementaryUnitsNotRequired.isDefined
 
-  private def updateExportsCache(itemId: String, updatedItem: SupplementaryUnits)(
-    implicit r: JourneyRequest[AnyContent]
-  ): Future[Option[ExportsDeclaration]] =
+  private def retrieveCommodityInfoIfAny(itemId: String)(implicit request: JourneyRequest[_]): Future[Option[CommodityInfo]] = {
+    if (tariffApiConfig.isCommoditiesEnabled) {
```
glidester, line 98: Would it not be better for the TariffApiService to reason about the flag's status rather than this controller?

test/services/TariffApiServiceSpec.scala
```diff
@@ -0,0 +1,189 @@
...
+    val commodityDetails = CommodityDetails(code, Some("Description"))
+    val item = anItem(withCommodityDetails(commodityDetails))
+    tariffApiService.retrieveCommodityInfoIfAny(aDeclaration(withItems(item)), item.id)
+  }
+
+  private def retrieveCommodityInfoIfAny(json: String): Option[CommodityInfo] = {
+    val response = Future.successful(Some(Json.parse(json)))
+    when(tariffApiConnector.getCommodity(any())).thenReturn(response)
+    retrieveCommodityInfoIfAny().futureValue
+  }
```
glidester, line 60: Nit: could these two methods be named slightly differently from each other? Overriding the same name makes the code that calls them a bit confusing (which version is being called).

glidester (approved):

glidester (approved):
:+1:

PR #1402, opened 2021-11-19
CEDS-3488 Handling undeliverable / unverified email addresses (Exports UI)

magicaljellybeans:
Updated to handle undeliverable email responses from the backend.
Created new route and template.


app/controllers/UndeclaredEmailController.scala
```diff
@@ -0,0 +1,39 @@
...
+  authenticate: AuthAction,
+  mcc: MessagesControllerComponents,
+  undeclared_email: undeclared_email,
+  appConfig: AppConfig
+) extends FrontendController(mcc) with I18nSupport {
+
+  val informUser: Action[AnyContent] = authenticate { implicit req =>
+    Ok(undeclared_email(appConfig.emailFrontendUrl))
+  }
+}
```
glidester, line 39: Do we really need this extra controller? Could we not repurpose the existing `UnverifiedEmailController` to handle both situations (just calling different view templates depending on the issue)?

conf/messages.en
```diff
@@ -393,6 +393,16 @@ emailUnverified.bullets.item4=duty deferment Direct Debit notices
...
 
+Undeliverable.title=There’s a problem with the CDS registered email address
+Undeliverable.heading=There’s a problem with the CDS registered email address
+Undeliverable.p1=We tried to send you an email but it could not be delivered. This could be because the inbox is full, or due to a technical problem with your email provider.
+Undeliverable.p2=This is the email address your organisation has registered for the Customs Declaration Service (CDS). You need to verify this email address or change it. You can verify the email instantly provided you have access to the email account.
+Undeliverable.p3=This will be the only email address we use for:
+Undeliverable.list.1=updates on changes to the Customs Declaration Service
+Undeliverable.list.2=urgent updates about goods in customs
+Undeliverable.list.3=some financial notifications, including Direct Debit notices and VAT
+Undeliverable.link-text=Verify or change email address
```
glidester, line 404: Convention is to keep keys camel case (so not starting with upper case letter). Maybe `emailUndeliverable.???` would be a better choice to tie it in with the existing `emailUnverified.???` keys?

app/views/undeclared_email.scala.html
```diff
@@ -0,0 +1,50 @@
+@*
```
glidester, line 1: 'undeclared'?? Should this not be 'undeliverable'?

magicaljellybeans:
Thanks Tim, these are fixed too

glidester (approved):
:+1:

PR #1429, opened 2021-12-08
[CEDS-3628] Prevent user in TDR from trying to register for CDS

glidester:
Also prevent users who are not logged in from hitting an Exports page
and causing an ERROR level log output (that triggers PagerDuty alerts).

### DIFF
--- a/app/controllers/actions/AuthAction.scala
+++ b/app/controllers/actions/AuthAction.scala
@@ -16,28 +16,30 @@
 
 package controllers.actions
 
-import scala.concurrent.{ExecutionContext, Future}
-
 import com.google.inject.{ImplementedBy, Inject, ProvidedBy}
 import com.kenshoo.play.metrics.Metrics
+import config.AppConfig
 import controllers.routes
-import javax.inject.Provider
+import models.{IdentityData, SignedInUser}
 import models.AuthKey.{enrolment, identifierKey}
 import models.requests.AuthenticatedRequest
-import models.{IdentityData, SignedInUser}
-import play.api.mvc._
 import play.api.{Configuration, Logging}
-import uk.gov.hmrc.auth.core.retrieve.v2.Retrievals.{agentCode, _}
-import uk.gov.hmrc.auth.core.retrieve.~
+import play.api.mvc._
 import uk.gov.hmrc.auth.core.{NoActiveSession, _}
+import uk.gov.hmrc.auth.core.retrieve.v2.Retrievals._
+import uk.gov.hmrc.auth.core.retrieve.~
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.http.HeaderCarrierConverter
+
+import javax.inject.Provider
+import scala.concurrent.{ExecutionContext, Future}
 
 class AuthActionImpl @Inject()(
   override val authConnector: AuthConnector,
   eoriAllowList: EoriAllowList,
   mcc: MessagesControllerComponents,
-  metrics: Metrics
+  metrics: Metrics,
+  appConfig: AppConfig
 ) extends AuthAction with AuthorisedFunctions with Logging {
 
   implicit override val executionContext: ExecutionContext = mcc.executionContext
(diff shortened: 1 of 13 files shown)


test/views/UnauthorisedViewSpec.scala
```diff
@@ -16,47 +16,83 @@
 
 package views
 
-import base.Injector
+import base.{Injector, OverridableInjector}
+import config.featureFlags.{SecureMessagingConfig, TdrUnauthorisedMsgConfig}
```
udinobi, line 20: `SecureMessagingConfig` is not used.

test/views/UnauthorisedViewSpec.scala
```diff
@@ -16,47 +16,83 @@
...
+        link must containMessage("unauthorised.tdr.body.link")
+        link must haveHref(s"mailto:${messages("unauthorised.tdr.body.link")}")
+        link.attr("target") mustBe "_blank"
+      }
     }
   }
+
+  private def getUnauthorisedPageView(tdrEnabled: Boolean = false) = {
+    when(tdrUnauthorisedMsgConfig.isTdrUnauthorisedMessageEnabled).thenReturn(tdrEnabled)
+    val unauthorisedPage = injector.instanceOf[unauthorised]
```
udinobi, line 95: This definition could be changed as a class field. Any reason to have it here?

app/controllers/actions/AuthAction.scala
```diff
@@ -16,28 +16,51 @@
...
 import models.AuthKey.{enrolment, identifierKey}
 import models.requests.AuthenticatedRequest
-import models.{IdentityData, SignedInUser}
-import play.api.mvc._
 import play.api.{Configuration, Logging}
-import uk.gov.hmrc.auth.core.retrieve.v2.Retrievals.{agentCode, _}
-import uk.gov.hmrc.auth.core.retrieve.~
+import play.api.mvc._
 import uk.gov.hmrc.auth.core.{NoActiveSession, _}
+import uk.gov.hmrc.auth.core.retrieve.v2.Retrievals.{
```
darrylmurphy, line 29: Can these be imported at once by the _ or is it neater to use Retrievals Mixin for access?

test/config/featureFlags/TdrUnauthorisedMsgConfigSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+package config.featureFlags
+
+import base.UnitWithMocksSpec
+import com.typesafe.config.ConfigFactory
+import features.Feature
+import play.api.Configuration
+
+class TdrUnauthorisedMsgConfigSpec extends UnitWithMocksSpec {
+
+  private def buildSfusConfig(enabled: Boolean = false, key: Feature.Value = Feature.tdrUnauthorisedMessage) = {
```
udinobi, line 26: The method's name sounds weird to me here.

app/views/unauthorised.scala.html
```diff
@@ -17,44 +17,59 @@
...
+    }
+} else {
+    @govukLayout(title = Title("unauthorised.heading")) {
+        @pageTitle(messages("unauthorised.heading"))
 
-    @pageTitle(messages("unauthorised.heading"))
+        @paragraphBody(messages("unauthorised.paragraph.1"))
 
-    @howToSection
+        @bulletList(List(
```
udinobi, line 53: Could not the `bulletList` be extracted, placed under flag condition and assigned to a "var", so to have a single `govukLayout` that includes the "maybe bulletList var"?

This seem more DRY to me.

app/views/unauthorised.scala.html
```diff
@@ -17,44 +17,59 @@
...
-        call = Call("GET", config.eoriService)
-      ))),
-      link(
-        id = Some("access_cds_link"),
-        text = messages("unauthorised.paragraph.1.bullet.2.link"),
-        call = Call("GET", config.cdsRegister)
-      )
-    ))
-}
+@if(tdrUnauthorisedMsgConfig.isTdrUnauthorisedMessageEnabled) {
```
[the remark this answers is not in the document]
udinobi: More than the two views, I would like to propose my comment below :smiley:

app/controllers/actions/AuthAction.scala
udinobi: Indeed `import uk.gov.hmrc.auth.core.retrieve.v2.Retrievals._` seems to be possible. Any reason for this?

app/views/unauthorised.scala.html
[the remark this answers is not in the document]
glidester: Don't see any gain in splitting it into two views tbh, plus this way confines the logic to a single file rather than also polluting the controller unnecessarily :slightly_smiling_face:

app/views/unauthorised.scala.html
glidester: But the two `govukLayout` instances are passed two very different title values and content blocks. The bullet list is not the only difference between the two conditional branches so not sure what could be gained :thinking:

udinobi: You're right :+1:  I did not notice the other differences between the 2 layouts.

test/views/UnauthorisedViewSpec.scala
glidester: Not sure I understand, it is a method as it takes a dynamic parameter, how could it become an immutable class field?

app/views/unauthorised.scala.html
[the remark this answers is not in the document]
udinobi: I second that, Darryl. Unfortunately our views have been built with a different philosophy.

test/views/UnauthorisedViewSpec.scala
udinobi: because the actual view's content is only built when the view's `apply` method is called and this takes place at the next line (96).

udinobi (approved):

app/views/unauthorised.scala.html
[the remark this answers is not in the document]
glidester: I agree that too much logic in views is to be avoided (especially as the twirl syntax is so odd) but a simple `if` statement won't tax the readers brains too much I feel ;)

PR #1431, opened 2021-12-13
CEDS-3577 Add error message for Y999 doc code

magicaljellybeans:
Adding of temporary error message stopping use of code "Y999"

Had to enable escaping of HTML in a couple of places to allow for this multi-line error message.

### DIFF
--- a/app/views/components/gds/errorSummary.scala.html
+++ b/app/views/components/gds/errorSummary.scala.html
@@ -23,8 +23,8 @@
 @errorList = @{errors.map { error =>
     val errorKey = error.key.replace(".", "_").replace('[', '_').replace("]", "")
     ErrorLink(
-        href = Some(s"#${errorFieldName.getOrElse(errorKey)}"),
-        content = Text(messages(error.message, error.args: _*))
+      href = Some(s"#${errorFieldName.getOrElse(errorKey)}"),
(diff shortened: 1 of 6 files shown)


app/views/components/gds/errorSummary.scala.html
```diff
@@ -22,6 +22,12 @@
 
 @errorList = @{errors.map { error =>
     val errorKey = error.key.replace(".", "_").replace('[', '_').replace("]", "")
+    if (error.message.contains("Y999"))
```
udinobi, line 25: What about...
```
    ErrorLink(
        href = Some(s"#${errorFieldName.getOrElse(errorKey)}"),
        content =
            if (error.message.contains("Y999")) HtmlContent(messages(error.message, error.args: _*))
            else Text(messages(error.message, error.args: _*))
    )
```

udinobi (dismissed):

glidester (changes requested):
Can we think about generalising this 'errors as HTML' feature so it is not 'Y999' specific (probably need to use this for other errors in future). :slightly_smiling_face:

magicaljellybeans:
Thanks for the pointers, changes made.

glidester (approved):
:+1:

udinobi (approved):

PR #1433, opened 2021-12-14
CEDS-3327 Implementation of the new page at '/inland-or-border'

udinobi:


app/controllers/actions/JourneyAction.scala
```diff
@@ -31,34 +32,56 @@ import uk.gov.hmrc.play.http.HeaderCarrierConverter
...
+    logger.warn(message)
+    Left(Results.Redirect(RootController.displayPage))
+  }
+
+  private def verifyDeclaration[A](id: String, request: AuthenticatedRequest[A], onCondition: ExportsDeclaration => Boolean): RefineResult[A] = {
+    implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequestAndSession(request, request.session)
+
+    cacheService.get(id).map {
+      case Some(declaration) =>
+        if (onCondition(declaration)) Right(new JourneyRequest(request, declaration)) else Left(Results.Redirect(RootController.displayPage))
```
glidester, line 82: If `onCondition` is not met (so a Left is returned) the WARN log message `"Could not retrieve from cache, for eori ${request.user.eori}, the declaration with id $id"` is not created (as it used to be in this scenario) :thinking:

app/controllers/helpers/TransportSectionHelper.scala
```diff
@@ -0,0 +1,31 @@
...
+
+package controllers.helpers
+
+import forms.declaration.ModeOfTransportCode
+import forms.declaration.ModeOfTransportCode.{FixedTransportInstallations, PostalConsignment}
+import forms.declaration.additionaldeclarationtype.AdditionalDeclarationType._
+
+object TransportSectionHelper {
+
+  val altAdditionalTypesOnTransportSection = List(STANDARD_PRE_LODGED, STANDARD_FRONTIER, SUPPLEMENTARY_SIMPLIFIED)
```
glidester, line 25: Not sure what the 'alt' bit stands for in this var name (alternative?). Are these not the 'additionalDecTypesAllowedOnTransportSection'? :thinking:

udinobi: I admit I was not creative enough, but it's tricky to find the right one :grin: 
Even `additionalDecTypesAllowedOnTransportSection` is not correct because it gives you the idea that these are the only ADTs allowed in the transport section, which is not true. It's just that under them there are specific conditions to take into account to correctly address the journey flow.
And yes, `alt` stands for alternative, which if you ask, I can use in full :slightly_smiling_face:

app/controllers/actions/JourneyAction.scala
udinobi: In truth, the previous log was partially wrong because it was covering 2 different error situations with the same message:
1. If the cache cannot retrieve the declaration by using the id, the error message was correct.
2. If the cache actually retrieve the declaration by using the id but the declaration's type is not the expected one, the message should be different imo.

I'll add a message for the 2nd one btw :+1:

app/controllers/helpers/TransportSectionHelper.scala
glidester: What do you think of maybe `additionalDecTypesAllowedOnInlandOrBorder` (as these are the only types allowed on this page by the `JourneyAction` refiner)? I understand they are also used elsewhere for other decision points but this seems like the strongest candidate IMO :slightly_smiling_face:  

If not then maybe some sort of comment explaining their significance/purpose?

glidester (approved):

darrylmurphy (approved):

darrylmurphy (approved):

PR #1434, opened 2021-12-14
[CEDS-3568] Conditional 3 versions on is-additional-information-required

glidester:


app/controllers/declaration/AdditionalInformationRequiredController.scala
```diff
@@ -69,9 +69,12 @@ class AdditionalInformationRequiredController @Inject()(
       )
   }
 
-  private def cachedItems(itemId: String)(implicit request: JourneyRequest[AnyContent]): Seq[AdditionalInformation] =
+  private def cachedAdditionalInformation(itemId: String)(implicit request: JourneyRequest[AnyContent]): Seq[AdditionalInformation] =
```
udinobi, line 72: This `def` could be maybe moved to `ExportsDeclaration` to be of general use. WDYT?

app/controllers/declaration/AdditionalInformationRequiredController.scala
```diff
@@ -69,9 +69,12 @@ class AdditionalInformationRequiredController @Inject()(
...
     request.cacheModel.itemBy(itemId).flatMap(_.additionalInformation).getOrElse(AdditionalInformationData.default).items
 
+  private def cachedProcedureCode(itemId: String)(implicit request: JourneyRequest[AnyContent]): Option[ProcedureCodesData] =
```
udinobi, line 75: This `def` could be maybe moved to `ExportsDeclaration` to be of general use... if not already there. WDYT?

app/views/helpers/AdditionalInformationRequiredHelper.scala
```diff
@@ -0,0 +1,100 @@
...
+import uk.gov.hmrc.govukfrontend.views.Aliases.HtmlContent
+import views.html.components.gds.{exportsInsetText, link, paragraphBody}
+
+import javax.inject.{Inject, Singleton}
+
+@Singleton
+class AdditionalInformationRequiredHelper @Inject()(appConfig: AppConfig, paragraphBody: paragraphBody, insetText: exportsInsetText, link: link) {
+  def getBodyContent(decType: DeclarationType, mayBeProcedureCode: Option[ProcedureCodesData])(implicit messages: Messages): Html =
+    (decType, mayBeProcedureCode.flatMap(_.procedureCode)) match {
+      case (CLEARANCE, _)                => getBodyContentForClearanceOr1040()
```
udinobi, line 34: I did not try myself but just to be safe, considering the order of the cases, are you sure that if you have, for instance, Clearance and 1042 this case will not be considered?

glidester: It will be caught by this case yes (that is the desired behavior, if on clearance journey then the PC does not matter) :slightly_smiling_face:

udinobi: I know you like pattern matching :smile: but tbh It's quite tricky to get in this form. At the end it's just a simple `if`, isn't it?
```
val is1040 = mayBeProcedureCode.exists(_.procedureCode == 1040)
if (is1040 || decType == CLEARANCE) getBodyContentForClearanceOr1040()
else getBodyContentForNot1040(pc)
```

udinobi: But it's subjective, of course. I have no problem with your syntax.

glidester: I get where you're coming from, but the pattern matching is also handling the optionality of `mayBeProcedureCode.flatMap(_.procedureCode)` (so covering off a default action if something odd happens like user skipping the PC pages). 
But yeah, its all subjective ;)

udinobi (approved):

PR #1435, opened 2021-12-15
CEDS-3537 Change order of Summary page rows

magicaljellybeans:
Also removed `Warehouse Information` heading.


glidester (changes requested):
Looks good. Should delete the `warehouse_section.scala.html` file if it is now no-longer used though :slightly_smiling_face:

glidester (dismissed):
:+1:

glidester (approved):

PR #1437, opened 2021-12-17
CEDS-3463 Add LRN validation error on summary page

magicaljellybeans:


app/controllers/declaration/SummaryController.scala
```diff
@@ -64,23 +69,41 @@ class SummaryController @Inject()(
   }
 
   val submitDeclaration: Action[AnyContent] = (authenticate andThen verifyEmail andThen journeyType).async { implicit request =>
-    LegalDeclaration.form.bindFromRequest
-      .fold(
+    def submit(form: Form[LegalDeclaration]): Future[Result] =
+      form.bindFromRequest.fold(
```
glidester, line 73: This might be by design but as the form binding is happening here, if the LRN is a duplicate then no form errors from the legal declaration form will be detected or shown to the user. Generally we try to display all the errors on the page at the same time but maybe this error is special? (Maybe check with Liz and/or Johnathan)?

test/controllers/declaration/SummaryControllerSpec.scala
```diff
@@ -133,6 +136,13 @@ class SummaryControllerSpec extends ControllerWithoutFormSpec with ErrorHandlerM
 
         status(result) must be(BAD_REQUEST)
       }
+
```
glidester, line 139: Need to add tests that exercise the code that handles the lrn is a duplicate condition.

glidester (changes requested):
Approach looks good. Just a few minor issues to resolve :slightly_smiling_face:

app/controllers/declaration/SummaryController.scala
magicaljellybeans: That was an oversight on my part, I have now added the LRN duplication error alongside so it will appear with any others.

app/controllers/declaration/SummaryController.scala
```diff
@@ -64,23 +69,41 @@ class SummaryController @Inject()(
...
+    lrn match {
+      case Some(lrn) => lrnValidator.hasBeenSubmittedInThePast48Hours(lrn)
+      case _         => Future.successful(false)
+    }
+
+  private def getLrn(declaration: ExportsDeclaration): Option[Lrn] =
+    declaration.consignmentReferences.map(_.lrn) match {
+      case Some(lrn) => Some(lrn)
+      case _         => None
+    }
```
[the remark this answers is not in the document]
magicaljellybeans: I removed this method but still had to have a matching statement on `ExportsDeclaration.lrn` because it's an `Option[String]` - hopefully this looks better?

test/controllers/declaration/SummaryControllerSpec.scala
magicaljellybeans: I have tried to add another test that stubs `LrnValidator.hasBeenSubmittedInThePast48Hours` to emulate a duplicated lrn being found.

However the test seems to ignore the stub and I haven't been able to figure out why - any ideas?

app/controllers/declaration/SummaryController.scala
```diff
@@ -64,23 +69,39 @@ class SummaryController @Inject()(
...
-
             case _ => handleError(s"Error from Customs Declarations API")
           }
         }
       )
+
+    isLrnADuplicate(request.cacheModel.lrn match {
+      case Some(lrn) => Some(Lrn(lrn))
+      case _         => None
+    }).flatMap {
```
glidester, line 87: To aid code readability try to avoid performing logic in calls to methods, instead you can do the logic just above the method and assign it to a var and then pass that var to the method, e.g,
```
val maybeLrn = request.cacheModel.lrn.map(Lrn(_))

isLrnADuplicate(maybeLrn).flatMap { ...

``` 
:wink:

app/controllers/declaration/SummaryController.scala
```diff
@@ -64,23 +69,38 @@ class SummaryController @Inject()(
...
+      }
+      case _ => submit(form)
+    }
   }
 
   private def containsMandatoryData(declaration: ExportsDeclaration, mode: Mode): Boolean =
     mode.equals(Mode.Draft) || declaration.consignmentReferences.exists(references => references.lrn.nonEmpty)
 
+  private def isLrnADuplicate(lrn: Option[Lrn])(implicit hc: HeaderCarrier): Future[Boolean] =
+    lrn match {
```
udinobi, line 99: If you want to have a version a bit less verbose...
```
lrn.fold(Future.successful(false))(lrnValidator.hasBeenSubmittedInThePast48Hours)
```
or even...
```
lrn.map(lrnValidator.hasBeenSubmittedInThePast48Hours).getOrElse(Future.successful(false))
```

udinobi (approved):

udinobi (approved):

glidester (approved):

PR #1448, opened 2022-01-05
CEDS-3346 New logic to decide whether to land on or skip the /inland-or-border page

udinobi:


test/controllers/declaration/ProcedureCodeControllerSpec.scala
```diff
@@ -313,13 +314,11 @@ class ProcedureCodeControllerSpec extends ControllerSpec with ErrorHandlerMocks
       }
     }
 
-    val warehouseRequiredProcedureCodesSuffixes = Seq("07", "71", "78")
-    warehouseRequiredProcedureCodesSuffixes.foreach { suffix =>
-      s"user provides procedure code indicating WarehouseIdentifier is required (ends with $suffix" when {
+    warehouseRequiredProcedureCodes.foreach { suffix =>
```
udinobi, line 317: This test is suspicious. Is `suffix` used somewhere in the test code? The test actually does not seem to match the documentation.

app/models/ExportsDeclaration.scala
```diff
@@ -135,6 +135,8 @@ case class ExportsDeclaration(
   def removeSupervisingCustomsOffice: ExportsDeclaration =
     copy(locations = locations.copy(supervisingCustomsOffice = None))
 
+  def skipInlandOrBorder: Boolean = parties.declarationHoldersData.exists(_.holders.exists(_.skipInlandOrBorder))
```
glidester, line 138: Not sure this belongs here in the model, it is about a concern of the UI flow logic. Better off in a UI 'helper' class.

app/services/DepCodes.scala
```diff
@@ -0,0 +1,41 @@
...
+import java.nio.file.{Files, Paths}
+import scala.collection.JavaConverters.asScalaIteratorConverter
+
+object DepCodes {
+
+  def isDesignatedExportPlaceCode(declaration: ExportsDeclaration): Boolean =
+    declaration.locations.goodsLocation.exists(location => designatedExportPlaceCodes.contains(location.code))
+
+  private val designatedExportPlaceCodes: Set[String] = {
+    val filename = "Designated_Export_Place_codes_for_Data_Element-5-23_of_the_CDS__v2__1_.csv"
```
glidester, line 31: Can this data not be loaded as a JSON file like procedure codes and countries? We are trying move all the existing CSV data into JSON to support the future reference values service.

test/controllers/declaration/ProcedureCodeControllerSpec.scala
glidester: No, I can't see how these suffixes alter the test conditions either. I'd assume they were supposed to be concatenated with the procedureCode value on line 329, maybe give that a try and see if the tests still pass :thinking:

app/services/DepCodes.scala
```diff
@@ -0,0 +1,41 @@
...
+
+package services
+
+import models.ExportsDeclaration
+
+import java.nio.charset.StandardCharsets.ISO_8859_1
+import java.nio.file.{Files, Paths}
+import scala.collection.JavaConverters.asScalaIteratorConverter
+
+object DepCodes {
```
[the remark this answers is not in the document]
udinobi: They are not Departure codes. They are DEP (called this way even on gov.uk as well as in the ticket) location codes.
DEP is for Designated Export Place which is already in the id of `val` and `def`. Named the class that way too would be definitely redundant.

app/services/DepCodes.scala
glidester: Agreed to replace this CSV file loading with the standard json loading service pattern after the Navigator object has been converted into a singleton. :+1:

glidester (approved):

PR #1452, opened 2022-01-18
CEDS-3216 New versioned content for page at /departure-transport

udinobi:


app/forms/declaration/DepartureTransport.scala
```diff
@@ -33,60 +32,24 @@ object DepartureTransport extends DeclarationPage {
...
+    mapping(
+      radioButtonGroupId -> mappingForRadioDepartureType(transportCodes),
+      mappingForInputDepartureNumber(transportCodes(0)),
+      mappingForInputDepartureNumber(transportCodes(1)),
+      mappingForInputDepartureNumber(transportCodes(2)),
+      mappingForInputDepartureNumber(transportCodes(3)),
+      mappingForInputDepartureNumber(transportCodes(4)),
+      mappingForInputDepartureNumber(transportCodes(5)),
+      mappingForInputDepartureNumber(transportCodes(6)),
+      mappingForInputDepartureNumber(transportCodes(7))
```
glidester, line 51: These 'magic numbers' risk throwing a `java.lang.IndexOutOfBoundsException` if anyone ever modifies the `transportCodesForV1`etc TransportCodes Lists. Is there a way to avoid this by mapping the Seq values that are sent? :thinking:

glidester: How does this work with `transportCodesForV3WhenPC0019` that has 9 elements?

app/views/helpers/DepartureTransportHelper.scala
```diff
@@ -0,0 +1,143 @@
...
+    versionSelection match {
+      case 1                                          => transportCodesForV1
+      case 2                                          => transportCodesForV2
+      case 3 if hasPCsEqualTo0019(request.cacheModel) => transportCodesForV3WhenPC0019
+      case 3                                          => transportCodesForV3
+    }
+
+  /*
+   The version of the page's content is chosen under the assumption that the
+   user has selected on /transport-leaving-leaving NEITHER 'Postal' or 'FTI'.
```
glidester, line 134: Typo "/transport-leaving-leaving"

test/views/declaration/DepartureTransportViewSpec.scala
```diff
@@ -98,147 +86,195 @@ class DepartureTransportViewSpec extends UnitViewSpec with CommonMessages with S
...
-            "declaration.transportInformation.meansOfTransport.nameOfInlandWaterwayVessel"
-          )
-          view
-            .getElementsByAttributeValue("for", s"meansOfTransportOnDepartureIDNumber_$NameOfInlandWaterwayVessel") must containMessageForElements(
-            "declaration.transportInformation.meansOfTransport.nameOfInlandWaterwayVessel.label"
-          )
+          val radioLabels = view.getElementsByClass("govuk-radios__label")
+          radioLabels.size mustBe transportCodes.size
+          radioLabels.iterator.asScala.zipWithIndex.foreach { elementAndIndex =>
+            val (element, index) = (elementAndIndex._1, elementAndIndex._2)
```
glidester, line 157: Nit: could just do `val (element, index) = elementAndIndex`

app/forms/declaration/DepartureTransport.scala
udinobi: Sadly form mapping in Play works by passing a specific number of parameters, and not, for instance, passing a vararg parameter. So, no :frowning_face: there is now way to avoid this, or at least it was me not able to find a DRY alternative. The only other possible alternative is to have a mapping for every version of the page (and we have 4 of them).

glidester: I was wondering if you could use a Tuple8 instead for the 'transportCodesForVX' vals, then if someone changed the number of elements we would get a compiler error. What do you think?

udinobi: This method is for mapping the input fields associated with the radio buttons (8 input fields). In `transportCodesForV3WhenPC0019` the 9th radio has no input field associated.

test/views/declaration/DepartureTransportViewSpec.scala
udinobi: :upside_down_face:

app/forms/declaration/DepartureTransport.scala
udinobi: I also thought to see if a tuple could be used but this is not feasible due to the fact that the elements the form receives as parameters could be 8 or 9, and we cannot express that with a tuple.

glidester (approved):

glidester (approved):

PR #1454, opened 2022-01-20
CEDS-3346 Fix navigation when RoRo is selected on /transport-leaving-the-border

udinobi:


magicaljellybeans (commented):
LGTM, but not fully confident approving without a second opinion

glidester (approved):

PR #1458, opened 2022-01-27
CEDS-3730 Amend content of page at /location-of-goods

udinobi:
Better review by commit. The 1st 2 commits are just renaming / refactoring of code not really relevant for the ticket at hand. Can explain the reason behind these 2 commits if required.

The 3rd commit is the relevant one.


glidester (approved):
Thank you for splitting out the refactoring work into separate commits :hugs:

PR #1468, opened 2022-02-11
CEDS-3684 audit event on file upload handoff

magicaljellybeans:


app/controllers/FileUploadController.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import javax.inject.Inject
+
+class FileUploadController @Inject()(sfusConfig: SfusConfig, authenticate: AuthAction, mcc: MessagesControllerComponents, auditService: AuditService)
+    extends FrontendController(mcc) {
+
+  def startFileUpload(mrn: String = ""): Action[AnyContent] = authenticate { implicit request =>
+    lazy val sfusLink = s"${sfusConfig.sfusUploadLink}/$mrn"
+
+    def auditData(eori: String, lrn: Option[String], mrn: String, docUploadUrl: String) =
```
udinobi, line 34: Any reason for `auditData` to not be placed in `AuditService.audit`? IMO `EventData` should be created by the service and not at Controller-level.

magicaljellybeans: Do you mean putting `auditData` in the actual `AuditService` file? Or just in the call?

If the former, don't I need to label my data before it gets sent to the service - because otherwise the service won't know what's what?

udinobi: Ah, OK. I see. `auditService.audit` is the generic method for auditing, used by other controllers as well. Sorry, ignore my comment.

udinobi (approved):

PR #1483, opened 2022-03-02
CEDS-3498 Remove ExportsDeclarationExchange class

udinobi:
- Remove `ExportsDeclarationExchange` class
- Refactor `TotalNumberOfItems` section, which now also includes `totalPackageQuantity` (meaning that now it is not any more a separated field of `ExportsDeclaration`


glidester (approved):
Looks really good Lucio, nice simplification of the service :+1:

PR #1495, opened 2022-03-10
Ceds 3389 reminder for exporters to special fiscal territory

magicaljellybeans:


app/controllers/declaration/AdditionalProcedureCodesController.scala
```diff
@@ -115,6 +125,14 @@ class AdditionalProcedureCodesController @Inject()(
       case None => (None, emptyProcedureCodesData)
     }
 
+  private def getHintText()(implicit request: JourneyRequest[AnyContent]): Option[String] =
```
udinobi, line 128: Any particular reason to not place this in a small helper under `views/helpers`? IMO we should try to have any logic concerning the pages/templates in that folder and not in controllers. But open to discuss the idea.

app/models/ExportsDeclaration.scala
```diff
@@ -117,7 +117,8 @@ case class ExportsDeclaration(
   def isNotExs: Boolean = !isExs
 
   def isEntryIntoDeclarantsRecords: Boolean = parties.isEntryIntoDeclarantsRecords.exists(_.answer == YesNoAnswers.yes)
-  def isNotEntryIntoDeclarantsRecords: Boolean = !isEntryIntoDeclarantsRecords
+  def isNotEntryIntoDeclarantsRecords: Boolean = parties.isEntryIntoDeclarantsRecords.exists(_.answer == YesNoAnswers.no)
+  def isNotEntryIntoDeclarantsRecordsOrNoAnswer: Boolean = isNotEntryIntoDeclarantsRecords || parties.isEntryIntoDeclarantsRecords.isEmpty
```
udinobi, line 121: Do we really need these 2 `def`? Isn't `isEntryIntoDeclarantsRecords`, the method at 119, enough for our needs?

I mean... do we actually need to differentiate between
```parties.isEntryIntoDeclarantsRecords.isEmpty``` and
```parties.isEntryIntoDeclarantsRecords.exists(_.answer == YesNoAnswers.no)``` ?

app/controllers/declaration/AdditionalProcedureCodesController.scala
magicaljellybeans: No particular reason, just an oversight on my part, thanks for pointing this out.

udinobi (approved):

PR #1503, opened 2022-03-16
Ceds 3694 content is add docs req

darrylmurphy:


app/views/declaration/summary/additional_documents.scala.html
```diff
@@ -72,6 +72,7 @@
 
 @tableCaption = @{
     if(licenceRequired && showWaiverInfo) messages("declaration.summary.items.item.licences")
+    else if(declarationType == models.DeclarationType.CLEARANCE || !showWaiverInfo) messages("declaration.summary.items.item.clearance.additionalDocuments")
```
glidester, line 75: This condition is a little confusing. If declarationType is CLEARANCE or Not(declarationType is not CLEARANCE or is999LEnabled. 
Could this be simplified to `if(declarationType == models.DeclarationType.CLEARANCE || !waiver999LConfig.is999LEnabled)` ?

glidester (approved):

PR #1518, opened 2022-04-01
CEDS-3783 new /invoices-and-exchange-rate-choice page

udinobi:
Note that only the 2nd commit is ticket-relevant. (sorry, this before the rebase)

### DIFF
--- a/app/models/ExportsDeclaration.scala
+++ b/app/models/ExportsDeclaration.scala
@@ -104,6 +104,11 @@ case class ExportsDeclaration(
 
   def isAdditionalDeclarationType(adt: AdditionalDeclarationType): Boolean = additionalDeclarationType.exists(_ == adt)
 
+  def isInvoiceAmountGreaterThan100000: Boolean =
+    totalNumberOfItems.exists(_.totalAmountInvoiced.fold(false)(_.trim.nonEmpty))
+
+  def isInvoiceAmountLessThan100000: Boolean = !isInvoiceAmountGreaterThan100000
+
   def hasAuthCodeRequiringAdditionalDocs: Boolean =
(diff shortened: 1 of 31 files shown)


app/models/ExportsDeclaration.scala
```diff
@@ -104,6 +104,11 @@ case class ExportsDeclaration(
 
   def isAdditionalDeclarationType(adt: AdditionalDeclarationType): Boolean = additionalDeclarationType.exists(_ == adt)
 
+  def isInvoiceAmountGreaterThan100000: Boolean =
+    totalNumberOfItems.exists(_.totalAmountInvoiced.fold(false)(_.trim.nonEmpty))
+
+  def isInvoiceAmountLessThan100000: Boolean = !isInvoiceAmountGreaterThan100000
+
```
glidester, line 111: Method names doesn't seem to describe what method does. Its not checking value is > 1000, just if it exists and is non-empty :thinking:

udinobi: Indeed. But that is the logic behind. When non-empty, the value in `totalAmountInvoiced` has always to be >= 100,000, if empty, or not defined, it is because the amount was < 100,000. So I would not change the methods names.

magicaljellybeans: There is the potential for a user to enter a total amount invoiced less than 100000. 

They'd be contradicting the answer they've given on the new page, but they could still do it.

glidester: I understand the business rule of < 100000 is what defines if the user needs to enter the value or not but personally as a dev reading this code (who might not know the business rule) it is a bit confusing. Not a deal breaker though I suppose :slightly_smiling_face:

glidester (approved):

magicaljellybeans (approved):

PR #1539, opened 2022-04-22
CEDS-3846: adjust content on dec type page

darrylmurphy:


app/views/declaration/additional_declaration_type.scala.html
```diff
@@ -31,6 +31,7 @@
     govukDetails: GovukDetails,
     govukRadios: GovukRadios,
     errorSummary: errorSummary,
+    yesNoRadios: yesNoRadios,
```
magicaljellybeans, line 34: This is imported but not used.

magicaljellybeans (approved):

PR #1559, opened 2022-05-05
CEDS-3863: split unauthorised eori into view, update unauthorised view

darrylmurphy:


udinobi (approved):

app/views/unauthorisedEori.scala.html
```diff
@@ -0,0 +1,44 @@
...
+@this(
+    govukLayout: gdsMainTemplate,
+    pageTitle: pageTitle,
+    paragraphBody: paragraphBody,
+    link: link,
+    bulletList: bulletList,
+    config: ExternalServicesConfig
+)
+
+@(displaySignOut: Boolean)(implicit request: Request[_], messages: Messages)
```
magicaljellybeans, line 30: Can remove `displaySignOut` parameter for this. Unauthorised EORI means the user will always be logged in so we can always show them the sign out link on this page.

app/views/unauthorisedEori.scala.html
```diff
@@ -0,0 +1,44 @@
...
+
+@govukLayout(title = Title("unauthorised.tdr.heading"), displaySignOut = displaySignOut) {
```
magicaljellybeans, line 32: So this can be `displaySignOut = true`

magicaljellybeans (changes requested):

magicaljellybeans (approved):

PR #1573, opened 2022-05-17
CEDS-3873 Add dynamic text on additional-procedure-codes

glidester:


magicaljellybeans (approved):

test/views/declaration/procedureCodes/AdditionalProcedureCodesViewSpec.scala
```diff
@@ -52,7 +52,8 @@ class AdditionalProcedureCodesViewSpec extends UnitViewSpec with ExportsTestData
 
     "have proper messages for labels" in {
       messages must haveTranslationFor("declaration.additionalProcedureCodes.title")
-      messages must haveTranslationFor("declaration.additionalProcedureCodes.paragraph")
+      messages must haveTranslationFor("declaration.additionalProcedureCodes.paragraph1")
+      messages must haveTranslationFor("declaration.additionalProcedureCodes.paragraph2")
```
udinobi, line 56: tbh, rather than this tests, I would like to see proper tests for the 2 paragraphs.

glidester: I really don't see the value in checking that a paragraph of text is there. The current test requires that the i18n key is present which might conceivably get altered or removed by accident, the others test that the common interaction components like buttons and links are present. But simple content like a line of text being present has no real value IMO

PR #1578, opened 2022-05-23
CEDS-3855 Add agent kick out page to service

magicaljellybeans:


app/controllers/UnauthorisedController.scala
```diff
@@ -20,24 +20,34 @@ import javax.inject.Inject
...
-    if (tdrFlagEnabled || unauthorizedDueToEoriNotAllowed) {
-      Ok(unauthorisedEoriPage())
-    } else {
-      Ok(unauthorisedPage(displaySignOut))
+    unauthorisedReason match {
+      case _ if tdrFlagEnabled => Ok(unauthorisedEoriPage())
+      case UserEoriNotAllowed  => Ok(unauthorisedEoriPage())
+      case UserIsNotEnrolled   => Ok(unauthorisedPage(displaySignOut = true))
+      case UrlDirect           => Ok(unauthorisedPage(displaySignOut = false))
     }
```
udinobi, line 44: What if `unauthorisedReason` is `UserIsAgent` and `tdrFlagEnabled` is `false`? Of course assuming that a call site
`onPageLoad(UserIsAgent)` is possible.

app/models/UnauthorisedReason.scala
```diff
@@ -0,0 +1,63 @@
...
+    override def bind(key: String, params: Map[String, Seq[String]]): Option[Either[String, UnauthorisedReason]] =
+      Some(
+        Right(
+          params
+            .get(key)
+            .flatMap(_.headOption)
+            .flatMap(reason => allUnauthorisedReasons.find(_.name == reason))
+            .getOrElse(UrlDirect)
+        )
+      )
```
udinobi, line 59: Do you have a test for verifying that the `Some(Left(...))` result is actually generated?

app/controllers/UnauthorisedController.scala
magicaljellybeans: As `UserIsAgent` is only passed in the `AuthActionImpl` on the /you-cannot-use-this-service url that can't happen currently.

udinobi: What I actually mean is... is it possible for the `onPageLoad` to receive an argument which is not
  `UserEoriNotAllowed` | `UserIsNotEnrolled` | `UrlDirect`
and having `tdrFlagEnabled` as false?

app/models/UnauthorisedReason.scala
udinobi: Unless you are only doing this due to the `bind` signature that you are overriding

udinobi (approved):

PR #1594, opened 2022-06-27
CEDS-3920 Add cancellation holding and result pages

magicaljellybeans:


app/models/declaration/submissions/Action.scala
```diff
@@ -25,11 +25,19 @@ case class Action(
   id: String,
   requestType: RequestType,
   requestTimestamp: ZonedDateTime = ZonedDateTime.now(defaultDateTimeZone),
-  notifications: Option[Seq[NotificationSummary]] = None
-)
+  notifications: Option[Seq[NotificationSummary]]
+) {
+  val latestNotificationSummary: Option[NotificationSummary] = if (notifications.isDefined && notifications.get.nonEmpty) {
```
glidester, line 30: To get the latest notification you can just take the last item in the Seq as they are already ordered by date in the collection :+1:

PR #1606, opened 2022-07-04
CEDS-3897 Refactor submit dec receipt page

glidester:


app/controllers/declaration/ConfirmationController.scala
```diff
@@ -76,16 +77,16 @@ class ConfirmationController @Inject() (
       logger.warn("Session on /confirmation does not include the submission's uuid!?")
       Future.successful(Redirect(SubmissionsController.displayListOfSubmissions()))
     } { submissionId =>
-      customsDeclareExportsConnector.findLatestNotification(submissionId).flatMap {
-        case Some(notification) if notification.isStatusDMSRej =>
+      customsDeclareExportsConnector.findSubmission(submissionId).flatMap {
+        case Some(submission) if submission.latestEnhancedStatus == Some(EnhancedStatus.ERRORS) =>
```
udinobi, line 81: Just to be sure... should `EXPIRED_NO_ARRIVAL` be also included in this check?

glidester: No I don't think so. EXPIRED_NO_ARRIVAL will only happen 30 days after you submit your declaration if no arrival message is received in that time. So it will never be sent in the first few seconds after a dec submission (where as ERRORS certainly will be sent immediately). :slightly_smiling_face:

PR #1611, opened 2022-07-11
CEDS-3936 Display Cancellation Request Notification status on timeline

udinobi:


app/views/helpers/TimelineEvents.scala
```diff
@@ -41,22 +42,23 @@ class TimelineEvents @Inject() (
...
+    }.sorted
 
-    val IndexToMatchForUploadFilesContent = sortedNotificationsSummaries.indexWhere(summary =>
-      summary.enhancedStatus == ADDITIONAL_DOCUMENTS_REQUIRED || summary.enhancedStatus == UNDERGOING_PHYSICAL_CHECK
-    )
-    val IndexToMatchForViewQueriesContent = sortedNotificationsSummaries.indexWhere(_.enhancedStatus == QUERY_NOTIFICATION_MESSAGE)
-    val IndexToMatchForFixResubmitContent = sortedNotificationsSummaries.indexWhere(_.enhancedStatus == ERRORS)
+    val notificationSummaries =
+      if (!allNotificationSummaries.exists(_.enhancedStatus == CANCELLED)) allNotificationSummaries
+      else allNotificationSummaries.filterNot(_.enhancedStatus == CUSTOMS_POSITION_GRANTED)
```
glidester, line 53: I read this as: if we have received a CANCELLED notification return all the notifications (including the CUSTOMS_POSITION_GRANTED!) else return all notifications apart from CUSTOMS_POSITION_GRANTED.

I think we should always be filtering out CUSTOMS_POSITION_GRANTED as we never want to show this to the user in any circumstance.

glidester (changes requested):

app/views/helpers/TimelineEvents.scala
udinobi: If we have **NOT** received a CANCELLED notification return all the notifications
else, If we have received a CANCELLED notification, return all notifications apart from CUSTOMS_POSITION_GRANTED

glidester: Ah, missed the `!` at the start of the if! Yes but we still should not show the CUSTOMS_POSITION_GRANTED in any circumstances

udinobi: I updated the ticket relatively to not show CUSTOMS_POSITION_GRANTED on the timeline.

app/models/declaration/submissions/EnhancedStatus.scala
```diff
@@ -25,15 +25,21 @@ object EnhancedStatus extends Enumeration {
 
   implicit val format: Format[EnhancedStatus.Value] = EnumJson.format(EnhancedStatus)
 
+  implicit class InclusionInSet(status: EnhancedStatus) {
+    def in(set: Set[EnhancedStatus]): Boolean = set.contains(status)
+  }
```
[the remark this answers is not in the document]
udinobi: Nothing compelling. Still, the implicit method `in` gives you more flexibility when using placeholders in higher-order functions.

glidester (approved):

PR #1619, opened 2022-07-22
CEDS-3989 Add "Save and return to errors" button to editor pages in E…

magicaljellybeans:
…rrorFix mode


app/controllers/declaration/AdditionalProcedureCodesController.scala
```diff
@@ -78,9 +78,10 @@ class AdditionalProcedureCodesController @Inject() (
     def handleActionTypes(procedureCode: ProcedureCode, validAdditionalProcedureCodes: Seq[AdditionalProcedureCodeModel]): Future[Result] = {
       val errorHandler =
         returnErrorPage(mode, itemId, boundForm.get, procedureCode, cachedData.additionalProcedureCodes, validAdditionalProcedureCodes)(_)
+      val actionsRequiringValidation = Seq(Add, SaveAndContinue, SaveAndReturn, SaveAndReturnToSummary, SaveAndReturnToErrors)
 
       formAction match {
-        case Add | SaveAndContinue | SaveAndReturn | SaveAndReturnToSummary if !boundForm.hasErrors =>
+        case _ if !boundForm.hasErrors && actionsRequiringValidation.contains(formAction) =>
```
darrylmurphy, line 84: Would it bet better to move the "Remove" case above the case where it catches all and then does a contain on the same objects

udinobi (dismissed):

darrylmurphy (approved):

PR #1633, opened 2022-08-09
CEDS-4052 New Copy-Declaration page

udinobi:


app/views/declaration_details.scala.html
```diff
@@ -83,13 +83,25 @@ <h2 class="declaration-details-refs govuk-heading-s">@text</h2>
...
-            text = messages("submissions.viewDeclaration"),
+            text = messages("submissions.view.declaration"),
             call = routes.SubmissionsController.viewDeclaration(submission.uuid)
         )
     </div>
   }
 }
 
+@copyDeclarationLink = {
+  @if(!isDeclarationRejected(submission)){
```
magicaljellybeans, line 94: The ticket mentions also not displaying the link if "in Draft mode", which I take to mean checking the `readyForSubmission` flag to be true.

magicaljellybeans: Or checking `status` in cache maybe?

magicaljellybeans: Yeah I'm just thinking neither of those are required because the declaration has already been submitted :laughing:

magicaljellybeans (approved):

app/views/declaration_details.scala.html
udinobi: Indeed :+1:

PR #1668, opened 2022-10-10
Ceds 4101 dashboard adjustment

darrylmurphy:


app/models/SubmissionsPages.scala
```diff
@@ -19,30 +19,43 @@ package models
 import play.api.mvc.QueryStringBindable
 
 case class SubmissionsPages(
-  rejectedPageNumber: Int = SubmissionsPages.DefaultPageNumber,
+  otherPageNumber: Int = SubmissionsPages.DefaultPageNumber,
```
magicaljellybeans, line 22: Nit: Can `other` be `submitted`. It would be less ambiguous when reading.

magicaljellybeans (approved):

app/models/SubmissionsPages.scala
darrylmurphy: I thought of this too. As I picked it up, _other_ was already in use, and as its calculated in *EnhancedStatus* to be a set comprised of all statuses _other than_ those _rejected_, _cancelled_ or _action needed_ I thought it best to leave it as is?

magicaljellybeans (approved):

app/models/viewmodels/SubmissionTabItem.scala
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package models.viewmodels
+
+import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.HtmlContent
+import uk.gov.hmrc.govukfrontend.views.viewmodels.tabs.{TabItem, TabPanel}
+
+case class SubmissionTabItem(tab: String) {
```
udinobi, line 23: It seems a bit over-engineered a case class here. Wouldn't be enough just to have in the Dashboard template something like ...
```
@tabItem(tab: String, content: HtmlContent) = @{
  TabItem(id = Some(s"$tab-submissions"), label = messages(s"submissions.$tab.tab.title"), panel = TabPanel(content = content))
}
```

udinobi (approved):

PR #1675, opened 2022-10-13
CEDS-4109 Show on pages the 'Save and return to summary' button only when the new model's 'summaryWasVisited' flag is true

udinobi:


app/views/declaration/summary/normal_summary_page.scala.html
```diff
@@ -44,7 +44,11 @@
     paragraphBody: paragraphBody
 )
 
-@(backLink: Call, errors: Seq[FormError] = Seq.empty[FormError], maybeContinuePlaceholder: Option[String] = None)(implicit request: JourneyRequest[_], messages: Messages, appConfig: AppConfig)
+@(mode: Mode,
```
magicaljellybeans, line 47: Why is the mode is reintroduced for summary page?

magicaljellybeans (approved):

app/views/declaration/summary/normal_summary_page.scala.html
udinobi: Because all sections in this page, like `@references_section`, `@parties_section`, ..., still need to be called with `mode` as first parameter, as they are still passing it to every internal route they contain.

My guess then was that it was safer to change them once we completely wipe out all modes (with another ticket).

magicaljellybeans (dismissed):

app/controllers/declaration/SummaryController.scala
```diff
@@ -82,36 +86,32 @@ class SummaryController @Inject() (
...
-      if (lrnIsDuplicate) Ok(normalSummaryPage(backlink, duplicateLrnError))
-      else Ok(amendDraftSummaryPage(backlink))
+      if (lrnIsDuplicate) Ok(normalSummaryPage(mode, backlink, duplicateLrnError))
+      else Ok(amendDraftSummaryPage(mode, backlink))
     }
   }
 
-  private val hrefSourcePattern = """href="/customs-declare-exports/declaration/.+\?mode=Change"""".r
-  private val hrefDestPattern = s"""href="$continuePlaceholder""""
+  private val hrefSource = """href="/customs-declare-exports/declaration/.+\?mode="""
```
udinobi, line 102: This is a temporary solution. When we remove all modes, we'll need to find a different solution to differentiate, in the "Summary" page, rows with a `Change` link from any other link included in the page pointing to  `/customs-declare-exports/declaration/{...}` but that is not part of the Summary table.

magicaljellybeans (approved):

PR #1684, opened 2022-11-14
CEDS-3489: rid service of pdf, adjust tests and stylesheets

darrylmurphy:


magicaljellybeans (changes requested):
Looks good. Can remove playfop dependency.

magicaljellybeans (approved):

PR #1687, opened 2022-11-29
Ceds 4262 create trader reference and confirm ducr pages

magicaljellybeans:
Things to do before merging:

- [x] Set appropriate page for backlink from /trader-reference (may need to be done in Lucio's ticket)
- [x] Set appropriate page for forward navigation from /confirm-ducr on "Yes" answer to /local-reference-number (when this work is merged by Darryl)


app/controllers/declaration/ConfirmDucrController.scala
```diff
@@ -0,0 +1,76 @@
...
+          case _                             => updateCache.map(_ => navigator.continueTo(routes.DucrEntryController.displayPage))
+        }
+      )
+    }
+  }
+
+  private def form: Form[YesNoAnswer] = YesNoAnswer.form(errorKey = "declaration.confirmDucr.error.empty")
+
+  private def updateCache(implicit request: JourneyRequest[_]): Future[ExportsDeclaration] =
+    updateDeclarationFromRequest(_.copy(consignmentReferences = None))
```
udinobi, line 75: Just a confirmation here... is it OK always set `consignmentReferences` to `None` when DUCR is not confirmed, because this will also remove the LRN if already entered.

magicaljellybeans: No it is potentially not, thanks for spotting. I think I'll need to change the backend too now to make Ducr optional.

PR #1704, opened 2022-12-23
CEDS-4032 Upgrade to scala 2.13 and Java 11

magicaljellybeans:
Removal of silencer plugin


project/AppDependencies.scala
```diff
@@ -6,7 +6,7 @@ object AppDependencies {
   val compile: Seq[ModuleID] = Seq(
```
udinobi, line 6: Would be worth to also update the test dependencies?
```
    "uk.gov.hmrc"            %% "bootstrap-test-play-28"   % bootstrapPlayVersion % "test",
    "org.mockito"            %% "mockito-scala"            % "1.17.12"            % "test",
```
Not sure, but I think in that case you can also remove `scalatestplus-play` and `flexmark-all`

udinobi (approved):

PR #1734, opened 2023-02-01
Ceds 4446 refactor of actions

darrylmurphy:


magicaljellybeans (approved):

glidester (changes requested):
Can you please make the test data in the specs representative of what they will actually be in reality (so Actions of type SubmissionRequest and CancellationRequest will have decId's with Some values rather than None). Thanks

glidester (approved):

PR #1740, opened 2023-02-16
CEDS-4415 Copy declaration when undergo under amendment. Draft implementation

udinobi:


app/controllers/AmendDeclarationController.scala
```diff
@@ -16,40 +16,38 @@
...
-      case Some((_, mrn, lrn, ducr)) => Future.successful(Ok(""))
-      case _                         => errorHandler.displayErrorPage
+  val submit: Action[AnyContent] = (authenticate andThen verifyEmail).async { implicit request =>
+    if (!declarationAmendmentsConfig.isEnabled) Future.successful(Redirect(RootController.displayPage))
+    else request.session.get(ExportsSessionKeys.submissionId) match {
+      case Some(submissionId) =>
+        connector.findOrCreateDraftForAmend(submissionId).map { declarationId =>
+          Redirect(SummaryController.displayPage)
+            .addingToSession(ExportsSessionKeys.declarationId -> declarationId)
+        }
```
udinobi, line 48: We still have a few "submission" keys in Session at this point. Do we need to remove them or we do not care to bring them forward on the next page?

glidester: If I'm understanding you correctly, keeping these session keys active should be fine as this new draft amend declaration is associated with the current submissionId (and will have the same DUCR, LRN and MRN values as it's parent). 
This however is opposed to when you create a completely new declaration by 'copying' an existing declaration. Then the new declaration will have a different submissionId & MRN (once it is submitted) and certainly will have a different DUCR and LRN. In that scenario these session keys should definitely be cleared. :+1:

app/controllers/CopyDeclarationController.scala
```diff
@@ -55,6 +56,7 @@ class CopyDeclarationController @Inject() (
       case Some(submission) if !isDeclarationRejected(submission) =>
         Redirect(CopyDeclarationController.displayPage)
           .addingToSession(ExportsSessionKeys.declarationId -> submissionId)
+          .removingFromSession(submissionDucr, ExportsSessionKeys.submissionId, submissionLrn, submissionMrn)
```
glidester, line 59: Thanks, good spot! :+1:

glidester (approved):
LGTM :+1:

PR #1751, opened 2023-03-07
CEDS-4558 Show Submission status in the title of the Timeline page

udinobi:


magicaljellybeans (approved):

glidester (approved):
Italian and Welsh, impressive! :wink: Did you google translate these new values to produce the Welsh? Just wondering if this is a good idea as it might not have produced well formed welsh sentences, and as you have added values these won't be picked up as needing translating by the Welsh team next time around :thinking: Maybe check with Aaron if he is happy with these values.

PR #1754, opened 2023-03-14
CEDS-4543 Add 'Amendment failed' event to the Timeline page

udinobi:


app/views/helpers/TimelineEvents.scala
```diff
@@ -92,12 +93,26 @@ class TimelineEvents @Inject() (
     }
   }
 
-  private def amendmentRejectedAsLatest(submission: Submission): Boolean =
-    submission.latestAction.exists { action =>
-      val isAmendmentRejected = action.requestType == AmendmentRequest
-      val hasLatestErrorNotification = action.notifications.headOption.exists(_.headOption.exists(_.enhancedStatus == ERRORS))
-      submission.blockAmendments == false && isAmendmentRejected && hasLatestErrorNotification
-    }
+  sealed trait AmendmentAsLatest
```
magicaljellybeans, line 96: I think this would make more sense as `AmendmentResult` or `AmendmentOutcome`

app/views/helpers/TimelineEvents.scala
```diff
@@ -92,12 +93,26 @@ class TimelineEvents @Inject() (
...
+  case object AmendmentFailed extends AmendmentAsLatest
+  case object AmendmentRejected extends AmendmentAsLatest
+
+  private def hasAmendmentAsLatest(submission: Submission): Option[AmendmentAsLatest] =
```
magicaljellybeans, line 100: And this method `hasAmendmentOutcome`

app/views/helpers/TimelineEvents.scala
```diff
@@ -115,15 +130,20 @@ class TimelineEvents @Inject() (
 
   private def isIndex1Primary(index1: Int, index2: Int): Boolean = index2 < 0 || index1 < index2
 
-  private def fixAndResubmitContent(id: String, isAmendmentRequest: Boolean)(implicit messages: Messages): Html = {
+  private def fixAndResubmitContent(id: String, maybeAmendmentToHandleAsLatest: Option[AmendmentAsLatest])(implicit messages: Messages): Html = {
```
magicaljellybeans, line 133: And `maybeAmendmentToHandleAsLatest` could be `amendmentOutcome` too?
`maybeAmendmentToHandleAsLatest` seems a bit obscure.

app/views/helpers/TimelineEvents.scala
magicaljellybeans: And then we can avoid the `maybe` prefix in variable names, because `None` on an `AmendmentOutcome` makes sense (to me at least!)

udinobi: What about `AmendmentAsLatestOutcome`? Because we are checking here if the `AmendmentRequest` action/notification is actually the latest one.

app/views/helpers/TimelineEvents.scala
udinobi: But `hasAmendmentOutcome` does not tell if the action/notification is the latest

glidester: How about `getAmendmentEventIfLatest`?

app/views/helpers/TimelineEvents.scala
udinobi: Sorry to correct... but it's the possible outcome of an amendment action and notification **as latest**.
If not latest `getAmendmentEventIfLatest()` will return `None`.

udinobi: imo, just name it as `AmendmentOutcome` could confuse in future who could add other amendment status even when the action/notification is not the latest.

magicaljellybeans: So harking back to your earlier comment Lucio - would `LatestAmendmentOutcome` work?

magicaljellybeans (dismissed):

PR #1772, opened 2023-03-27
CEDS-4412 Show 'External amendment' event on the timeline page

udinobi:


magicaljellybeans (dismissed):

app/views/helpers/TimelineEvents.scala
```diff
@@ -49,18 +49,30 @@ class TimelineEvents @Inject() (
...
   sfusConfig: SfusConfig,
+  secureMessagingInboxConfig: SecureMessagingInboxConfig,
+  declarationAmendmentsConfig: DeclarationAmendmentsConfig,
   uploadFilesPartialForTimeline: upload_files_partial_for_timeline
 ) {
   def apply(submission: Submission)(implicit messages: Messages): Seq[TimelineEvent] = {
 
-    val notificationEvents = createNotificationEvents(submission)
+    val allNotificationEvents = createNotificationEvents(submission)
+    val hasExternalAmendments = allNotificationEvents.exists(_.requestType == ExternalAmendmentRequest)
```
glidester, line 60: This is a little confusing. It looks like it is saying "If we have not received any ExternalAmendmentRequests" then display all notifications. I can see digging into the code that no ExternalAmendmentRequests events will be generated if the feature flag is disabled, but its not obvious from this code block.

Could you instead maybe always add both the ExternalAmendmentRequests and the SubmissionRequest notification of type AMEND and then filter them here based on the feature flag? So, if disabled filter out the ExternalAmendmentRequests events, if enabled filter out the SubmissionRequest notification? That would be more readable IMO :thinking:

glidester (approved):
Looks like it does what is needed but maybe it could be done in a more obvious way in the code? :)

app/views/helpers/TimelineEvents.scala
```diff
@@ -111,9 +117,27 @@ class TimelineEvents @Inject() (
...
 
-  sealed abstract class AmendmentEventIfLatest { val actionId: String }
-  case class AmendmentFailed(actionId: String) extends AmendmentEventIfLatest
-  case class AmendmentRejected(actionId: String) extends AmendmentEventIfLatest
+    if (declarationAmendmentsConfig.isEnabled) {
+      val hasExternalAmendments = allEvents.exists(_.requestType == ExternalAmendmentRequest)
+      if (hasExternalAmendments) allEvents.filterNot { event =>
+        event.requestType == SubmissionRequest && event.notificationSummary.enhancedStatus == AMENDED
+      }
+      else allEvents
```
glidester, line 125: Is this check necessary? Why not just always filter the events (if there are no ExternalAmendmentRequests then nothing will be filtered out)

udinobi: I do not understand. What check do you want to remove?

glidester (approved):

PR #1773, opened 2023-03-27
CEDS-4601: add extra txt to legal dec page for cancel

darrylmurphy:


app/controllers/declaration/SubmissionController.scala
```diff
@@ -84,6 +85,15 @@ class SubmissionController @Inject() (
         )
   }
 
+  def cancelAmendment(decId: Option[String]): Action[AnyContent] = (authenticate andThen verifyEmail) { implicit request =>
+    decId match {
+      case Some(id) if declarationAmendmentsConfig.isEnabled =>
+        Redirect(routes.SubmissionController.submitAmendment(Some("cancel"))).addingToSession((ExportsSessionKeys.declarationId, id))
+      case _ =>
+        Redirect(controllers.routes.SignOutController.userSignedOut)
```
udinobi, line 93: Are the requirements actually asking to redirect to the SignOutController if the flag is disabled? Or should the redirection be to the root controller?

darrylmurphy: You're right, that's been left in there from development

udinobi (approved):

udinobi (approved):

PR #1785, opened 2023-04-25
CEDS-4631 Modify DiffTools to handle explicitly sequenced objects

magicaljellybeans:
- Ensure DiffTools emits pointers pointing to explicitly sequenced object itself when added or removed.
- Add traits for ESOs to handle sequence id incrementing.
- Update controllers to use methods defined in those traits.


app/controllers/declaration/ItemsSummaryController.scala
```diff
@@ -154,11 +155,7 @@ class ItemsSummaryController @Inject() (
   private def removeItemFromCache(itemId: String)(implicit request: JourneyRequest[AnyContent]): Future[ExportsDeclaration] =
     request.cacheModel.itemBy(itemId) match {
       case Some(itemToDelete) =>
-        val updatedItems =
-          request.cacheModel.items.filterNot(_ == itemToDelete).zipWithIndex.map { case (item, index) =>
-            item.copy(sequenceId = index + 1)
-          }
-
+        val updatedItems = request.cacheModel.items.filterNot(_ == itemToDelete)
```
udinobi, line 158: As I assume the `==` op might be quite costly considering how nested is the `ExportItem` case class, wouldn't be just using the `id` field sufficient (and probably safer) for filtering out the instance to remove? WDYT?

magicaljellybeans: Yeah agree with that

udinobi (approved):

PR #1796, opened 2023-05-16
CEDS-4730 Fix for the 'Failed|Rejected Amendment' page where errors were not displayed

udinobi:


app/controllers/RejectedNotificationsController.scala
```diff
@@ -40,15 +40,40 @@ class RejectedNotificationsController @Inject() (
...
+                  case _ => errorHandler.internalServerError(s"Failed|rejected amended Notification not found for Action($actionId)??")
+                }
+
+              case _ => errorHandler.internalError(s"Failed|rejected amended declaration($declarationId) not found for Action($actionId)??")
+            }
+
+          case _ => errorHandler.internalError(s"The Action($actionId) does not have decId for a failed|rejected amendment??")
+        }
+
+      case _ => errorHandler.internalError(s"Action($actionId) not found for a failed|rejected amendment??")
```
glidester, line 76: Could this be converted to a for comprehension style syntax (deeply nested is hard to read).

glidester (changes requested):

app/controllers/RejectedNotificationsController.scala
udinobi: I thought about that, but then how do you handle all different error messages explaining what was the problem?

udinobi: In any case, this is not a "for comps", as the `flatMap`s are intertwined with a couple of pattern matches. I could split the block in multiple `def`s, but I'm not sure it would be any more readable.

app/controllers/RejectedNotificationsController.scala
```diff
@@ -40,15 +40,40 @@ class RejectedNotificationsController @Inject() (
...
+        action.decId match {
+          case Some(declarationId) =>
+            customsDeclareExportsConnector.findDeclaration(declarationId).flatMap {
+              case Some(declaration) =>
+                customsDeclareExportsConnector.findLatestNotification(actionId).map {
+                  case Some(notification) =>
+                    Ok(rejectedNotificationPage(declaration, notification.mrn, Some(declarationId), notification.errors))
+
+                  case _ => errorHandler.internalServerError(s"Failed|rejected amended Notification not found for Action($actionId)??")
+                }
```
glidester, line 68: Looking at this code trying to for comprehension it, I realised you don't need to call the DB to lookup the latest notification for that actionId as you retrieved the Action instance at the start of this block that contains all the notifications associated with it. That makes whole new BE endpoint redundant no?

udinobi: But we need to retrieve the list of errors, isn't it? And they are in the notification. Am I wrong?

glidester: Ok, ignore above my bad. The Action only contains a NotificationSummary not the full Notification.

app/controllers/RejectedNotificationsController.scala
glidester: This is the best I could come up with:
```
(for {
      action <- customsDeclareExportsConnector.findAction(actionId)
      maybeDeclarationId = action.flatMap(_.decId)
      maybeDeclaration <- customsDeclareExportsConnector.findDeclaration(maybeDeclarationId.get) if maybeDeclarationId.isDefined
      maybeNotification <- customsDeclareExportsConnector.findLatestNotification(actionId)
    } yield (maybeDeclaration, maybeNotification) match {
      case (Some(declaration), Some(notification)) =>
        Ok(rejectedNotificationPage(declaration, notification.mrn, maybeDeclarationId, notification.errors))
      case _ =>
        errorHandler.internalServerError(s"Failed|rejected amended Declaration or Notification not found for Action($actionId)??")
    }).recover {
      case e: Throwable => errorHandler.internalServerError(s"Failed|rejected amended Declaration or Notification not found for Action($actionId)??")
    }
```
Slightly better but missing some of the error detail of your solution. I'm not convinced that the above is significantly better than the current implementation so I'll leave it as an 'up to you' suggestion. :slightly_smiling_face:

glidester (approved):
One up to you suggestion

PR #1799, opened 2023-05-26
CEDS-4787 DEV: Fix issues with DiffTool producing incorrect pointers

glidester:
Fix Taric code pointer to point to 'taricCode'
Fix National Additional code pointer to point to 'nactCode' 
Fix CarrierDetails pointer to not repeat 'carrierDetails' part of pointer 
Fix ConsigneeDetails pointer to not repeat 'consigneeDetails' part of pointer 
Fix ConsignorDetails pointer to not repeat 'consignorDetails' part of pointer

### DIFF
--- a/app/forms/declaration/NactCode.scala
+++ b/app/forms/declaration/NactCode.scala
@@ -33,7 +33,7 @@ case class NactCode(nactCode: String) extends Ordered[NactCode] {
 object NactCode extends DeclarationPage with FieldMapping {
   implicit val format = Json.format[NactCode]
 
-  val pointer: ExportsFieldPointer = CommodityDetails.combinedNomenclatureCodePointer
+  val pointer: ExportsFieldPointer = "nactCode"
 
   val nactCodeKey = "nactCode"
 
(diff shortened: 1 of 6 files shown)


magicaljellybeans (dismissed):

app/forms/declaration/NactCode.scala
```diff
@@ -33,7 +33,7 @@ case class NactCode(nactCode: String) extends Ordered[NactCode] {
 object NactCode extends DeclarationPage with FieldMapping {
   implicit val format = Json.format[NactCode]
 
-  val pointer: ExportsFieldPointer = CommodityDetails.combinedNomenclatureCodePointer
+  val pointer: ExportsFieldPointer = "nactCode"
```
magicaljellybeans, line 36: When the pointers get produced for this and taric, will they have a sequenceId?  Because if we are deciding to point at the parent, we won't want that.

magicaljellybeans: `ExportsPointerToWCOPointer` throws an exception when there are mismatching quantities of $ sequence placeholders either side of a mapping.

glidester: Yes they will have sequenceId. We will have to make the BE handle these sequenceId (ignore them) because we want to keep this sequenceId data in the FE for other functionality (like showing users the difference between two declaration versions). :+1:

magicaljellybeans (approved):

PR #1807, opened 2023-06-12
CEDS-4778 Alter templates and messages to use the index plus one of t…

magicaljellybeans:
…he item when displaying an item to the user


app/controllers/helpers/SequenceIdHelper.scala
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package controllers.helpers
+
+import models.DeclarationMeta
+import models.DeclarationMeta.sequenceIdPlaceholder
+import models.declaration.DeclarationStatus.preSubmissionStatuses
+import models.declaration.{EsoKeyProvider, ExplicitlySequencedObject}
+
+class SequenceIdHelper {
```
udinobi, line 24: Why a `class`, which you need to inject, instead f an `object` ?

udinobi (approved):

app/controllers/helpers/SequenceIdHelper.scala
udinobi: Since you are unit-testing the controllers, you could check the "final" product, rather the intermediate steps (the params passed to `SequenceIdHandler`), by using `verify` on the resulting declaration passed as parameter to the `ExportsCache` mock. That's the final goal of unit tests, isn't it?

udinobi (approved):

PR #1814, opened 2023-06-15
CEDS-4798 Add submission id to session for legal declaration page, wh…

magicaljellybeans:
…ere it is used to backlink to declaration details page if the user came from there

### DIFF
--- a/app/views/declaration/summary/legal_declaration.scala.html
+++ b/app/views/declaration/summary/legal_declaration.scala.html
@@ -16,10 +16,12 @@
 
 @import config.AppConfig
 @import controllers.declaration.routes.{SubmissionController, SummaryController}
+@import controllers.routes.DeclarationDetailsController
 @import forms.declaration.LegalDeclaration._
 @import forms.declaration.LegalDeclaration
 @import models.declaration.submissions.EnhancedStatus.ERRORS
 @import models.requests.JourneyRequest
+@import models.requests.SessionHelper._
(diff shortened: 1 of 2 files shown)


app/views/helpers/TimelineEvents.scala
```diff
@@ -168,7 +168,7 @@ class TimelineEvents @Inject() (
         case _: AmendmentRejected | _ => linkButton("declaration.details.fix.resubmit.button", fixAndResubmit)
       }
 
-      val cancelUrl = SubmissionController.cancelAmendment(amendmentEventAsLatest.action.decId)
+      val cancelUrl = SubmissionController.cancelAmendment(amendmentEventAsLatest.action.decId, submission.uuid)
```
glidester, line 171: I see you are passing the SubmissionId here that the SubmissionController then sets as a session var. What I don't understand is how you can be on the timeline page without the SubmissionId already being set :thinking:

magicaljellybeans: So the submissionId is available on the timeline page, and it is used to form the timeline as a parameter to `TimelineEvents.apply(submission)`, here I am using that parameter in the formation of the url.

magicaljellybeans: But I need the submissionId on the legal declaration page, where previously it was not available.

glidester (approved):

glidester (approved):

PR #1820, opened 2023-06-21
CEDS-4812 Secure Exports FE login on TDR with additional secret value

glidester:
Adding new 'secret.tdrHashSalt' config value that will only be set in the TDR environment.

Adding additional authentication check that if the 'tdrHashSalt' value is present in config, then the user must specify an additional HMRC-CUS-ORG enrolment value of 'TDRSecret' who's value must be a valid hash of the supplied 'EORINumber' value.

Also removing redundant legacy appconfig values: "allowList.shutterPage", "allowList.ips","allowList.excludedPaths" & "allowList.enabled"

### DIFF
--- a/app/controllers/actions/AuthAction.scala
+++ b/app/controllers/actions/AuthAction.scala
@@ -20,18 +20,19 @@ import com.google.inject.{ImplementedBy, Inject, ProvidedBy}
 import com.kenshoo.play.metrics.Metrics
 import config.AppConfig
 import controllers.routes
-import models.AuthKey.{enrolment, identifierKey}
+import models.{IdentityData, SignedInUser}
+import models.AuthKey.{enrolment, hashIdentifierKey, identifierKey}
 import models.UnauthorisedReason.{UrlDirect, UserEoriNotAllowed, UserIsAgent, UserIsNotEnrolled}
 import models.requests.AuthenticatedRequest
-import models.{IdentityData, SignedInUser}
-import play.api.mvc._
 import play.api.{Configuration, Logging}
-import uk.gov.hmrc.auth.core.AffinityGroup.{Individual, Organisation}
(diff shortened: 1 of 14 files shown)


app/controllers/actions/AuthAction.scala
```diff
@@ -142,6 +148,28 @@ class AuthActionImpl @Inject() (
...
+  def allowListAuthentication(eori: String): Boolean = {
+    val eoriOnAllowList = eoriAllowList.allows(eori)
+
+    if (!eoriOnAllowList)
+      logger.info("Authentication Rejected: User's EORI not on allow list")
+
+    eoriOnAllowList
+  }
+
+  def tdrSecretAuthentication(eori: String, maybeHiddenSalt: Option[String], providedHash: String): Boolean =
```
magicaljellybeans, line 161: Nit: I don't like the prefix `maybe` when we can see that in the type, feels redundant

magicaljellybeans: fussy :smile:

magicaljellybeans (approved):

app/controllers/actions/AuthAction.scala
udinobi: Sorry if I disagree with you Tom :blush: I find it useful because then in higher-order functions you can then use `hiddenSalt =>`.
Less important but still to point out it's quite an accepted pattern in the FP community.

udinobi:
Weird heading... `Movements FE login`? Isn't this branch in Exports FE? :smirk:

app/controllers/actions/AuthAction.scala
magicaljellybeans: fair enough, I guess the type isn't always explicit

PR #1823, opened 2023-06-22
CEDS-4791: update goods-locations-codes json files per spreadsheet

Dweik97:


magicaljellybeans (changes requested):
Didn't notice before, but can you put the "en" and "cy" values in this format like the others:
_COMPANY NAME_ - _Address_
Just for the sake of uniformity.

udinobi (dismissed):

magicaljellybeans (approved):

PR #1869, opened 2023-08-31
CEDS-5015 Cancellation of amendments

udinobi:


test/services/SubmissionServiceSpec.scala
```diff
@@ -171,7 +171,7 @@ class SubmissionServiceSpec
         "declaration.totalNumberOfItems.totalAmountInvoiced",
         "declaration.totalNumberOfItems.exchangeRate"
       )
-      val expectedSubmissionAmendment = SubmissionAmendment(submissionId, "id2", expectedFieldPointers)
+      val expectedSubmissionAmendment = SubmissionAmendment(submissionId, "id2", false, expectedFieldPointers)
```
glidester, line 174: Should we add a test here for cancellation amendments (now that cancellations are distinct from other amendments)? :thinking:

udinobi: All methods are the same for both submission or cancellation of amendments. The only difference is this `Boolean` which btw does not change their behaviour, as all work is done by the BE. If you think, even the test at line #83 is a bit redundant since the DiffTool would in any case an empty list (unless I'm wrong).

udinobi: Still, I'll add a test to verify the connector receives the expected `SubmissionAmendment` instance.

glidester (approved):

PR #1897, opened 2023-09-27
Ceds 5030 total package qty gg je

darrylmurphy:


app/controllers/navigation/CommonNavigator.scala
```diff
@@ -40,7 +40,6 @@ trait CommonNavigator extends CacheDependentNavigators {
     case RepresentativeEntity          => routes.RepresentativeAgentController.displayPage
     case RepresentativeStatus          => routes.RepresentativeEntityController.displayPage
     case AdditionalDeclarationTypePage => routes.DeclarationChoiceController.displayPage
-    case NatureOfTransaction           => routes.TotalPackageQuantityController.displayPage
```
udinobi, line 43: Cannot this just be moved in `commonCacheDependent` instead of replicating it in every journey-specific section?

udinobi (approved):

PR #1901, opened 2023-10-02
Ceds 5124 mucr validation update

darrylmurphy:


conf/messages.cy
```diff
@@ -1,4 +1,5 @@
 date.day = Diwrnod
+date.day = Diwrnod
```
udinobi, line 2: Isn't this a duplication?

udinobi (approved):

PR #1905, opened 2023-10-06
CEDS-5156: fix tariff expanders in some C21 pages

Dweik97:


glidester (changes requested):
Can you add in a unit test to check correct links are displayed for the given dec type? The ATs won't be covering this sort of thing in future so the unit tests must. Thanks

glidester (approved):

PR #1929, opened 2023-10-25
CEDS-5180: fix print button post CSP changes

Dweik97:


app/views/components/buttons/print_page_button.scala.html
```diff
@@ -14,10 +14,20 @@
  * limitations under the License.
  *@
 
-@(marginTop: Int = 0, marginBottom: Int = 0)(implicit messages: Messages)
+@import views.html.helper.CSPNonce
+
+@(marginTop: Int = 0, marginBottom: Int = 0, buttonId: Int = 1)(implicit request: Request[_], messages: Messages)
```
udinobi, line 19: Not sure, but I think you do not need to pass the implicit `request: Request[_]`.
If that so, you can also avoid to pass it across all the call-sites.

udinobi (approved):

PR #1930, opened 2023-10-26
CEDS-5193: add tests and logic for skip of departure-transport on occ…

darrylmurphy:
…asional, simplified


app/controllers/declaration/DepartureTransportController.scala
```diff
@@ -71,23 +71,26 @@ class DepartureTransportController @Inject() (
...
+    else fun()
   }
 
-  private def nextPage(implicit request: JourneyRequest[AnyContent]): Result = {
-    val call =
+  private def nextPage(implicit request: JourneyRequest[AnyContent]): Result =
+    navigator.continueTo {
       if (request.declarationType == CLEARANCE) ExpressConsignmentController.displayPage
-      else if (request.cacheModel.isInlandOrBorder(Border)) TransportCountryController.displayPage
+      else if (request.cacheModel.isInlandOrBorder(Border) || (isSimplifiedOrOccasional && !isPostalOrFTI)) TransportCountryController.displayPage
```
udinobi, line 82: I think you do not need to change this. It should just be as it was before...
```
else if (request.cacheModel.isInlandOrBorder(Border)) TransportCountryController.displayPage
```

darrylmurphy: Do you mean to remove the `(isSimplifiedOrOccasional && !isPostalOrFTI)` condition here?

udinobi (approved):

PR #1933, opened 2023-10-27
CEDS-5187 Style as a Card the 'Declaration references' section of the CYA page

udinobi:
- 1st commit is just refactoring from `class` to `object` of `SequenceIdHelper`
- 2nd commit is what you should actually review according to the ticket's requirements.


glidester (approved):
:+1:

PR #1942, opened 2023-11-06
Ceds 5192 transport cya card

darrylmurphy:


app/views/helpers/summary/Card7ForTransport.scala
```diff
@@ -0,0 +1,241 @@
...
+
+  private def inlandModeOfTransportCode(declaration: ExportsDeclaration): Option[InlandModeOfTransportCode] =
+    declaration.locations.inlandModeOfTransportCode.filterNot(_.inlandModeOfTransportCode contains Empty)
+
+  private def borderTransport(transport: Transport, actionsEnabled: Boolean)(implicit messages: Messages): Option[SummaryListRow] =
+    transport.borderModeOfTransportCode.map { borderModeOfTransportCode =>
+      SummaryListRow(
+        key("transport.departure.transportCode.header"),
+        valueKey(s"declaration.summary.transport.departure.transportCode.${borderModeOfTransportCode.getCodeValue}"),
+        classes = "borderTransport",
```
udinobi, line 80: May I ask you to follow the same pattern I'm using? Even to respect the standard way to name CSS classes.
They are normally lower case with words separated by hyphens. So `border-transport`, `warehouse-id`, `supervising-office` and so on.

app/views/helpers/summary/Card7ForTransport.scala
```diff
@@ -0,0 +1,241 @@
...
+        changeLink(TransportLeavingTheBorderController.displayPage, "transport.departure.transportCode.header", actionsEnabled)
+      )
+    }
+
+  private def warehouseIdentification(declaration: ExportsDeclaration, actionsEnabled: Boolean)(implicit messages: Messages): Option[SummaryListRow] =
+    declaration.locations.warehouseIdentification.map { warehouseIdentification =>
+      SummaryListRow(
+        key(
+          if (warehouseIdentification.identificationNumber.isDefined) "transport.warehouse.id"
+          else "transport.warehouse.no.label"
```
udinobi, line 90: Nit: this could be a `fold` instead of `if else`

app/views/helpers/summary/Card7ForTransport.scala
```diff
@@ -0,0 +1,241 @@
...
+          classes = "modeOfTransport",
+          changeLink(InlandTransportDetailsController.displayPage, "transport.inlandModeOfTransport", actionsEnabled)
+        )
+      }
+    }
+
+  private def transportReference(transport: Transport, actionsEnabled: Boolean)(implicit messages: Messages): Option[SummaryListRow] =
+    if (transport.meansOfTransportOnDepartureType.isDefined || transport.meansOfTransportOnDepartureIDNumber.isDefined) {
+
+      val messagesForDepartureMeansOfTransport: Seq[String] =
```
udinobi, line 135: Honestly this block seems a bit complicated imo. It could be just simplified in
```
List(
  transport.meansOfTransportOnDepartureType.map(meansType => messages(s"declaration.summary.transport.departure.meansOfTransport.$meansType")),
  transport.meansOfTransportOnDepartureIDNumber.map(identity)
).flatten
```

app/views/helpers/summary/Card7ForTransport.scala
```diff
@@ -0,0 +1,241 @@
...
+
+  private def containers(transport: Transport, actionsEnabled: Boolean)(implicit messages: Messages): Option[List[SummaryListRow]] =
+    transport.containers flatMap { containers =>
+      val rows = containers.flatMap(containerRows(_, actionsEnabled))
+
+      if (rows.isEmpty) {
+        val header =
+          SummaryListRow(
+            key("containers"),
+            valueKey("site.no"),
```
udinobi, line 213: For consistency with other multi-rows sections, please use `site.none`

darrylmurphy: This has been pulled from existing code so it would represent a content change, so I think unless stated it should stay the same? Do we want to ask Tim?

(see: https://github.com/hmrc/customs-declare-exports-frontend/blob/main/app/views/helpers/summary/ContainersHelper.scala#L59)

app/views/helpers/summary/Card7ForTransport.scala
darrylmurphy: Agreed, ugly code, its re-used code from the section we're replacing. I'll have a look at refactoring.

(see: https://github.com/hmrc/customs-declare-exports-frontend/blame/main/app/views/declaration/summary/sections/transport_section.scala.html#L39)

Looks like its been there 4 years!

app/views/helpers/summary/Card7ForTransport.scala
darrylmurphy: No problem! Accident rather than by design

udinobi (approved):

PR #1948, opened 2023-11-13
CEDS-5208: Improve audit event for a Cancelled Declaration

Dweik97:


app/controllers/CancelDeclarationController.scala
```diff
@@ -51,53 +52,67 @@ class CancelDeclarationController @Inject() (
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport with Logging with WithUnsafeDefaultFormBinding {
 
-  def displayPage: Action[AnyContent] = (authenticate andThen verifyEmail).async { implicit request =>
+  def displayPage(adt: String): Action[AnyContent] = (authenticate andThen verifyEmail).async { implicit request =>
```
glidester, line 55: Not sure what 'adt' stands for. Can we expand this to a more meaningful name please

app/controllers/CancelDeclarationController.scala
```diff
@@ -51,53 +52,67 @@ class CancelDeclarationController @Inject() (
...
-    auditService.auditAllPagesDeclarationCancellation(cancelDeclaration)
+    for {
+      _ <- auditService.auditAllPagesDeclarationCancellation(cancelDeclaration)
+      maybeSubmission <- customsDeclareExportsConnector.findSubmission(submissionId)
+      status <- customsDeclareExportsConnector.createCancellation(cancelDeclaration)
+    } yield maybeSubmission match {
 
-    val context = exportsMetrics.startTimer(cancellationMetric)
+      case Some(sub) =>
+        auditData(sub, adt, userInput, Failure.toString, lrn, ducr, mrn).map { eventData =>
```
glidester, line 110: Why are we hardcoding this as a `Failure.toString()`???

glidester (dismissed):

app/controllers/CancelDeclarationController.scala
Dweik97: 'adt' is short for additionalDeclarationType. It's used elsewhere in the code but perhaps the lack of context here makes it vague. I could:
1- Add a comment to clarify what it stands for
2- Use the full form i.e. additionalDeclarationType
3- Use a more meaningful form such as addDecType

Which do you prefer?

glidester: 2,  personally i would use the name 'additionalDecType' :slightly_smiling_face:

Dweik97: I went for `additionalDecType` throughout 👍

conf/app.routes
```diff
@@ -50,8 +50,8 @@ GET         /ead-print-view/:mrn                           controllers.EADContro
 
 # Cancel declaration
 
-GET         /cancel-declaration                            controllers.CancelDeclarationController.displayPage
-POST        /cancel-declaration                            controllers.CancelDeclarationController.onSubmit
+GET         /cancel-declaration                            controllers.CancelDeclarationController.displayPage(additionalDecType: String)
```
glidester, line 53: I'd rather not pass this `additionalDecType` parameter on the query string unchecked. Can you instead add to the CancelDeclarationController the `JourneyAction` and grab the `additionalDecType` from the cache. This will prevent the (granted unlikely) possibility of user's re-writing the url and sticking in some rubbish query string value.

test/controllers/CancelDeclarationControllerSpec.scala
```diff
@@ -64,21 +70,28 @@ class CancelDeclarationControllerSpec extends ControllerWithoutFormSpec with Err
...
+        val result = controller.displayPage(additionalDeclarationType)(getRequestWithSession(sessionData.toSeq: _*))
         status(result) must be(OK)
       }
 
       "cancellation is requested with duplicate request error" in new SetUp {
-        cancelDeclarationResponse(CancellationAlreadyRequested)
-        val result = controller.onSubmit()(postRequestWithSession(correctCancelDeclarationJSON, sessionData.toSeq))
+        cancelDeclarationResponse(CancellationResult(CancellationAlreadyRequested, Some("conversationId")))
+
+        when(mockAuditService.auditAllPagesDeclarationCancellation(any())(any())).thenReturn(Future.successful(Success))
```
glidester, line 80: Maybe add a `verify` statement to prove that the audit service's method is called as expected

glidester (changes requested):

glidester (dismissed):

glidester (approved):

PR #1973, opened 2023-11-28
CEDS-5239 Show an error message on the Summary page if trying to submit a declaration without items

udinobi:


udinobi:
> * `SBT` can be updated to `1.9.7` in [project/build.properties](https://github.com/hmrc/customs-declare-exports-frontend/blob/ceds-5239-error-submitting-without-items/project/build.properties#L1)
> 
>     * `Scala` can be updated from `2.13.8` to `2.13.12` in [build.sbt](https://github.com/hmrc/customs-declare-exports-frontend/blob/ceds-5239-error-submitting-without-items/build.sbt#L57)
> 
>     * `uk.gov.hmrc.sbt-distributables` can be updated to `2.4.0` in [project/plugins.sbt](https://github.com/hmrc/customs-declare-exports-frontend/blob/ceds-5239-error-submitting-without-items/project/plugins.sbt#L6)

We'll need proper tickets for each of these hints, because applying any of them the compilations fails.

conf/declaration.routes
```diff
@@ -533,7 +533,9 @@ POST        /containers/:containerId/seals/:sealId/remove    controllers.declara
 
 # Summary
 
-GET         /saved-summary                       controllers.declaration.SummaryController.displayPage
+GET         /summary                             controllers.declaration.SummaryController.displayPage
```
glidester, line 536: Do we need to rename this endpoint? Just any tickets that reference this url will become stale so wondering if this is really needed? :thinking:

udinobi: Just to be more consistent with other related endpoints. In addition `saved-summary` was also sort of misleading IMO

glidester (approved):

PR #1975, opened 2023-11-30
CEDS-5150 Remove the 'Change' link for 'Net weight' on 'Items' card

udinobi:


glidester (approved):
Thanks for adding all those missing unit tests :+1:

PR #2000, opened 2024-01-10
CEDS-5330: Restrict 'Address' field on /consignee-details page to 35 characters

Dweik97:


app/views/components/fields/field_accessible_location_autocomplete_with_header.scala.html
```diff
@@ -74,7 +74,7 @@
     @if(hintParagraphs.nonEmpty) {
         <div id="@(fieldNameWithoutBrackets)-hint">
             @hintParagraphs.map { hint =>
-                <p class="govuk-hint">@hint</p>
+                <p class="govuk-hint" style="margin-top: 0px">@hint</p>
```
udinobi, line 77: Not a good practice to use `style`. It can be replaced by using the CSS class `govuk-!-margin-top-0`

udinobi (approved):

PR #2003, opened 2024-01-17
CEDS-5339: Enable switching between errors summary views and adding banners

Dweik97:


app/controllers/RejectedNotificationsController.scala
```diff
@@ -48,7 +48,7 @@ class RejectedNotificationsController @Inject() (
     extends FrontendController(mcc) with I18nSupport with Logging {
 
   def displayPage(id: String): Action[AnyContent] = (authenticate andThen verifyEmail).async { implicit request =>
-    if (newErrorReportConfig.isNewErrorReportEnabled)
+    if (newErrorReportConfig.isNewErrorReportEnabled && getValue(errorReportView).getOrElse("default") != "original")
```
glidester, line 51: Maybe use "version1|version2" rather than "default|original", more easily understood. Also should create some constants for these values rather than repeating these strings :wink:

app/models/requests/SessionHelper.scala
```diff
@@ -31,6 +31,8 @@ object SessionHelper extends Logging {
   val submissionLrn = "submission.lrn"
   val submissionMrn = "submission.mrn"
 
+  val errorReportView = "default"
```
glidester, line 34: Ah you have a constant, good. Just need to use it above :+1:

app/views/errors_reported.scala.html
```diff
@@ -51,9 +52,19 @@
     routes.DeclarationDetailsController.displayPage(submissionId.fold(declaration.id)(identity))
 }
 
+@researchBannerBody = @{
+     messages("rejected.notification.v2.banner.text",
+         link(
+            messages("rejected.notification.v2.banner.linkText.0"),
+            routes.RejectedNotificationsController.addToSessionThenDisplay(submissionId.fold(declaration.id)(identity), "original"),
```
glidester, line 59: Use constant for "original" value.

app/controllers/RejectedNotificationsController.scala
```diff
@@ -67,6 +67,11 @@ class RejectedNotificationsController @Inject() (
     }
   }
 
+  def addToSessionThenDisplay(id: String, errorReportVersion: String): Action[AnyContent] = (authenticate andThen verifyEmail).async {
+    implicit request =>
+      Future(Redirect(routes.RejectedNotificationsController.displayPage(id)).addingToSession(errorReportView -> errorReportVersion))
```
glidester, line 72: Here you are using a HTTP 303 redirect to essentially just call the `displayPage` method above (causing a second round trip to the browser and server. Could you not just call the `displayPage` method instead (similar to what you are doing on line 52)? :thinking:

glidester (commented):
Generally look good Mo. Just a couple of small niggles really :+1:

glidester (approved):

PR #2012, opened 2024-01-30
CEDS-5424 Add an info banner to the 'National Code' page

udinobi:


app/views/components/gds/notificationBanner.scala.html
```diff
@@ -16,9 +16,11 @@
 
 @this()
 
-@(title: String, content: Html)
+@(title: String, content: Html, classes: Seq[String] = List.empty)
 
-<div class="govuk-notification-banner" role="region" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
+@cssClasses = @{ if (classes.isEmpty) "" else s""" ${classes.mkString(" ")}""" }
```
glidester, line 21: Is the `if` necessary? if `classes ` is an empty List would not `${classes.mkString(" ")` just produce an empty string anyway? :thinking:

glidester (commented):
One small niggle

app/views/components/gds/notificationBanner.scala.html
udinobi: I was a bit punctilious here... In case of an empty list the html would be `<div class="govuk-notification-banner "` with a trailing blank after `govuk-notification-banner` . The `if` avoid that :smirk:

udinobi: Yes, it's necessary. We need to prefix with a blank the given classes from the `govuk-notification-banner` class.
Please note the blank before the dollar sign => `s"""<blank>$`

glidester: Can I suggest:
`@cssClasses = @{ (classes :+ "govuk-notification-banner").mkString(" ") }`
as a simpler solution?

glidester (approved):

PR #2030, opened 2024-02-26
Ceds 5484 errors v2 to convert base values to human readable formats

glidester:


app/views/helpers/PointerRecord.scala
```diff
@@ -16,18 +16,33 @@
...
+  def fetchRawValue(dec: ExportsDeclaration, args: Int*): Option[String]
+  def fetchReadableValue(
+    dec: ExportsDeclaration,
+    args: Int*
+  )(implicit msgs: Messages, countryHelper: CountryHelper, codeListConnector: CodeListConnector): Option[String]
   val pageLink1Param: Option[Call] = None
   val pageLink2Param: Option[String => Call] = None
 }
 
+abstract class DefaultPointerRecord extends PointerRecord {
```
udinobi, line 38: This could still be a `Trait`, isn't it?

conf/application.conf
```diff
@@ -109,7 +109,7 @@ microservice {
       declarationAmendments = enabled
       default = disabled
       merchandiseInBag = disabled
-      newErrorReport = disabled
+      newErrorReport = enabled
```
magicaljellybeans, line 112: Will this now be enabled in all environments?

glidester: Yes. Dhanist will be writing the full AC tests for the new Errors screen for this ticket

app/views/helpers/PointerRecord.scala
glidester: I only added this abstract class to get around the annoying issue of "parameter x method is never used" error if I left the default implementation in the trait. For some reason Scala does not like you defining default method implementations in traits where you don't use all the method parameters, but method implementations in abstract classes it has no problems with! :shrug:

udinobi (approved):

PR #2033, opened 2024-02-28
CEDS-5421 Cache border transport nationality as country code while ma…

magicaljellybeans:
…intaining country names in UI

### DIFF
--- a/app/forms/declaration/TransportCountry.scala
+++ b/app/forms/declaration/TransportCountry.scala
@@ -29,7 +29,7 @@ import play.api.data.validation.{Constraint, Invalid, Valid, ValidationError}
 import play.api.data.{Form, Forms, Mapping}
 import play.api.i18n.Messages
 import play.api.libs.json.{Json, OFormat}
-import services.Countries.isValidCountryName
+import services.Countries.isValidCountryCode
 import utils.validators.forms.FieldValidator._
 
 case class TransportCountry(countryName: Option[String]) extends Ordered[TransportCountry] with Amendment {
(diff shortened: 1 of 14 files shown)


app/forms/declaration/TransportCountry.scala
```diff
@@ -44,6 +46,9 @@ case class TransportCountry(countryName: Option[String]) extends Ordered[Transpo
 
   def value: String = countryName.getOrElse("")
 
+  def displayValue(countryHelper: CountryHelper)(implicit codeListConnector: CodeListConnector, messages: Messages): String =
```
udinobi, line 49: IMO this method does not belong here. I would move it in `view.helpers.CountryHelper` and call it from there in the view or where else need it.

app/forms/declaration/countries/Country.scala
```diff
@@ -36,6 +39,9 @@ case class Country(code: Option[String]) extends Ordered[Country] with Amendment
 
   def value: String = code.getOrElse("")
 
+  def displayValue(countryHelper: CountryHelper)(implicit codeListConnector: CodeListConnector, messages: Messages): String =
```
udinobi, line 42: Same here.

app/forms/declaration/TransportCountry.scala
magicaljellybeans: The call site is in PointerRecord for the various fields that have a cached country code.
I agree, it should be in the views package though.

udinobi: I could not find it because `PointerRecord`, being a big file, is by default not expanded by Github :slightly_smiling_face:

PR #2034, opened 2024-03-01
Ceds 5421 make border transport nationality field populate with country code

magicaljellybeans:


app/views/helpers/PointerRecord.scala
```diff
@@ -44,7 +46,8 @@ abstract class DefaultPointerRecord extends PointerRecord {
 }
 
 // scalastyle:off
-object PointerRecord {
+@Singleton
+class PointerRecords @Inject() (countryHelper: CountryHelper) {
```
glidester, line 50: Never considered making this a singleton instance, probably a better choice than the companion object :+1: 

But, if we are making this a singleton instance with injected dependencies then we might as well do the same for the CodeListConnector dependency and remove both from the `fetchReadableValue` method implicit signature :thinking:

glidester (dismissed):

glidester (approved):

PR #2051, opened 2024-04-09
CEDS-5683 Improvements to the 'Ducr' pages of section-1

udinobi:


app/views/declaration/ducr_entry.scala.html
```diff
@@ -36,13 +36,19 @@
...
     formHelper: FormWithCSRF,
-    navigator: Navigator,
     appConfig: AppConfig
 )
 
 
 @(form: Form[Ducr])(implicit request: JourneyRequest[_], messages: Messages)
 
+@backLink = @{
+  request.cacheModel.ducr.fold(DucrChoiceController.displayPage) { ducr =>
```
magicaljellybeans, line 46: Can this logic not go in one of the navigator classes?

app/views/declaration/local_reference_number.scala.html
```diff
@@ -36,21 +37,27 @@
...
     inputWithTitleAsLabel: inputWithTitleAsLabel,
     formHelper: FormWithCSRF,
-    navigator: Navigator,
     appConfig: AppConfig
 )
 
 
 @(form: Form[Lrn])(implicit request: JourneyRequest[_], messages: Messages)
 
+@backLink = @{
```
magicaljellybeans, line 46: Same point here

app/views/declaration/ducr_entry.scala.html
udinobi: For simple logic like this, and even more considering that is common to all journey types, honestly I do not see the point to involve the Navigator. IMO it would only make things more complicated.

magicaljellybeans: It seems like splitting responsibility for navigation by having some in the view and some in the navigator - I know we probably do this on other pages too - but it could be confusing for someone unfamiliar with the codebase.

udinobi: In any case you always go before to the page to figure out the key used for the Navigator's pattern matching.

glidester: I agree, if we can put this logic in the Navigator class then we should to keep the logic in one place wherever possible :+1:

udinobi: And personally I would remove The "Common" pattern from the the Navigator. It seems to me we are over-engineering the Navigator just for the "responsibility"

PR #2059, opened 2024-04-16
Ceds 5706 add in missing amendment cancellation pre submit audit event

glidester:


Dweik97 (approved):

app/services/SubmissionService.scala
```diff
@@ -77,12 +77,10 @@ class SubmissionService @Inject() (connector: CustomsDeclareExportsConnector, au
...
-          val auditType =
-            if (isCancellation) AmendmentCancellation
-            else {
-              auditService.auditAllPagesUserInput(AmendmentPayload, declaration)
-              AuditTypes.Amendment
-            }
+
+          val auditTypeAndFunction =
+            if (isCancellation) (AmendmentCancellation, () => auditService.auditAllPagesUserInput(AmendmentCancellationPayload, declaration))
+            else (AuditTypes.Amendment, () => auditService.auditAllPagesUserInput(AmendmentPayload, declaration))
```
udinobi, line 83: It really seems over-engineered. It could simply be...
```
val auditValues =
  if (isCancellation) (AmendmentCancellation, AmendmentCancellationPayload)
  else (AuditTypes.Amendment, AmendmentPayload)
```
and then below instead of `auditTypeAndFunction._2()` you could use `auditService.auditAllPagesUserInput(auditValue._2, declaration)`

glidester: Yeah, reason I went that way is that I felt uncomfortable having a tuple with two elements of the same type (as its too easy then to mistakenly use ._1 instead of ._2 and the compiler would still be happy).
But on reflection it is just replacing one code smell with another so I've taken your suggestion :slightly_smiling_face:

PR #2060, opened 2024-04-16
CEDS-5658 Improve validation of auto-complete input fields in case of errors

udinobi:


app/utils/validators/forms/AutoCompleteFieldBinding.scala
```diff
@@ -0,0 +1,40 @@
...
+import play.api.data.FormBinding
+import play.api.mvc.Request
+
+trait AutoCompleteFieldBinding {
+
+  // Suffix added by 'auto-complete-fix.js' to the form-binding key of the auto-complete input field.
+  private val suffix = "-autocomp"
+  private val suffixLen = suffix.length
+
+  def formValuesFromRequest(key: String)(implicit request: Request[_], formBinding: FormBinding): Map[String, Seq[String]] = {
```
magicaljellybeans, line 28: I found this in this comment where (I think) this implicit is coming from
```
object FormBinding {
  object Implicits {

    /**
     * Convenience implicit for testing and other scenarios where it's not important
     * to verify payload limits (e.g. prevent OutOfMemoryError's).
     *
     * Prefer using a FormBinding provided by PlayBodyParsers#formBinding since that honours play.http.parser.maxMemoryBuffer limits.
     */
    implicit val formBinding: FormBinding = new DefaultFormBinding(Form.FromJsonMaxChars, Form.FromJsonMaxDepth)
  }
}
```

magicaljellybeans: I think that's probably fine in this instance though

app/utils/validators/forms/FieldValidator.scala
```diff
@@ -51,11 +50,11 @@ object FieldValidator {
 
   val isPresent: Option[_] => Boolean = _.nonEmpty
 
-  val isMissing: Option[_] => Boolean = _.isEmpty
+  val isNone: Option[_] => Boolean = _.isEmpty
```
glidester, line 53: Nit: If changing this name then `isPresent` should be renamed to `isSome` for consistency :+1:

glidester (dismissed):

magicaljellybeans (approved):

PR #2089, opened 2024-05-21
CEDS-5774 Ensure that when any details are updated that the country i…

magicaljellybeans:
…s displayed as a name not a code

Remove some unnecessary tests for details where it is impossible for the user to add an address

### DIFF
--- a/app/views/helpers/AmendmentDetailsHelper.scala
+++ b/app/views/helpers/AmendmentDetailsHelper.scala
@@ -226,18 +226,35 @@ class AmendmentDetailsHelper @Inject() (
     def fetchCountry(countryCode: Option[String]): Option[String] =
       countryCode.flatMap(countryHelper.getShortNameForCountryCode)
 
-    def updateRoutingCountry(routingCountry: RoutingCountry): Option[RoutingCountry] =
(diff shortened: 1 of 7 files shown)


app/views/helpers/AmendmentDetailsHelper.scala
```diff
@@ -226,18 +226,34 @@ class AmendmentDetailsHelper @Inject() (
...
 
-    def updateRoutingCountry(routingCountry: RoutingCountry): Option[RoutingCountry] =
+    def handleRoutingCountry(routingCountry: RoutingCountry): Option[RoutingCountry] =
       Some(routingCountry.copy(country = Country(fetchCountry(routingCountry.country.code))))
 
-    def handlePotentialCountryCode(maybeCode: String): String =
-      if (maybeCode.isBlank) maybeCode
+    def handleString(maybeCode: String): Option[String] =
+      if (maybeCode.isBlank) Some(maybeCode)
       else if (maybeCode.length == 2 && isValidCountryCode(maybeCode))
```
udinobi, line 234: Nit: would it maybe worth to have here something like `else if (maybeCode.trim.length == 2` ? Since you are also testing with `isBlank`

udinobi (dismissed):

udinobi (approved):

PR #2097, opened 2024-05-30
CEDS-5850 Change the Html of the MRN paragraph on 'Error Notification' pages

udinobi:


udinobi:
>     * `STDOUT_IGNORE_NETTY` can be removed from [conf/logback.xml](https://github.com/hmrc/customs-declare-exports-frontend/blob/ceds-5850-update-html-of-errors-page/conf/logback.xml#L17) - Netty server is no longer used (Prior to Play 2.6.x)

Did not notice that there were 2 definitions of `STDOUT_IGNORE_NETTY` and I removed only one. I'll remove the 2nd one in my next PR.

glidester (approved):

PR #2102, opened 2024-05-31
CEDS-5827 Suppress any exception thrown when EAD DIS fails to find the related declaration

udinobi:
2nd commit is actually related to the ticket. The 1st commit is just some test refactoring


test/controllers/EADControllerSpec.scala
```diff
@@ -16,46 +16,50 @@
...
+
+    "return 200 and display the EAD page" when {
+      "the declaration information are found" in {
         val result = controller.generateDocument(mrn).apply(getRequest())
 
         status(result) must be(OK)
+      }
+    }
+
+    "return 400 (BAD_REQUEST)" when {
```
magicaljellybeans, line 55: Nit-picky but should we not be propagating the 404 here?

magicaljellybeans (approved):

test/controllers/EADControllerSpec.scala
udinobi: Actually it's the ErrorHandler handling this and the standard error page. Since we just log the error, it's not important if we propagate a 404 or a 400. In Kibana we'll only see the message from the logger.

PR #2107, opened 2024-06-05
CEDS-5855 Remove 'Secure-Messaging' feature

udinobi:


app/services/audit/AuditService.scala
```diff
@@ -149,15 +133,12 @@ class AuditService @Inject() (connector: AuditConnector, appConfig: AppConfig)(i
...
         json
     }
-
-  private def tagsForMessageInboxPartialRetrieved(path: String)(implicit hc: HeaderCarrier): Map[String, String] =
-    AuditExtensions.auditHeaderCarrier(hc).toAuditTags(transactionName = "callExportPartial", path = path)
 }
 
 object AuditTypes extends Enumeration {
   type Audit = Value
-  val SaveDraftValue, Submission, SubmissionPayload, Cancellation, CancellationPayload, NavigateToMessages, Amendment, AmendmentPayload,
```
glidester, line 159: Are we sure 'NavigateToMessages' isn't used for the SFUS links? :thinking:

glidester (commented):
Just the audit event question. Unsure if we are supposed to audit the handing over to another service as an explicit event :thinking:

app/services/audit/AuditService.scala
udinobi: It was only used when auditing an event for partial retrieved messages to the inbox, so it was expected to be removed

glidester (approved):

PR #2116, opened 2024-06-12
CEDS-5833 Update connector call to match modified BE endpoint

magicaljellybeans:


udinobi (approved):

udinobi (approved):
:+1:

app/models/ExportsDeclaration.scala
```diff
@@ -40,6 +40,7 @@ import services.DiffTools.{combinePointers, compareDifference, compareIntDiffere
 case class ExportsDeclaration(
   id: String,
   declarationMeta: DeclarationMeta,
+  eori: String = "", // Included for parity with BE model only
```
glidester, line 43: Why do we need to default to empty string? :thinking:

magicaljellybeans: I thought that would make more sense given that we aren't using the field in the FE instead of pulling the actual EORI into it.

glidester (approved):

PR #2119, opened 2024-06-13
CEDS-5868 Fix issue when trying to cancel a failed/rejected amendment

udinobi:


app/views/helpers/TimelineEvents.scala
```diff
@@ -193,8 +194,18 @@ class TimelineEvents @Inject() (
...
-      val cancelLink = link(messages("declaration.details.cancel.amendment"), cancelUrl, id = Some("cancel-amendment"))
+      val cancelLink =
+        (for {
+          amendmentEventIfLatest <- maybeAmendmentEventIfLatest
+          declarationId <- amendmentEventIfLatest.action.decId
+        } yield {
+          val cancelUrl = SubmissionController.cancelAmendment(declarationId)
+          link(messages("declaration.details.cancel.amendment"), cancelUrl, id = Some("cancel-amendment"))
+        }).getOrElse {
+          logger.warn(s"Failed/rejected amendment for submission(${submission.uuid}) cannot be cancelled due to missing action.decId")
```
glidester, line 205: Is this log statement accurate? A user might get here simply because this is not a failed amendment scenario (but a failed submission). Any time `maybeAmendmentEventIfLatest` is None, rather than only if `amendmentEventIfLatest.action.decId` is None :thinking:

udinobi: Yes, because if `maybeAmendmentEventIfLatest` is None, we only return a `linkButton`. There's a `fold` to check that.

When we enter the "true" part of the `fold` it's safe to attest that the amendment is either failed or rejected, and for both of them we want to show the `cancel` link.

And we can only show the `cancel` link if we know the id of the amended declaration, which is optional in `Action`.

glidester (approved):

PR #2120, opened 2024-06-14
CEDS-5890 New implementation for stripping empty Json values from Audit Events

udinobi:


glidester (commented):
Do we have any tests to check this method handles the scenario that the old method did not (where it threw the warning log messages)?

udinobi:
We have 2 tests in [AuditServiceSpec](https://github.com/hmrc/customs-declare-exports-frontend/blob/19f0471ac138d45a5044395aa15915529bbb2e5b/test/services/audit/AuditServiceSpec.scala#L95)

glidester (approved):

PR #2125, opened 2024-06-19
Ceds 5846 add missing error pointers

magicaljellybeans:


udinobi (approved):

app/models/viewmodels/errors/ErrorInterpreter.scala
```diff
@@ -99,7 +100,18 @@ trait ErrorInterpreter {
...
+        case `pointerToAdditionalProcedureCodes` =>
+          messages(
+            field.pointer.messageKey,
+            field.pointer.sequenceArgs.head,
+            field.pointer.sequenceArgs.last.toIntOption.fold("")(x => (x - 1).toString)
+          )
+        case _ =>
+          messages(field.pointer.messageKey, field.pointer.sequenceArgs: _*)
+
+      }
```
glidester, line 114: Do you still need this now? :thinking:

magicaljellybeans: Yes, this is for the CDS12056 error that occurs with an additional procedure code pointer, which we previously didn't have.

magicaljellybeans: Well, technically it would be for any error code that comes through with an APC pointer.

glidester (approved):

udinobi (approved):

PR #2147, opened 2024-07-24
CEDS-5830 draft

magicaljellybeans:
- Addition of AmendmentHelper
- Addition of amendKeys to PointerRecord class (partially complete)

TODO add logic that fetches all leaf (only) pointer records when a parent level pointer comes through (via the DiffTool)
Capture any logic from valueAdded and other methods on case classes extending Amendment/AmendmentOp traits

### DIFF
--- a/app/controllers/helpers/AmendmentHelper.scala
+++ b/app/controllers/helpers/AmendmentHelper.scala
@@ -38,9 +38,9 @@ class AmendmentHelper @Inject() (implicit codeListConnector: CodeListConnector, 
       }
     }
 
-    val pointersAndRecords = amendedPointers.map(pointer => (pointer, PointerRecord.pointersToPointerRecords(pointer.pattern)))
+    amendedPointers.flatMap { pointer =>
+      val record = PointerRecord.pointersToPointerRecords(pointer.pattern)
 
-    pointersAndRecords.flatMap { case (pointer, record) =>
       record.amendKey.fold {
         logger.warn(s"No amend key found for pointer [$pointer]")
         Option.empty[AmendmentInstance]
--- a/app/forms/common/YesNoAnswer.scala
+++ b/app/forms/common/YesNoAnswer.scala
@@ -18,9 +18,8 @@ package forms.common
 
 import forms.common.YesNoAnswer.YesNoAnswers.yes
 import forms.mappings.MappingHelper.requiredRadio
+import models.Amendment
 import models.ExportsFieldPointer.ExportsFieldPointer
-import models.declaration.Parties
-import models.{Amendment, ExportsDeclaration}
 import play.api.data.{Form, Forms, Mapping}
 import play.api.i18n.Messages
 import play.api.libs.json.{Json, OFormat}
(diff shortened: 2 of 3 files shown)


app/controllers/helpers/AmendmentHelper.scala
```diff
@@ -0,0 +1,58 @@
...
+      val diff = amendedDeclaration.createDiff(originalDeclaration)
+      diff.flatMap { alteredField =>
+        PointerPatterns.expandPointer(Pointer.apply(alteredField.fieldPointer), originalDeclaration, amendedDeclaration)
+      }
+    }
+
+    val pointersAndRecords = amendedPointers.map(pointer => (pointer, PointerRecord.pointersToPointerRecords(pointer.pattern)))
+
+    pointersAndRecords.flatMap { case (pointer, record) =>
+      record.amendKey.fold {
```
udinobi, line 44: The `amendedPointers.map(` loop seems superfluous. Cannot L41-L44 be reduced to...
```
amendedPointers.flatMap { pointer =>
  val record = PointerRecord.pointersToPointerRecords(pointer.pattern)
  record.amendKey.fold {
```

app/forms/common/YesNoAnswer.scala
```diff
@@ -35,28 +33,15 @@ case class YesNoAnswer(answer: String) extends Ordered[YesNoAnswer] with Amendme
...
 object YesNoAnswer {
 
   implicit val format: OFormat[YesNoAnswer] = Json.format[YesNoAnswer]
 
   private lazy val parties = s"${ExportsDeclaration.pointer}.${Parties.pointer}"
-  private lazy val transport = s"${ExportsDeclaration.pointer}.${Transport.pointer}"
 
-  lazy val mappingsForAmendment = Map(
-    s"$parties.personPresentingGoodsDetails.eori" -> "declaration.summary.parties.eidr",
-    s"$transport.expressConsignment" -> "declaration.summary.transport.expressConsignment"
```
udinobi, line 58: Just to be sure... don't we need any more the `expressConsignment` mapping?

app/views/helpers/PointerPatterns.scala
```diff
@@ -0,0 +1,228 @@
...
+  val expandAdditionalActors = (p: Pointer, orig: ExportsDeclaration, amend: ExportsDeclaration) =>
+    Seq(Pointer(p.sections :+ PointerSection("eori", FIELD)), Pointer(p.sections :+ PointerSection("type", FIELD)))
+
+  val expandCusCode = (p: Pointer, orig: ExportsDeclaration, amend: ExportsDeclaration) =>
+    Seq(Pointer(p.sections :+ PointerSection("cusCode", FIELD)))
+
+  val expandConsignorDetails = (p: Pointer, orig: ExportsDeclaration, amend: ExportsDeclaration) =>
+    Seq(Pointer(p.sections :+ PointerSection("eori", FIELD)))
+
+  val expandDetails = (p: Pointer, orig: ExportsDeclaration, amend: ExportsDeclaration) => Seq(Pointer(p.sections :+ PointerSection("eori", FIELD)))
```
udinobi, line 37: What's this for? I do not see differences with `expandConsignorDetails`

glidester: Yes good spot, looks like we can just reuse the generic `expandDetails` function.

app/controllers/helpers/AmendmentHelper.scala
glidester: Yes, will rationalise it down to a single loop :+1:

app/forms/common/YesNoAnswer.scala
glidester: We don't need this construct at all now. Will remove it from the code completely :+1:

udinobi (approved):

PR #2148, opened 2024-07-25
CEDS-5921 Update the main content of 'Location of Goods' for arrived declarations

udinobi:


app/forms/section3/LocationOfGoods.scala
```diff
@@ -69,23 +69,50 @@ object LocationOfGoods extends DeclarationPage {
...
-  private def mapping(implicit messages: Messages, codeListConnector: CodeListConnector): Mapping[LocationOfGoods] =
+  lazy val goodsLocations: Seq[String] = List(
+    "GBAUDEUDEUDEUGVM",
+    "GBAUDVRDOVDVRGVM",
+    "GBAUEUTEUTEUTGVM",
+    "GBAUFISFISFISGVM",
+    "GBAUHYMHEYHYMGVM",
+    "GBAUHLYHLDHLYGVM",
+    "GBAULIVLIVLIVGVM",
+    "GBAUPEDMILPEDGVM",
```
glidester, line 90: Can we avoid hard coding these values here? This will be lost/forgotten if we ever need to update these in our reference data files.

glidester (changes requested):

glidester (dismissed):

PR #2170, opened 2024-09-09
CEDS-5867 Add 'Back-to-Choice-page' link to 'Error' template

udinobi:


kenneth-rayner (approved):

test/controllers/general/ErrorHandlerSpec.scala
```diff
@@ -41,21 +44,35 @@ class ErrorHandlerSpec extends UnitWithMocksSpec with Stubs with OptionValues wi
...
-      result must include("message")
+    "include default title, heading and message" in {
+      checkView(errorHandler.standardErrorTemplate("global.error.title", "", "")(request).futureValue)
+    }
+
+    "include expected title, heading and message" in {
+      val messages = messagesApi.preferred(request).messages
+      val heading = messages("global.error.heading")
+      val message = messages("global.error.message")
+      checkView(errorHandler.standardErrorTemplate("global.error.title", heading, message)(request).futureValue)
```
glidester, line 67: This test does not seem to prove that you can override the default values as it uses the default values, so is it using the ones specified or the default we can't prove (as they are the same) :thinking:

app/controllers/general/ErrorHandler.scala
```diff
@@ -57,11 +55,14 @@ class ErrorHandler @Inject() (override val messagesApi: MessagesApi, errorTempla
...
-    titleKey: String = "global.error.title",
-    headingKey: String = "global.error.heading",
-    messageKey: String = "global.error.message"
-  )(implicit request: Request[_]): Html = errorTemplate(titleKey, headingKey, messageKey)
+  def defaultErrorTemplate(titleKey: String = "global.error.title", heading: String = "", message: String = "")(
+    implicit request: Request[_]
+  ): Html = {
+    lazy val messages = messagesApi.preferred(request).messages
+    val headingText = if (heading.isEmpty) messages("global.error.heading") else heading
+    val messageText = if (message.isEmpty) messages("global.error.message") else message
```
glidester, line 63: Need to exercise these logic flows in our tests

glidester (changes requested):

test/controllers/general/ErrorHandlerSpec.scala
udinobi: ??
The 1st test only use the default messages, passing the `heading` and `message` parameters as empty Strings, then testing the `true` branch at #L62 and #L63, whereas in the 2nd test the 2 parameters' values are retrieved at #L65 and #L66, so testing the `false branch at #L62 and #L63. The fact that we are using the same message keys is not relevant.

glidester (approved):

PR #2182, opened 2024-10-04
CEDS-6081 Update the 'Rejected Notifications' page to prevent a A11y violation

udinobi:


magicaljellybeans (approved):

app/models/viewmodels/errors/ErrorInterpreter.scala
```diff
@@ -57,31 +57,35 @@ trait ErrorInterpreter {
     }
   }
 
-  def fieldsTableHeader(implicit messages: Messages) = Html(s"""<table class="govuk-table">
+  def fieldsTableHeader(implicit messages: Messages): Html = Html(
+    // The <table> element is enclosed in a <div> as <dl> elements can only
+    // contain <dt> and <dd> groups, <script>, <template> or <div> elements.
```
glidester, line 62: Not sure we need this comment here. Maybe just have it as part of the commit comment (or as a PR comment maybe) :+1:

glidester (approved):

app/models/viewmodels/errors/ErrorInterpreter.scala
udinobi: I added because I'm worried that in future someone could remove the `<div>` element thinking that it might be not useful.

magicaljellybeans (approved):

PR #2189, opened 2024-10-11
CEDS-5998 Enable users to amend declarations even in case of DMSRES notifications

udinobi:


app/models/declaration/submissions/Submission.scala
```diff
@@ -41,12 +41,12 @@ case class Submission(
     } yield notificationSummaries.map(_.enhancedStatus)
   ).getOrElse(Seq.empty[EnhancedStatus])
 
-  lazy val blockAmendments: Boolean = latestDecId.isEmpty
+  lazy val blockOldAmendments: Boolean = latestDecId.isEmpty
```
glidester, line 44: Just thinking the name `blockOldAmendments` might be confusing to our successors. Maybe leave as `blockAmendments` and add a comment explaining this `isEmpty` will only be true for legacy declarations created before this change?

app/models/declaration/submissions/SubmissionStatus.scala
```diff
@@ -25,16 +25,14 @@ object SubmissionStatus extends Enumeration {
   implicit val format: Format[SubmissionStatus.Value] = EnumJson.format(SubmissionStatus)
 
   val PENDING, REQUESTED_CANCELLATION, ACCEPTED, RECEIVED, REJECTED, UNDERGOING_PHYSICAL_CHECK, ADDITIONAL_DOCUMENTS_REQUIRED, AMENDED, RELEASED,
-    CLEARED, CANCELLED, CUSTOMS_POSITION_GRANTED, CUSTOMS_POSITION_DENIED, GOODS_HAVE_EXITED_THE_COMMUNITY, DECLARATION_HANDLED_EXTERNALLY,
```
glidester, line 28: Why have you removed `GOODS_HAVE_EXITED_THE_COMMUNITY` sub status?

udinobi: Never  used

glidester: Yes please. It is certainly used on the BE https://github.com/hmrc/customs-declare-exports/blob/ad2daf7d76c82ad14de2f44ecdbb1c8e2036803e/app/uk/gov/hmrc/exports/models/declaration/submissions/SubmissionStatus.scala#L47. Need to keep FE and BE in sync

glidester (approved):

PR #2202, opened 2024-10-29
CEDS-6129 Fix the issue with the 'Amendment Details' page showing no differences

udinobi:


glidester (approved):
We will need to test submitting and amending all elements that have sequence numbers in QA to ensure normal operation

PR #2207, opened 2024-10-31
CEDS-6119 Standardise return links in the declaration journey

ponmani-sundarapandian:
Changes for the [CEDS-6119](https://jira.tools.tax.service.gov.uk/browse/CEDS-6119)

### DIFF
--- a/app/views/section1/not_declarant.scala.html
+++ b/app/views/section1/not_declarant.scala.html
@@ -37,7 +37,7 @@
 
 @govukLayout(
     title = Title("notDeclarent.title"),
-    backButton = Some(BackButton(messages("site.backToPreviousQuestion"), RootController.displayPage))
+    backButton = Some(BackButton(messages("site.backToSelectionPage"), RootController.displayPage))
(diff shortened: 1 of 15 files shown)


app/views/section1/not_declarant.scala.html
```diff
@@ -37,7 +37,7 @@
 
 @govukLayout(
     title = Title("notDeclarent.title"),
-    backButton = Some(BackButton(messages("site.backToPreviousQuestion"), RootController.displayPage))
+    backButton = Some(BackButton(messages("site.backToSelectionPage"), RootController.displayPage))
```
udinobi, line 40: You could use here `ChoiceController.displayPage` instead of `RootController.displayPage` in order to avoid a redirection

ponmani-sundarapandian: The RootController was already existing that's reason we didn't change it initially . We changed it now as per the comments. Thanks @udinobi

udinobi (approved):

PR #2221, opened 2024-11-20
Revert 2208 revert 2207 ceds 6119

ponmani-sundarapandian:


app/views/helpers/BackLinkHelper.scala
```diff
@@ -0,0 +1,42 @@
...
+package views.helpers
+
+import controllers.navigation.Navigator
+import controllers.section4.routes.NatureOfTransactionController
+import controllers.summary.routes.SectionSummaryController
+import forms.section4.Document
+import models.DeclarationType
+import models.requests.JourneyRequest
+
+object BackLinkHelper {
```
udinobi, line 26: As we already have a `PreviousDocumentsHelper` class I would move this `object` in that file as
```
object PreviousDocumentsHelper {
  def backLinkText...
}
```

This way it would be more clear what page this method is for, as at the moment this is not evident.

udinobi (approved):

PR #2233, opened 2025-02-28
DCWL-2930 PR011820 - Stop auto population of data element 2/3 and update UI content

IanWood-HMRC:


IanWood-HMRC:
Replaced by https://github.com/hmrc/customs-declare-exports-frontend/pull/2234

PR #2256, opened 2025-10-01
DCWL-3924

tomndennis:


AdamCoops:
Is it worth upgrading the plat and bootstrap vers to remove the platOps errors?

PR #2288, opened 2026-05-07
DCWL 4505 Remove 4.9.0 Optional Flag

TJSarno:


app/controllers/section4/InvoiceAndExchangeRateController.scala
```diff
@@ -72,9 +71,7 @@ class InvoiceAndExchangeRateController @Inject() (
           InvoiceAndPackageTotals(
             totalAmountInvoiced = invoiceAndExchangeRate.totalAmountInvoiced,
             totalAmountInvoicedCurrency = invoiceAndExchangeRate.totalAmountInvoicedCurrency,
-            agreedExchangeRate = if (appConfig.isOptionalFieldsEnabled) {
-              if (invoiceAndExchangeRate.exchangeRate.nonEmpty) Some("Yes") else Some("No")
-            } else Some(invoiceAndExchangeRate.agreedExchangeRate),
+            agreedExchangeRate = if (invoiceAndExchangeRate.exchangeRate.nonEmpty) Some("Yes") else Some("No"),
```
sergi-hmrc, line 74: Should we keep this optional if we always set a value (other than in the cache with none). We could have it as String defaulted to `""`.

app/views/section6/transport_country.scala.html
```diff
@@ -77,19 +77,14 @@
 
         @heading(messages(s"$prefix.title", transportMode), "govuk-label--xl", "h1", Some(form(transportCountry).id))
 
-            @if(appConfig.isOptionalFieldsEnabled) {
-                @insetTextOnRoRo
-            } else {
-                @nonInsetTextOnRoRo
-            }
+            @insetTextOnRoRo
```
sergi-hmrc: nitpick: indentation for this one and the if below

sergi-hmrc (dismissed):
lgtm but I think we also need to delete `features.optionalFields` in the `Stubs`

sergi-hmrc (approved):

